PR #1, opened 2018-05-18
Verification: Add verification time

MAltgelt:
Add a time for verification, allowing to verify a certificate for a time
different than the system time.


src/x509.cpp
```diff
@@ -78,6 +78,17 @@ void X509Certificate::VerificationContext::validityCheck() const
     }
 }
 
+X509Certificate::VerificationContext&
+X509Certificate::VerificationContext::setVerificationCheckTime(Asn1Time checkTime)
+{
+    try {
+        _verificationCheckTime = std::chrono::system_clock::to_time_t(checkTime.toTimePoint());
```
garw: I think this is a good time to add a toUnixtime() function to the Asn1Time object. Where we also do the range management and let it throw in case that it doesn't fit.

MAltgelt:
I've added a toTimeT method to Asn1Time. This should also extend the potential amount of supported times (time_point may have a lower supported time range than time_t on some systems).

src/openssl_wrap.cpp
```diff
@@ -614,6 +614,61 @@ time_point _asn1TimeToTimePoint(const ASN1_TIME *time)
     return epoch + (24h * days) + (1s * seconds);
 }
 
+/**
+ * @brief Convert an ASN1_TIME instance to a time_t
+ * @param [in] time An ASN1 encoded time object
+ * @return A time_t object with the same time
+ */
+time_t _asn1TimeToTimeT(const ASN1_TIME *time)
```
garw: This function introduces a lot of code duplication with _asn1TimeToTimepoint. I don't really like this. 
Couldn't we refactor the common parts into another function that is then instrumented to either create time_points or time_t?

src/openssl_wrap.cpp
```diff
@@ -614,6 +614,61 @@ time_point _asn1TimeToTimePoint(const ASN1_TIME *time)
...
+    // first we check if the ASN1_TIME fits into time_t. We do so by comparing it to the
+    // smallest and biggest possible time_t.
+    SSL_ASN1_TIME_Ptr minAsn1TimeT;
+    SSL_ASN1_TIME_Ptr maxAsn1TimeT;
+
+    try {
+        minAsn1TimeT = _ASN1_TIME_from_time_t(minTimeT);
+    } catch (const OpenSSLException&) {
+        // Smallest time_t is smaller than even the smallest Asn1Time.
+        // We don't need to check for lower limits in this case.
```
garw: When should this happen? Is this really an ignorable error? The Creation might also fail because out of memory or something similar?

src/openssl_wrap.cpp
```diff
@@ -614,6 +614,61 @@ time_point _asn1TimeToTimePoint(const ASN1_TIME *time)
...
+    }
+
+    try {
+        maxAsn1TimeT = _ASN1_TIME_from_time_t(maxTimeT);
+    } catch (const OpenSSLException&) {
+        // Biggest time_t is bigger than even the biggest Asn1Time.
```
garw: same here

tests/unit/test_asn1time.cpp
```diff
@@ -80,6 +80,12 @@ TEST_F(Asn1TimeTest, fromTimePointToTimePointIsIdentical)
     EXPECT_THAT(t.toTimePoint(), Eq(tp_2017_3_27__19_00_38));
 }
 
+TEST_F(Asn1TimeTest, fromTimeTToTimeTIsIdentical)
+{
+    auto t = Asn1Time::fromTimeT(tt_2017_3_27__19_00_38);
+    EXPECT_THAT(t.toTimeT(), Eq(tt_2017_3_27__19_00_38));
+}
+
```
garw: Please also add a test for the out of bounds cases.

src/openssl_wrap.cpp
MAltgelt: Done. I've replaced this with some other code that uses numeric comparison instead.

src/openssl_wrap.cpp
MAltgelt: Done.

src/openssl_wrap.cpp
MAltgelt: Done.

tests/unit/test_asn1time.cpp
MAltgelt: For time_point these tests already exist.
This is not as simple for time_t because the "out of bounds" cases depend on the platform. On the platform I'm using, time_t is a 128 bit integer (which makes its scope way larger than Asn1Time's).
Do you have an idea for a platform independent out of bounds test case?

tests/unit/test_verification.cpp
```diff
@@ -129,6 +149,42 @@ TEST_F(VerificationTest, testFutureCertValidationFails)
...
+TEST_F(VerificationTest, testFutureCertValidationSucceedsWithTimeSetAccordingly)
+{
+    VerificationContext ctx;
+    ctx.addTrustedCertificate(*_root1_future.get())
+       .setVerificationCheckTime(_root1_future->getNotBeforeAsn1() + Asn1Time::Seconds(1));
+
+    _root1_future->verify(ctx);
+}
+
+TEST_F(VerificationTest, testCertValidationFailsWithTimeInFuture)
```
vkrikun: @MAltgelt , tests content is fine, but names are a bit confusing. Wouldn't testCertValidationFailWhenCertificateIsNotYetValid () better reflect what we actually testing? (same for other added tests)

MAltgelt: Yes, but that name might also refer to a certificate that only becomes valid in the future. It doesn't imply that the time for the verification check is set.
Could we agree on testCertValidationFailWhenCertificateIsNotYetValidAtVerificationTime and similar names?

vkrikun: I think testCert**Verification**FailWhenCertificateIsNotYetValidAt**Verification**Time() is an overkill. To me, it is obvious that we test "verification" at "verification" time. I do not have strong feelings about it, so if you disagree, let's leave it unchanged.

MAltgelt: I disagree with your proposed nomenclature, I think, but I haven't got strong feelings about the test names either, so should someone else request a change here, I'll go along.

tests/unit/test_asn1time.cpp
MAltgelt: As discussed, I've added a static_assert in the test case that fails should time_t not be large enough.

PR #5, opened 2018-07-13
Create hashing module with initial sha256 and sha256 support

miroshko:


src/mococrw/hash.h
```diff
@@ -0,0 +1,54 @@
...
+ * #L%
+ */
+#pragma once
+
+#include "mococrw/openssl_wrap.h"
+
+namespace mococrw
+{
+
+void sha256(std::string message, std::string *outDigest);
```
garw: I am ok with these convenience functions. However, why don't they just return the digest? Moreover, I have the impression that it is strange that they sometimes return a string an sometimes raw bytes. We no have the convenience hex conversion function. Hence, I would opt for always return bytes (in form of a memory managed vector)

src/mococrw/hash.h
```diff
@@ -0,0 +1,54 @@
...
+void sha512(std::string message, std::string *outDigest);
+void sha512(std::vector<uint8_t> message, std::vector<uint8_t> *outDigest);
+void sha512(uint8_t* message, int messageLength, std::vector<uint8_t> *outDigest, int *outDigestLength);
+
+class Hash
+{
+public:
+    static Hash sha256();
+    static Hash sha512();
+    void digest (std::vector<uint8_t> &outValue);
```
garw: see above: Why not just return it? If we want to offer a function to return hex here, we can also add (like it is in the python api) a hexdigest function.

src/mococrw/hash.h
```diff
@@ -0,0 +1,54 @@
...
+    void digest (std::string &outValue);
+    void digest (uint8_t* outValue, unsigned int &outDigestLength);
+    unsigned int getDigestLengthBytes ();
+    Hash& update (const std::vector<uint8_t> &chunk);
+    Hash& update (const std::string &chunk);
+    Hash& update (const uint8_t* chunk, const int &length);
```
garw: Ints as const reference? This looks strange to me.

src/mococrw/hash.h
```diff
@@ -0,0 +1,54 @@
...
+    static std::map<openssl::DigestTypes, int> lengthBytes;
```
garw: Why is this a public member? Moreover, it is non-const. This has the potential for problems as other people could modify it globally.

src/mococrw/hash.h
```diff
@@ -0,0 +1,54 @@
...
+private:
+    Hash(const openssl::DigestTypes _digestType);
```
garw: Our coding convention is to use underscore  for member variables not for parameters.

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+ */
+#include "mococrw/hash.h"
+#include "mococrw/util.h"
+#include "mococrw/openssl_wrap.h"
+
+namespace mococrw
+{
+using namespace openssl;
+
+static const EVP_MD* _getDigestFn (DigestTypes digestType) {
```
garw: static for free functions is not really modern C++ style. If you want to keep your functions local, wrap them into an anonymous (unnamed) namespace. Prefixing them with _ is ok to make it clear directly from reading.

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+            throw std::runtime_error("Unknown digest type");
+    }
+}
+
+static EVP_MD_CTX _initCtx (DigestTypes digestType) {
+    const EVP_MD* digestFn = _getDigestFn(digestType);
+    EVP_MD_CTX digestCtx;
+    EVP_MD_CTX_init(&digestCtx);
+    EVP_DigestInit_ex(&digestCtx, digestFn, NULL);
+    return digestCtx;
```
garw: This looks potentially dangerous to me. This relies on 1:1 struct copy for the locally allocated and initialized context. Please double check that this is ok with openssl.

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+static void _finalizeDigest (EVP_MD_CTX& digestCtx, uint8_t* outDigest, unsigned int &outLength) {
+    EVP_DigestFinal_ex(&digestCtx, outDigest, &outLength);
+    EVP_MD_CTX_cleanup(&digestCtx);
+}
```
garw: I am a bit unsure if we really want to design it like this. If you read the class implementations, you always have to redirect one step to these free functions, which (themselves) implement essentially one of the functions of the class. Why not implement this as part of the class?

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+static EVP_MD_CTX _initCtx (DigestTypes digestType) {
+    const EVP_MD* digestFn = _getDigestFn(digestType);
+    EVP_MD_CTX digestCtx;
+    EVP_MD_CTX_init(&digestCtx);
+    EVP_DigestInit_ex(&digestCtx, digestFn, NULL);
+    return digestCtx;
+}
+
+static void _finalizeDigest (EVP_MD_CTX& digestCtx, uint8_t* outDigest, unsigned int &outLength) {
```
garw: Please use output parameters only if this is unavoidable as code with output parameters is in general hard to read. If there is only one output parameter, I don't see it.

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+static void _hash(DigestTypes digestType, const uint8_t* message, const int &length, uint8_t* outDigest, unsigned int &outLength) {
+    EVP_MD_CTX digestCtx = _initCtx(digestType);
+    EVP_DigestUpdate(&digestCtx, message, length);
+    _finalizeDigest(digestCtx, outDigest, outLength);
+}
+
+static void _hash(DigestTypes digestType, const std::string &message, std::string &outDigest) {
+    uint8_t value[Hash::lengthBytes[digestType]];
```
garw: As a general comment on your whole patch: Variable length local arrays are not really portable. Please use a vector instead.

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+    EVP_DigestInit_ex(&digestCtx, digestFn, NULL);
+    return digestCtx;
+}
+
+static void _finalizeDigest (EVP_MD_CTX& digestCtx, uint8_t* outDigest, unsigned int &outLength) {
+    EVP_DigestFinal_ex(&digestCtx, outDigest, &outLength);
+    EVP_MD_CTX_cleanup(&digestCtx);
+}
+
+static void _hash(DigestTypes digestType, const uint8_t* message, const int &length, uint8_t* outDigest, unsigned int &outLength) {
```
garw: This function does no memory management at all and kind of outsources all the work to the caller. This is generally a rather C-style design and should be avoided in C++ as your callers have a tendency of doing things way too often. I would prefer a design that implements all functionality in a clean way in the class. The free functions should just use the class interface to implement their behavior. They are just convenience wrappers, not a high speed shortcut.

See also my other comments on output parameters.

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+    _hash(digestType, reinterpret_cast<const uint8_t*>(message.c_str()), message.length(), value, length);
+    outDigest = utility::toHex(value, length);
+}
+
+static void _hash(DigestTypes digestType, const std::vector<uint8_t> &message, std::vector<uint8_t> &outDigest) {
+    uint8_t value[Hash::lengthBytes[digestType]];
+    unsigned int length;    
+    _hash(digestType, &message[0], message.size(), value, length);
+    outDigest.insert(outDigest.begin(), value, value + length);
+    outDigest.resize(length);
```
garw: You end of with this strange pattern because you can't really really on the state of your output parameter. This is one of these situations where you see why this is just not preferred design.

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+    uint8_t value[Hash::lengthBytes[digestType]];
+    unsigned int length;
+    _hash(digestType, reinterpret_cast<const uint8_t*>(message.c_str()), message.length(), value, length);
+    outDigest = utility::toHex(value, length);
+}
+
+static void _hash(DigestTypes digestType, const std::vector<uint8_t> &message, std::vector<uint8_t> &outDigest) {
+    uint8_t value[Hash::lengthBytes[digestType]];
+    unsigned int length;    
+    _hash(digestType, &message[0], message.size(), value, length);
```
garw: consider using vector.data() instead of &vector[0]. I consider it less noisy

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+void Hash::digest (std::vector<uint8_t> &outValue) {
+    unsigned int length;
+    uint8_t value[getDigestLengthBytes()];
+    digest(value, length);
+    outValue.insert(outValue.begin(), value, value + length);
+    outValue.resize(length);
+}
+
+void Hash::digest (uint8_t* outValue, unsigned int &length) {
```
garw: What happens if this is called on an already finalized object? Is this sane?

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+    _hash(DigestTypes::SHA512, message, outDigest);
+}
+
+std::map<DigestTypes, int> Hash::lengthBytes = {
+    { DigestTypes::SHA256, 256 / 8 },
+    { DigestTypes::SHA512, 512 / 8 }
+};
+
+Hash::Hash(const DigestTypes _digestType) : _digestType(_digestType) {
+    _digestCtx = _initCtx(_digestType);
```
garw: We don't have RAII pattern on the context. Consequently, there is a leak if the hash is (for whatever reason) never finalized. Use RAII here to avoid this.

src/util.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "mococrw/util.h"
+#include <sstream>
+#include <iomanip>
+
+namespace mococrw
+{
+namespace utility
+{
+
+std::string toHex(std::vector<uint8_t> &data) {
```
garw: const?

src/hash.cpp
```diff
@@ -0,0 +1,156 @@
...
+    return update(reinterpret_cast<const uint8_t*>(chunk.c_str()), chunk.length());
+}
+
+Hash& Hash::update (const std::vector<uint8_t> &chunk) {
+    return update(&chunk[0], chunk.size());
+}
+
+Hash& Hash::update (const uint8_t* chunk, const int &length) {
+    if(_finalized) {
+        throw std::runtime_error("update method cannot be called after digest was called");
```
garw: Reconsider exception strategy. Do we throw runtime error anywhere else? I think this library mostly throws MoCoCrW error.

tests/unit/test_hash.cpp
```diff
@@ -0,0 +1,143 @@
...
+TEST_F(HashTest, throwsIfCalledUpdateAfterDigest)
+{
+    std::string value;
+    EXPECT_THROW({
+        Hash sha256 = Hash::sha256();
+        sha256.update("foo");
+        sha256.digest(value);
+        sha256.update("bar");
+    }, std::runtime_error);
+}
```
garw: Tests for toHex are missing. Moreover tests for the binary output of the hash functions are missing.
Yes, they are tested implicitly by the other tests but if such a test fails it gives you a wrong indication where the problem is and kind of negates what you want to achieve with testing (avoid lengthy debugging)

src/mococrw/openssl_wrap.h
```diff
@@ -48,6 +48,16 @@ struct SSLDeleter
     }
 };
 
+/*
+ * Deleter template which doesn't delete, e.g. to wrap const pointers
+ * but not delete them.
+ */
+struct noopDeleter
```
garw: Smart pointers are a  drop-in replacement that look like pointers but do memory management. If no memory management needs to be done, there is no need for smart pointers. This looks like over engineering to me. What's the point of having a smart pointer that does exactly nothing on delete?

src/hash.cpp
```diff
@@ -24,33 +24,22 @@ namespace mococrw
...
+    lib::OpenSSLLib::SSL_EVP_MD_CTX_init(digestCtx.get());
+    lib::OpenSSLLib::SSL_EVP_DigestInit_ex(digestCtx.get(), digestFn.get(), NULL);
     return digestCtx;
 }
 
-static void _finalizeDigest (EVP_MD_CTX& digestCtx, uint8_t* outDigest, unsigned int &outLength) {
-    EVP_DigestFinal_ex(&digestCtx, outDigest, &outLength);
-    EVP_MD_CTX_cleanup(&digestCtx);
+static void _finalizeDigest (SSL_EVP_MD_CTX_Ptr &digestCtx, uint8_t* outDigest, unsigned int &outLength) {
+    lib::OpenSSLLib::SSL_EVP_DigestFinal_ex(digestCtx.get(), outDigest, &outLength);
```
garw: I guess we need to talk about this again. This is not the style of the library and also does not do the error handling that we would like to see here.  The high level code should not call openssl functions directly but call abstractions from openssl_wrap that do the error handling for you. This piece of code has the effect now that error transferred via return value of the openssl functions are (silently) not handled.

src/hash.cpp
```diff
@@ -24,33 +24,22 @@ namespace mococrw
...
+    lib::OpenSSLLib::SSL_EVP_MD_CTX_cleanup(digestCtx.get());
```
garw: This call is not necessary anymore as it is already captured by the RAII pattern in SSL_EVP_MD_CTX_Ptr. In fact, this could (depending on openssl implementation) lead to a double free bug.

src/mococrw/hash.h
miroshko: could be just int, will change

src/mococrw/hash.h
miroshko: should be const, I agree. It is public so the users can get this info e.g. for creating uint8_t array of the necessary size

src/mococrw/openssl_wrap.h
miroshko: no practical point, the only argument was keeping it coherent with the rest of pointers returned by the library (the comment in the file suggests "Wrap all the pointer-types returned by openssl.")

src/mococrw/hash.h
miroshko: switched all functions to return a vector

src/mococrw/hash.h
miroshko: all functions return vectors now. hex can be reached by using utility::hex() so hexdigest ist probably not needed

src/mococrw/hash.h
miroshko: changed to int

src/mococrw/hash.h
miroshko: hidden to a private member

src/hash.cpp
miroshko: not needed anymore since all free private functions are gone now

src/hash.cpp
miroshko: done

src/hash.cpp
miroshko: done

src/hash.cpp
miroshko: done

src/hash.cpp
miroshko: done

src/hash.cpp
miroshko: added a check and caching the output value now

src/util.cpp
miroshko: done

src/hash.cpp
miroshko: done

tests/unit/test_hash.cpp
miroshko: added tests for toHex

src/mococrw/openssl_wrap.h
miroshko: changed to using the direct OpenSSL type

tests/unit/test_hash.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+TEST_F(HashTest, returnsDigestLength) {
+    EXPECT_THAT(Hash::sha256().getDigestLengthBytes(), Eq(32));
+    EXPECT_THAT(Hash::sha512().getDigestLengthBytes(), Eq(64));
+}
+
+TEST_F(HashTest, throwsIfCalledUpdateAfterDigest)
+{
+    std::string value;
+    EXPECT_THROW({
```
garw: Better confine where we expect the throw. We want it to not throw until digest and only when update() is called afterwards. This way you might mask different problems.

src/hash.cpp
garw: looks still like runtime_error to me? Or is there any rationale behind this.

src/hash.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+Hash Hash::sha512 () {
+    return Hash{DigestTypes::SHA512};
+}
+
+std::vector<uint8_t> Hash::digest () {
+    if (_finalDigestValue.empty()) {
+        unsigned int length;
+        _finalDigestValue.resize(getDigestLengthBytes());
+        lib::OpenSSLLib::SSL_EVP_DigestFinal_ex(_digestCtx.get(), _finalDigestValue.data(), &length);
```
garw: Please do not call openssl functions from the code directly. Please define a wrapper that does the necessary error-handling in openssl_wrap.c. See other functions or ask on how this is usually done with this callChecked templates.

This way you just ignore the return value and do no error handling in case of openssl errors. One of the major goals of this libraries to get rid of this kind of behavior.

src/hash.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+Hash& Hash::update (const std::vector<uint8_t> &chunk) {
+    return update(&chunk[0], chunk.size());
+}
+
+Hash& Hash::update (const uint8_t* chunk, int length) {
+    if(!_finalDigestValue.empty()) {
+        throw std::runtime_error("update method cannot be called after digest was called");
+    }
+    lib::OpenSSLLib::SSL_EVP_DigestUpdate(_digestCtx.get(), chunk, length);
```
garw: see above on calling openssl functions directly.

src/hash.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+std::map<DigestTypes, int> Hash::lengthBytes = {
+    { DigestTypes::SHA256, 256 / 8 },
+    { DigestTypes::SHA512, 512 / 8 }
+};
+
+Hash::Hash(const DigestTypes digestType) : _digestType(digestType) {
+    const EVP_MD* digestFn = _getMDPtrFromDigestType(digestType);
+    _digestCtx = _EVP_MD_CTX_create();
+    lib::OpenSSLLib::SSL_EVP_MD_CTX_init(_digestCtx.get());
```
garw: Please see below on calling openssl functions directly.

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+#pragma once
+
+#include "mococrw/openssl_wrap.h"
+
+namespace mococrw
+{
+
+std::vector<uint8_t> sha256(const std::vector<uint8_t> &message);
+std::vector<uint8_t> sha256(uint8_t* message, int messageLength);
```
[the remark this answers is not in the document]
miroshko: done

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+
+std::vector<uint8_t> sha512(const std::vector<uint8_t> &message);
+std::vector<uint8_t> sha512(uint8_t* message, int messageLength);
+
+class Hash
+{
+public:
+    static Hash sha256();
+    static Hash sha512();
+    std::vector<uint8_t> digest ();
```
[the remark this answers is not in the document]
miroshko: done

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+    std::string hexdigest ();
+    unsigned int getDigestLengthBytes ();
```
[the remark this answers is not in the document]
miroshko: done

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+    Hash& update (const std::vector<uint8_t> &chunk);
+    Hash& update (const std::string &chunk);
+    Hash& update (const uint8_t* chunk, int length);
```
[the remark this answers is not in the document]
miroshko: done

tests/unit/test_hash.cpp
miroshko: done

src/hash.cpp
miroshko: I created basic wrappers. Do you think any custom error handling inside the wrappers would make sense here, e.g. when update is called after finalizing

src/hash.cpp
miroshko: done

src/hash.cpp
miroshko: done

src/openssl_wrap.cpp
```diff
@@ -140,6 +140,22 @@ SSL_X509_REQ_Ptr _X509_REQ_new()
     return SSL_X509_REQ_Ptr{OpensslCallPtr::callChecked(lib::OpenSSLLib::SSL_X509_REQ_new)};
 }
 
+int _EVP_DigestFinal_ex(EVP_MD_CTX* ctx, unsigned char* md, unsigned int* s) {
```
garw: This is the place where OpenSSL error handling should take place. In particular, conversion of integer based function return codes to C++ exceptions. This is what all the ::callChecked API is for. The functions itself should be void if they have no apparent output.

src/util.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <iomanip>
+
+namespace mococrw
+{
+namespace utility
+{
+
+std::string toHex(const std::vector<uint8_t> &data) {
+    std::stringstream result;
+    for(int i = 0, size = data.size(); i < size; i++) {
```
garw: size_t? why size=data.size(). Let the compiler do the optimizations.

src/hash.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+Hash Hash::sha512() {
+    return Hash{DigestTypes::SHA512};
+}
+
+std::vector<uint8_t> Hash::digest() {
+    if (_finalDigestValue.empty()) {
+        unsigned int length;
+        _finalDigestValue.resize(getDigestLengthInBytes());
+        _EVP_DigestFinal_ex(_digestCtx.get(), _finalDigestValue.data(), &length);
```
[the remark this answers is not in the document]
miroshko: done

src/hash.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+std::vector<uint8_t> sha512(const std::string &message) {
+    return Hash::sha512().update(message).digest();
+}
+
+std::vector<uint8_t> sha512(const std::vector<uint8_t> &message) {
+    return Hash::sha512().update(message).digest();
+}
+
+std::map<DigestTypes, size_t> Hash::lengthInBytes = {
```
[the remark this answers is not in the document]
miroshko: done

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+#pragma once
+
+#include "mococrw/openssl_wrap.h"
+
+namespace mococrw
+{
+
+std::vector<uint8_t> sha256(const std::vector<uint8_t> &message);
+std::vector<uint8_t> sha256(uint8_t* message, size_t messageLength);
```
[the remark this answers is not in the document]
miroshko: done

src/util.cpp
miroshko: done

src/openssl_wrap.cpp
miroshko: done

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+
+std::vector<uint8_t> sha512(const std::vector<uint8_t> &message);
+std::vector<uint8_t> sha512(uint8_t* message, size_t messageLength);
```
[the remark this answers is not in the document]
miroshko: done

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+
+class Hash
+{
+public:
+    static Hash sha256();
+    static Hash sha512();
+    std::vector<uint8_t> digest();
+    std::string hexdigest();
+    size_t getDigestLengthInBytes();
```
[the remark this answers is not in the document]
miroshko: done

miroshko:
@garw i fixed your comments. please take another look

src/hash.cpp
```diff
@@ -0,0 +1,102 @@
...
+size_t Hash::getDigestLengthInBytes() const {
+    return Hash::lengthInBytes.at(_digestType);
+}
+
+Hash& Hash::update(const std::string &chunk) {
+    return update(reinterpret_cast<const uint8_t*>(chunk.c_str()), chunk.length());
+}
+
+Hash& Hash::update(const std::vector<uint8_t> &chunk) {
+    return update(&chunk[0], chunk.size());
```
vkrikun: chunk.data() might look a bit cleaner

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+std::vector<uint8_t> sha512(const uint8_t* message, size_t messageLength);
+
+class Hash
+{
+public:
+    static Hash sha256();
+    static Hash sha512();
+    std::vector<uint8_t> digest();
+    std::string hexdigest();
+    size_t getDigestLengthInBytes() const;
```
vkrikun: getDigestLengthInBytes() does not fit into the concept of Hash implemented as a builder neither a stand-alone function. In case of a stand-alone function, you can't access it at all. In case of a builder, you would have to break the chain for update().update().digest() calls.

Maybe a static method or a static variable?

tests/unit/test_hash.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+using testing::Eq;
+
+class HashTest : public ::testing::Test
+{
+public:
+    void SetUp() override;
+};
+
+void HashTest::SetUp()
```
vkrikun: is it really needed?

src/mococrw/hash.h
```diff
@@ -0,0 +1,51 @@
...
+std::vector<uint8_t> sha512(const std::vector<uint8_t> &message);
+std::vector<uint8_t> sha512(const uint8_t* message, size_t messageLength);
+
+class Hash
+{
+public:
+    static Hash sha256();
+    static Hash sha512();
+    std::vector<uint8_t> digest();
+    std::string hexdigest();
```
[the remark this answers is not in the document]
miroshko: agree. it was even not implemented.
removed

src/hash.cpp
miroshko: done

src/mococrw/hash.h
miroshko: agree, because vector is anyway the only type we return and it has the length

tests/unit/test_hash.cpp
miroshko: no. removed

miroshko:
@vkrikun done

garw:
merged

PR #6, opened 2018-07-20
Added support for the x509_check_ca function

fmonteiro93:
Added functionality to check if a certificate is a CA to MoCoCrw


src/mococrw/x509.h
```diff
@@ -112,6 +112,13 @@ class X509Certificate
 
     AsymmetricPublicKey getPublicKey() const;
 
+    /**
+     * @brief Verify if the certificate is a CA
+     *
+     * @return Whether the certificate is a CA or not
+     */
+    bool checkIfCA() const;
```
garw: maybe just isCA?

src/openssl_wrap.cpp
```diff
@@ -440,6 +440,10 @@ X509_VERIFY_PARAM* _X509_STORE_CTX_get0_param(X509_STORE_CTX *ctx) {
     return OpensslCallPtr::callChecked(lib::OpenSSLLib::SSL_X509_STORE_CTX_get0_param, ctx);
 }
 
+void _X509_check_ca(X509 *cert) {
+    OpensslCallIsPositive::callChecked(lib::OpenSSLLib::SSL_X509_check_ca, cert);
```
garw: The callChecked facility is meant for error handling. If I have a look at the openssl code, it looks like this function does not return error codes but 0 if it is not a CA and other positive codes depending on which kind of CA was detected. Hence, I would propose to make this function return a boolean (other things are not of interest to us anyways) and return true iff the call returns a positive number.

src/x509.cpp
```diff
@@ -181,6 +181,17 @@ AsymmetricPublicKey X509Certificate::getPublicKey() const
     return AsymmetricPublicKey(std::move(pubkey));
 }
 
+bool X509Certificate::checkIfCA() const
+{
+    bool result = true;
+    try {
+         _X509_check_ca(_x509.get());
+    } catch (const OpenSSLException& e) {
```
garw: Exceptions should only be used in exceptional cases. I would argue that this is not really exceptional. Once you adapted the callChecked as remarked above, you can just forward this call to it.

src/mococrw/x509.h
fmonteiro93: Agreed.

src/openssl_wrap.cpp
fmonteiro93: Understood. Corrected on new version.

src/x509.cpp
fmonteiro93: Agreed and corrected

tests/unit/test_x509.cpp
```diff
@@ -568,3 +568,11 @@ TEST_F(X509Test, testGetSerialNumber)
     EXPECT_EQ("-42", negativeSerialCert.getSerialNumberDecimal());
     EXPECT_EQ(std::vector<uint8_t>{0x2a}, negativeSerialCert.getSerialNumberBinary());
 }
+
+TEST_F(X509Test, testCACheck)
+{
+    EXPECT_EQ(false, _root1_expired->isCA())
```
vkrikun: I would consider EXPECT_TRUE and EXPECT_FALSE  instead of asserting equality of boolean variables.

tests/unit/test_x509.cpp
```diff
@@ -568,3 +568,11 @@ TEST_F(X509Test, testGetSerialNumber)
     EXPECT_EQ("-42", negativeSerialCert.getSerialNumberDecimal());
     EXPECT_EQ(std::vector<uint8_t>{0x2a}, negativeSerialCert.getSerialNumberBinary());
 }
+
+TEST_F(X509Test, testCACheck)
```
vkrikun: I would separate positive and negative tests and give them more self-documented names. E.g testIntermediateCertificateIsCA(), testExpiredCertificateIsNotCA(), testClientCertificateIsNotCa() or something.

tests/unit/test_x509.cpp
```diff
@@ -568,3 +568,11 @@ TEST_F(X509Test, testGetSerialNumber)
...
+{
+    EXPECT_EQ(false, _root1_expired->isCA())
+        << "X509Certificate::isCA() should have returned false";
```
vkrikun: This assert description repeats what is in the code and what will be reported by the test framework. It does not give you a better understanding of **why** the test is failing. I would you something like "Client Certificate cannot be CA" or, if your test-cases are small enough, remove messages at all.

tests/unit/test_x509.cpp
```diff
@@ -62,6 +62,9 @@ class X509Test : public ::testing::Test
     std::unique_ptr<X509Certificate> _root1_int1_cert1;
 
     std::unique_ptr<X509Certificate> _root2;
+    std::unique_ptr<X509Certificate> _root2_int1;
```
vkrikun: I wouldn't bloat test class and SetUp() method this something which is used once in one single test-case. If you need it in one place - limit the scope and create it right in the test method. Make sense?

fmonteiro93: Makes sense. Changed the scope of the added certificates accordingly.

tests/unit/test_x509.cpp
```diff
@@ -603,7 +604,7 @@ TEST_F(X509Test, testClientCertificateIsNotCA)
 
 TEST_F(X509Test, testExpiredCertificateIsNotCA)
 {
```
garw: I find this test very confusing as it suggests that an expired CA certificate is not marked as CA anymore. However, this is not tested here _root1_expired is an experired (client) certificate issued by root1-CA - so it never was a CA in the first place. I would assume that also invalid certificates (because they are not valid yet or already invalid) are (correctly) marked as CA. Please adapt the test. I think among these "yearXXXX" certificates, there should be suitable ones (maybe none which is expired but some that are not valid yet).

fmonteiro93: Adapted test as suggested. It was necessary to add a new certificate that was both a CA and expired because none of the yearXXXX are expired (year in file name refer to the notBefore time).

PR #8, opened 2018-09-05
 Adapted MoCOCrW to work with OpenSSL v1.1.x 

fmonteiro93:
- Changes were made in the the MoCOCrW source code for it
to be able to work with OpenSSL v1.1.x but still maintain compatibility
with OpenSSL v1.0.2.
- Updated Unit Tests and added some missing coverage


fmonteiro93:
Just a small note. Commits before f7a8917 have already been up streamed before (https://github.com/bmwcarit/MoCOCrW/commit/9f218f03389f8f374acca68378decb23ff28ff59) on the 25th of July. They only show up here because I used the same fork.

src/mococrw/openssl_lib.h
```diff
@@ -115,17 +115,19 @@ class OpenSSLLib
     static int SSL_EVP_PKEY_CTX_set_rsa_keygen_bits(EVP_PKEY_CTX* ctx, int mbits) noexcept;
     static int SSL_EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b) noexcept;
 
+    #if OPENSSL_VERSION_NUMBER < 0x10100000L
```
garw: It seems to me that this method isn't used anyways. Just remove it

src/mococrw/openssl_wrap.h
```diff
@@ -209,15 +209,17 @@ void _EVP_PKEY_CTX_set_rsa_keygen_bits(EVP_PKEY_CTX* ctx, int mbits);
...
+ * This function is only available when using an OpenSSL version older than 1.1.0
  *
  * @param pointer A pointer to the reference counter
  * @param amount The value that the reference counter should be changed with.
  *               For example, 1 to increment one or -1 to decrement by one.
  * @param type  The lock type that should be used.
  * @return The new value of the reference counter
  */
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 int _CRYPTO_add(int *pointer, int amount, int type);
```
garw: See above

src/openssl_lib.cpp
```diff
@@ -54,7 +54,15 @@ namespace lib
...
  */
 
-void OpenSSLLib::SSL_CRYPTO_malloc_init() noexcept { CRYPTO_malloc_init(); }
+void OpenSSLLib::SSL_CRYPTO_malloc_init() noexcept 
+{ 
+    //If added to maintain retrocompatibility with older OpenSSL versions
+    #if OPENSSL_VERSION_NUMBER < 0x10100000L
+        CRYPTO_malloc_init();
+    #else
+        OPENSSL_malloc_init();
```
garw: OpenSSL documentation says: 

> OPENSSL_malloc_init()
>  sets the lower-level memory allocation functions to their default implementation. It is generally not necessary to call this, except perhaps in certain shared-library situations.

Do we really need this?

src/openssl_lib.cpp
```diff
@@ -174,11 +184,15 @@ X509_REQ* OpenSSLLib::SSL_PEM_read_bio_X509_REQ(BIO *bp,
...
 
-BIO_METHOD* OpenSSLLib::SSL_BIO_s_mem() noexcept { return BIO_s_mem(); }
+const BIO_METHOD* OpenSSLLib::SSL_BIO_s_mem() noexcept { return BIO_s_mem(); }
 
 void OpenSSLLib::SSL_BIO_free_all(BIO* ptr) noexcept { BIO_free_all(ptr); }
 
-BIO* OpenSSLLib::SSL_BIO_new(BIO_METHOD* method) noexcept { return BIO_new(method); }
+BIO* OpenSSLLib::SSL_BIO_new(const BIO_METHOD* method) noexcept 
+{ 
+    //Const cast is needed in order to maintain retrocompatibility with OpenSSL 1.0.2 and older.
```
garw: Please create a new branch and provide a clean soltion. I want to get the code base cleaned up and not the maximum compatibility. OpenSSL 1.0.2x support will be dropped then.

src/openssl_lib.cpp
```diff
@@ -427,7 +441,7 @@ STACK_OF(X509)* OpenSSLLib::SSL_sk_X509_new_null() noexcept
 
 int OpenSSLLib::SSL_sk_X509_push(STACK_OF(X509)* stack, const X509 *crt) noexcept
 {
-    return sk_X509_push(stack, crt);
+    return sk_X509_push(stack,  const_cast<X509*>(crt));
```
garw: Are you sure this is needed? This looks very dubious to me. OPENSSL_sk_push takes const void*. So it should accept const ptrs here as well.

src/openssl_lib.cpp
```diff
@@ -556,11 +570,21 @@ int OpenSSLLib::SSL_X509_CRL_verify(X509_CRL* a, EVP_PKEY* r) noexcept
 }
 ASN1_TIME* OpenSSLLib::SSL_X509_CRL_get_nextUpdate(const X509_CRL* x) noexcept
 {
-    return X509_CRL_get_nextUpdate(x);
+    //Added to maintain retrocompatibility with older OpenSSL versions
+    #if OPENSSL_VERSION_NUMBER < 0x10100000L
+        return X509_CRL_get_nextUpdate(x);
+    #else
+        return const_cast<ASN1_TIME*>(X509_CRL_get0_nextUpdate(x));
```
garw: Why? Please rather update this method to return const.

src/openssl_lib.cpp
```diff
@@ -556,11 +570,21 @@ int OpenSSLLib::SSL_X509_CRL_verify(X509_CRL* a, EVP_PKEY* r) noexcept
...
+    #endif
 }
 ASN1_TIME* OpenSSLLib::SSL_X509_CRL_get_lastUpdate(const X509_CRL* x) noexcept
 {
-    return X509_CRL_get_lastUpdate(x);
+    //Added to maintain retrocompatibility with older OpenSSL versions
+    #if OPENSSL_VERSION_NUMBER < 0x10100000L
+        return X509_CRL_get_lastUpdate(x);
+    #else
+        return const_cast<ASN1_TIME*>(X509_CRL_get0_lastUpdate(x));
```
garw: see above

src/openssl_lib.cpp
```diff
@@ -596,7 +620,7 @@ STACK_OF(X509_CRL)* OpenSSLLib::SSL_sk_X509_CRL_new_null() noexcept
 }
 int OpenSSLLib::SSL_sk_X509_CRL_push(STACK_OF(X509_CRL)* stack, const X509_CRL* crl) noexcept
 {
-    return sk_X509_CRL_push(stack, crl);
+    return sk_X509_CRL_push(stack, const_cast<X509_CRL*>(crl));
```
garw: see above. Shoudln't be needed. Clarify why. Remove if humanly possible

tests/unit/test_asn1time.cpp
```diff
@@ -65,7 +65,7 @@ TEST_F(Asn1TimeTest, parseFromUTCTime)
 TEST_F(Asn1TimeTest, parseFromGeneralizedTime)
 {
     auto refFromTimePoint = Asn1Time::fromTimePoint(tp_2017_3_27__19_00_38);
-    auto fromGeneralizedTime = Asn1Time::fromString("20170327190038");
+    auto fromGeneralizedTime = Asn1Time::fromString("20170327190038Z");
```
garw: Is this a breaking change in OpenSSL? When we might want to compensate for this in our library to avoid client code adaptions

tests/unit/test_crl.cpp
```diff
@@ -176,3 +208,23 @@ TEST_F(CRLTest, testConvertionFromAndToPEMIsNoOp)
...
+    auto crlList = mococrw::util::loadCrlPEMChain(crlPemChain);
+
+    //Issuer: CN = CRL Sub CA 1, C = DE, OU = BMW Car IT, O = BMW
+    auto issuerName =crlList.at(0).getIssuerName();
+    EXPECT_EQ("CRL Sub CA 1", issuerName.commonName());
+    EXPECT_EQ("DE", issuerName.countryName());
+    EXPECT_EQ("BMW", issuerName.organizationName());
+    EXPECT_EQ("BMW Car IT", issuerName.organizationalUnitName());
+
+    //Issuer: C = PT, ST = Porto, L = Porto, O = CTW, OU = JC7
```
garw: Please avoid BMW specific names in new tests.

tests/unit/test_opensslwrapper.cpp
```diff
@@ -77,26 +80,34 @@ void OpenSSLWrapperTest::SetUp()
...
+
 /*
  * Test that EVP_PKEY instances are
  * allocated correctly and that the custom deleter
  * will use the correct free function.
  */
 TEST_F(OpenSSLWrapperTest, keyMemoryManagement)
 {
-    EVP_PKEY allocatedKey;
+    EVP_PKEY* allocatedKey  = EVP_PKEY_new();
```
garw: Try to avoid plan memory management

tests/unit/test_opensslwrapper.cpp
```diff
@@ -145,15 +156,15 @@ TEST_F(OpenSSLWrapperTest, keyGenTest)
             .WillOnce(Return(::testutils::somePkeyCtxPtr()));
     EXPECT_CALL(_mock(), SSL_EVP_PKEY_keygen_init(_)).WillOnce(Return(1));
 
-    EVP_PKEY someKey;
+    EVP_PKEY* someKey  = EVP_PKEY_new();
```
garw: See above

tests/unit/test_crl.cpp
fmonteiro93: Agreed, Removed.

src/openssl_lib.cpp
garw: Could you please investigate this further. OpenSSL source code looks to me as if the stack functions should be able to accept a const parameter. Where exactly is the first place where const is breaking things? Breaking const correctness here would be a step backwards. I was under the impression that openssl was moving forward on this end. What is the exact version you are testing against?

tests/unit/test_opensslwrapper.cpp
```diff
@@ -92,22 +92,20 @@ void OpenSSLWrapperTest::TearDown()
  */
 TEST_F(OpenSSLWrapperTest, keyMemoryManagement)
 {
-    EVP_PKEY* allocatedKey  = EVP_PKEY_new();
+    std::unique_ptr<EVP_PKEY, SSLDeleter<EVP_PKEY, EVP_PKEY_free>> allocatedKey{EVP_PKEY_new()};
```
garw: This type already exists as an alias: SSL_EVP_PKEY_Ptr

src/openssl_lib.cpp
fmonteiro93: In the snippets bellow we can see the OpenSSL implementation of 
the function above. The implementation is based of a macro that is completed with the type to use. In this case X509. But the same is valid for the x509_CRL.  


##### x509.h
```
DEFINE_STACK_OF(X509_CRL)
```

##### safestack.h 

```
# define SKM_DEFINE_STACK_OF(t1, t2, t3) \
.
.
.
    static ossl_inline int sk_##t1##_push(STACK_OF(t1) *sk, t2 *ptr) \
    { \
        return OPENSSL_sk_push((OPENSSL_STACK *)sk, (const void *)ptr); \
    } \
.
.
.

# define DEFINE_STACK_OF(t) SKM_DEFINE_STACK_OF(t, t, t)
```

Code after type substitution (Example with X509): 
```
    static ossl_inline int sk_X509_push(STACK_OF(X509) *sk, X509 *ptr) \
    { \
        return OPENSSL_sk_push((OPENSSL_STACK *)sk, (const void *)ptr); \
    } \
```

As we can see, the function no longer accepts const X509 pointers, but the pointer will be casted to const again inside the sk_X509_push function. This means that const correctness will
only be lost during the OpenSSLLib::SSL_sk_X509_CRL_push function.
The other option would be to use the OPENSSL_sk_push function directly instead of using the X509_CRL specific function:
```  
    return OPENSSL_sk_push((OPENSSL_STACK *)stack, crt);
```
With this approach we maintain const correctness but it could be argued that the code is not as easily understood.

Honestly I would be fine with either approach, we just need to decide which. 

A third approach would be to maintain const correctness by removing const from several of Mococrw's high level functionalities. After a first look, at least some of the interfaces in x509.h and bio.h would need to be changed, as well as the stack_utils.h. Personally, I don't like this approach as much as the other two.

.travis.yml
```diff
@@ -3,7 +3,7 @@ services:
  - docker
 
 before_install:
- - docker pull neverpanic/mococrw-build-env
+ - docker pull fmonteiro93/mococrw-build-env
```
neverpanic: Maybe we should open a new axis and use a matrix build here to build in both the OpenSSL 1.0 and the OpenSSL 1.1 container?

According to https://docs.travis-ci.com/user/environment-variables/, something along the lines of

```yaml
env:
  matrix:
    - CONTAINER="fmonteiro93/mococrw-build-env"
    - CONTAINER="neverpanic/mococrw-build-env"
```

and changing the docker pull as well as the docker run to use `$CONTAINER` should probably work.

Also, you'll also have to change the `docker run` statements below to use your container.

neverpanic: Nevermind, just read that you don't plan on keeping 1.0 compatibility. In this case, you'll only have to adjust the `docker run` statements.

neverpanic: And you'll either have to set up automatic builds on hub.docker.com for your new repository or send me a pull request on my repository, ideally for a new branch. I'd then merge the change and configure the automatic builds to build the branch and tag it accordingly.

Ideally, this would all be part of the bmwcarit org as well, but we haven't moved it there yet.

fmonteiro93: Thanks for the input!! I discussed this with @garw and we agreed to leave the setup of Travis outside the scope of this PR and remove the TravisCI verification for the OpenSSL1.1 branch for now. This will for sure be added in the future.

PR #10, opened 2018-10-18
Signature: Add interface for signature and verification

pmvfonseca:
- Supported encryption types: PKCS1 and PSS
- Configurable parameters:
    - Hashing function: SHA256 and SHA512
    - Masking function: SHA256 and SHA512
    - Salt length
- Can sign and verify either a plain text message or a message digest
- Accepts private key for signing
- Accepts public key or certificate for verification
- Added new padding mode class (taken from the RSA encryption pull request)


src/signature.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    return create(ctx, messageDigest);
+}
+
+std::vector<uint8_t> SignatureUtils::create(Context<AsymmetricPrivateKey> &ctx,
+                                            std::vector<uint8_t> messageDigest)
+{
+    std::vector<uint8_t> signature;
+    const auto keyCtx = _EVP_PKEY_CTX_new(ctx.getKey().internal());
```
vkrikun: How the context is released? Am I missing something?

src/signature.cpp
```diff
@@ -0,0 +1,209 @@
...
+        _EVP_PKEY_CTX_set_rsa_padding(keyCtx.get(), static_cast<int>(ctx.getPadding()));
+        _EVP_PKEY_CTX_set_signature_md(keyCtx.get(),
+                                       _getMDPtrFromDigestType(ctx.getHashingFunction()));
+
+        if (RSAPaddingMode::PSS == ctx.getPadding()) {
+            _EVP_PKEY_CTX_set_rsa_pss_saltlen(keyCtx.get(), ctx.getSaltLength());
+            _EVP_PKEY_CTX_set_rsa_mgf1_md(keyCtx.get(),
+                                          _getMDPtrFromDigestType(ctx.getMaskingFunction()));
+        }
+
```
vkrikun: I wonder if this piece of code can be refactored into something like EVP_PKEY_CTX* createSigningContext() and reused in create() and verify().

tests/unit/test_signature.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+using namespace mococrw;
+
+class SignatureTest : public ::testing::Test
+{
+public:
+    void SetUp() override;
+
+protected:
+    mococrw::AsymmetricKeypair _keyPair = AsymmetricKeypair::generate();
```
vkrikun: Redundant as new keypair is generated  for every test in SetUp().

tests/unit/test_signature.cpp
```diff
@@ -0,0 +1,427 @@
...
+/**
+ * @brief Successful signature creation and verification
+ * Cases:
+ * - Sign and verify using PKCS1
+ * - Sign and verify using PCSS
+ * - Sign and verify using SHA512 hashing
+ * - Sign and verify using SHA512 masking
+ * - Sign and verify using non-default salt length
+ */
+TEST_F(SignatureTest, testSignatureSuccessfulVerification)
```
vkrikun: It might be a good idea to split this big test into several test-cases. In case of failures, the bug would be better isolated and we would not reuse the contexts (which might have side-effects).

tests/unit/test_signature.cpp
```diff
@@ -0,0 +1,427 @@
...
+/**
+ * @brief Tests whether the interfaces for creating a signature are all outputing the correct
+ *        PKCS1 signature by comparing it to a result known to be correct.
+ * Cases:
+ * - Sign a plain text message using a signature context
+ * - Sign a message digest using a signature context
+ * - Sign a plain text message using a private key
+ * - Sign a message digest using a private key
+ */
+TEST_F(SignatureTest, testSignaturePKCS1)
```
vkrikun: It is better to have smaller, focused and well isolated tests (see comment below)

tests/unit/test_signature.cpp
```diff
@@ -0,0 +1,427 @@
...
+    /* Use default salt length */
+    ASSERT_NO_THROW(SignatureUtils::create(signatureContext, _testMessage));
+
+    signatureContext.setSaltLength(200);
+    ASSERT_NO_THROW(SignatureUtils::create(signatureContext, _testMessage));
+
+    signatureContext.setSaltLength(0);
+    ASSERT_NO_THROW(SignatureUtils::create(signatureContext, _testMessage));
+
+    signatureContext.setSaltLength(-1);
```
vkrikun: but it might be a good idea to drop a comment here why -1 and -2 are valid (special cases).

tests/unit/test_signature.cpp
```diff
@@ -0,0 +1,427 @@
...
+// Signature Failure cases
+
+/**
+ * @brief Unsuccessful case where the message digest length is different of padding mode,
+ *        considering the usage of PKCS1.
+ */
+TEST_F(SignatureTest, testSignatureInvalidDigestLengthPKCS1)
+{
+    std::vector<uint8_t> signature;
+    std::vector<uint8_t> messageDigest;
```
vkrikun: Mater of taste, but declaring variables way before using them looks very C-ish. (also in other places)

tests/unit/test_signature.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+#include "IOUtils.h"
+
+#include "signature.cpp"
+
+using namespace std::string_literals;
+
+using namespace mococrw;
+
+class SignatureTest : public ::testing::Test
```
vkrikun: Some generic observations:
* Have I missed a use-case where we sign, modify the content and check the signature?
* There is a lot of unit-test which check various age-cases in configuring the context. It basically tests OpenSSL, not our wrapper. Value of such tests is questionable.
* After a brief look, it seems like we have a lot of overlapping tests. I would recommend using test-coverage tooling to reduce the number of tests (if possible).

src/mococrw/signature.h
```diff
@@ -0,0 +1,373 @@
...
+     * This is a simpler verify method that allows the user to pass a public key directly instead
+     * of having to build a context. The drawback is that the default values will be used and
+     * cannot be changed.
+     *
+     * @param publicKey The public key used to verify the message
+     * @param signature The signature to be verified
+     * @param messageDigest The message digest the signature is verified with
+     * @throw MoCOCrWException if the verification fails.
+     */
+    static void verify(const AsymmetricPublicKey &publicKey,
```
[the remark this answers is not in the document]
vkrikun: even though I don't like it, you might be right that it is an accepted design pattern in MoCOCrW. Would be nice, though, that user can distinguish signature mismatch and, for instance, wrong padding configuration. Throwing MoCOCrWException on everything is not the best idea.

src/signature.cpp
vkrikun: Sorry, I was lost in layers of abstractions.

PR #11, opened 2018-10-25
Added interface for asymmetric encryption

fmonteiro93:
- Supported encryption/decryption types: NoPadding, PKCS and OAEP
- Configurable parameters:
    -- Hashing function: SHA256 and SHA512
    -- Masking function: SHA256 and SHA512
    -- Label: User defined

### DIFF
--- a/src/asymmetric_encryption.cpp
+++ b/src/asymmetric_encryption.cpp
@@ -1,0 +1,242 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2018 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include "mococrw/asymmetric_encryption.h"
+
+#include <fstream>
+
+#include "mococrw/error.h"
+#include "mococrw/hash.h"
+#include "mococrw/padding_mode.h"
+#include "mococrw/util.h"
+
+namespace mococrw {
+
+using namespace openssl;
+
+using SSL_RSA_OAEP_LABEL_Ptr = std::unique_ptr<uint8_t, SSLFree<uint8_t>>;
+
+using SSL_RSA_ENCRYPTION_DATA_Ptr = std::unique_ptr<unsigned char, SSLFree<unsigned char>>;
+
+AsymmetricEncryption::CryptoData::CryptoData(const std::string& data)
+    : _data{data.begin(), data.end()}
+{
+}
+
+AsymmetricEncryption::CryptoData::CryptoData(const std::vector<uint8_t>& data)
+    : _data{data}
+{
+}
+
+AsymmetricEncryption::CryptoData& AsymmetricEncryption::CryptoData::operator=(
+                                                                    const std::vector<uint8_t>& data)
+{
+    _data=data;
+    return *this;
+}
+
+AsymmetricEncryption::CryptoData& AsymmetricEncryption::CryptoData::operator=(
+                                                                    const std::string& data)
+{
+    _data=std::vector<uint8_t>(data.begin(), data.end());
+    return *this;
+}
+
+std::string AsymmetricEncryption::CryptoData::toString() const {
+    return std::string(_data.begin(), _data.end());
+}
+
+std::vector<uint8_t> AsymmetricEncryption::CryptoData::toByteArray() const {
+    return _data;
+}
+
+std::string AsymmetricEncryption::CryptoData::toHex() const {
+    return utility::toHex(_data);
+}
+
+std::ostream& operator<<(std::ostream& os, const AsymmetricEncryption::CryptoData& data)
+{
+    os << data.toString();
+    return os;
+}
+
+std::vector<uint8_t> AsymmetricEncryption::encrypt(AsymmetricPublicKey key,
+                                                   const RSAPadding &pad,
+                                                   const CryptoData& message)
+{
+    SSL_RSA_OAEP_LABEL_Ptr label_copy{nullptr};
+    SSL_RSA_ENCRYPTION_DATA_Ptr encryptedMessage{nullptr};
+    size_t encryptedMessageLen{0};
+    int maxSize{0};
+
+    try {
+        const auto keyCtx = _EVP_PKEY_CTX_new(key.internal());
+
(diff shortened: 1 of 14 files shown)


src/mococrw/asymmetric_encryption.h
```diff
@@ -0,0 +1,104 @@
...
+public:
+    /**
+     * @brief CryptoData
+     * 
+     * This helper class defines the input and output data type used for encryption and decryption.
+     * Currently this class abstracts two underlying types (std::string and std::vector<unit_8t>).
+     * The user is free to call the encrypt and decrypt functionalities with either of the available
+     * underlying types.
+     */
+    class CryptoData
```
vkrikun: I like the idea of having reach collection of types. I wounder if we should use something like CryptoData widely across the library.

src/mococrw/asymmetric_encryption.h
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * This class is used to perform RSA asymmetric encryption and decryption
+ */
+class AsymmetricEncryption
+{
+public:
+    AsymmetricEncryption() = delete;
+
+private:
+    std::vector<uint8_t> message;
```
vkrikun: Unused?

src/asymmetric_encryption.cpp
```diff
@@ -0,0 +1,242 @@
...
+    try {
+        const auto keyCtx = _EVP_PKEY_CTX_new(key.internal());
+
+        const auto paddingMode = pad.getPadding();
+
+        if (!keyCtx.get()) {
+            throw MoCOCrWException("Encryption context is empty");
+        }
+
+        switch(paddingMode) {
```
vkrikun: What if you move this logic into RSAPadding subclasses? Instead of asking for a padding mode and casting, just ask for a maximum supported message length.

src/asymmetric_encryption.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+        /* Validate message size */
+        if (static_cast<int>(message.toByteArray().size()) > maxSize) {
+            throw MoCOCrWException("Message too long for RSA key size");
+        }
+        _CRYPTO_malloc_init();
+        _EVP_PKEY_encrypt_init(keyCtx.get());
+        _EVP_PKEY_CTX_set_rsa_padding(keyCtx.get(), static_cast<int>(paddingMode));
+
+        if (RSAPaddingMode::OAEP == paddingMode) {
```
vkrikun: Again, instead of requesting a type of padding, switching and casting, couldn't have something like `padding.configureContext(keyCtx)`

src/asymmetric_encryption.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+                /* Release ownership from the unique_ptr since the function above takes ownership of
+                 * the label pointer unless it throws an exception. Unique_ptr should only keep
+                 * ownership if an exception is thrown by _EVP_PKEY_CTX_set_rsa_oaep_label.*/
+                label_copy.release();
+            }
+        }
+
+        /* First call to determine the buffer length */
+        _EVP_PKEY_encrypt(keyCtx.get(),
```
vkrikun: If you move the definition of `encryptedMessageLen` from the top of the function here, you wouldn't even need to explain why `_encrypt()` is called twice. It is a good practice to define variables right before their first usage.

src/asymmetric_encryption.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+        /* Allocate memory for the buffer, based on the size returned by _EVP_PKEY_encrypt */
+        encryptedMessage.reset(static_cast<unsigned char*>(
+                                        _OPENSSL_malloc(static_cast<int>(encryptedMessageLen))));
+
+        /* Second call to perform the actual encryption */
+        _EVP_PKEY_encrypt(keyCtx.get(),
+                          encryptedMessage.get(),
+                          &encryptedMessageLen,
+                          reinterpret_cast<const unsigned char *>(&message.toByteArray()[0]),
```
vkrikun: message.data()

src/asymmetric_encryption.cpp
```diff
@@ -0,0 +1,242 @@
...
+    try {
+        auto keyCtx = _EVP_PKEY_CTX_new(key.internal());
+        const auto paddingMode = pad.getPadding();
+
+        _CRYPTO_malloc_init();
+
+        _EVP_PKEY_decrypt_init(keyCtx.get());
+        _EVP_PKEY_CTX_set_rsa_padding(keyCtx.get(), static_cast<int>(paddingMode));
+
+        if(openssl::RSAPaddingMode::OAEP == paddingMode) {
```
vkrikun: duplication from `encrypt()`? See my proposal to refactor it into a method of  RSAPadding`.

src/mococrw/asymmetric_encryption.h
```diff
@@ -0,0 +1,104 @@
...
+     * @returns The decrypted message
+     * @throw MoCOCrWException if the decryption fails.
+     */
+    static CryptoData decrypt(AsymmetricPrivateKey key,
+                              const RSAPadding& pad,
+                              const std::vector<uint8_t>& message);
+
+
+private:
+    static const int c_pkcsMaxSizeSubtract = 11;
```
vkrikun: Property of `PKCSPadding`, not `AsymmetricEncryption`

tests/unit/test_asymencryption.cpp
```diff
@@ -0,0 +1,1004 @@
...
+        "ZqdLbBnSE7a46ZB4BPFLycuORnozP5HcmevMjSDq/pFIDLM5a1SxQsBXWRbtncAZ\n"
+        "qm+/DAbtE/0SXS1rwNuUP8nhVaM+S/4x9eBXD7kCgYAQmsS/L5ZZEDTVtqJhiyGj\n"
+        "tnpq6kJKVSsa3WItN47HCE0nuPpjtH3Df/w8CQg2AUA/IjvYLcpTYgJlYt9mm+bF\n"
+        "ytMgYE9OF7bDJzqfAqolINa5KQIXZhmdE5vdDWogxZzy0SsaYCo/+0pR3Q0PCuQa\n"
+        "UwMtJkDrJSjst498ZuyGpg==\n"
+        "-----END PRIVATE KEY-----\n"
+    }
+};
+
+/// \brief Data set used to test successful (nominal) use-cases, both encryption and decryption
```
vkrikun: Value-Parameterized Tests from gtest might be a more elegant solution to performing the same test with different inputs.

tests/unit/test_asymencryption.cpp
```diff
@@ -0,0 +1,1004 @@
...
+}
+
+/**
+ * @brief Tests the CryptoData class
+ * The following use-cases are covered:
+ * - constructors
+ * - assignment operators
+ * - operator<<
+ */
+TEST_F(AsymmetricEncryptionTest, testCryptoData)
```
vkrikun: It may be moved into a separate test file as it has nothing to do with symmetric encryption.

src/asymmetric_encryption.cpp
fmonteiro93: I miss-understood your comment original. Changed to use data().

PR #13, opened 2018-11-26
Added support to create ECC asymmetric keys

fmonteiro93:
- Extended AsymmetricPublicKey and AsymmetricKeyPair to support
ECC keys
- Updated unit tests
- Added guards to functionality that currently only supports RSA
- Fixed issue found on the OEAPpadding mode class, used for the
  Asymmetric encryption functionality

### DIFF
--- a/src/asymmetric_encryption.cpp
+++ b/src/asymmetric_encryption.cpp
@@ -82,6 +82,10 @@ std::vector<uint8_t> AsymmetricEncryption::encrypt(AsymmetricPublicKey key,
     SSL_RSA_ENCRYPTION_DATA_Ptr encryptedMessage{nullptr};
     size_t encryptedMessageLen{0};
 
+    if(key.getType() != AsymmetricKey::KeyTypes::RSA){
+        throw mococrw::MoCOCrWException("Asymmetric encryption only supports RSA keys");
+    }
+
     try {
         const auto paddingMode = pad.getPadding();
         if (paddingMode != RSAPaddingMode::PKCS1  &&
@@ -95,7 +99,12 @@ std::vector<uint8_t> AsymmetricEncryption::encrypt(AsymmetricPublicKey key,
             throw MoCOCrWException("Encryption context is empty");
         }
 
-        int maxSize{pad.getDataMaxSize(key)};
+        int maxSize{pad.getDataBlockSize(key)};
+        /* Validate message size when not using padding*/
+        if (pad.getPadding() == RSAPaddingMode::NONE &&
+            maxSize != static_cast<int>(message.toByteArray().size())){
+            throw MoCOCrWException("Message size is different from the key size");
+        }
         /* Validate message size */
         if (static_cast<int>(message.toByteArray().size()) > maxSize) {
             throw MoCOCrWException("Message too long for RSA key size");
@@ -139,6 +148,10 @@ AsymmetricEncryption::CryptoData AsymmetricEncryption::decrypt(AsymmetricPrivate
 {
     size_t decryptedMessageLen{0};
     SSL_RSA_ENCRYPTION_DATA_Ptr decryptedMessage{nullptr};
+
+    if (key.getType() != AsymmetricKey::KeyTypes::RSA){
+        throw MoCOCrWException("Asymmetric decryption is only supported for RSA Keys");
+    }
 
     try {
         const auto paddingMode = pad.getPadding();
--- a/src/key.cpp
+++ b/src/key.cpp
@@ -22,6 +22,7 @@
 
 #include "mococrw/bio.h"
 #include "mococrw/key.h"
+#include "mococrw/error.h"
 
 #include "mococrw/openssl_wrap.h"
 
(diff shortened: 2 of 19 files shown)


src/asymmetric_encryption.cpp
```diff
@@ -96,6 +100,11 @@ std::vector<uint8_t> AsymmetricEncryption::encrypt(AsymmetricPublicKey key,
         }
 
         int maxSize{pad.getDataMaxSize(key)};
+        /* Validate message size when not using padding*/
```
garw: Please split this into a separate commit. I would like to have one commit per feature.

src/mococrw/key.h
```diff
@@ -40,24 +40,13 @@ class AsymmetricKey
...
-    void _createSpecFromPKEY() { switch(getType()) .... }
-
-#endif
+    /**
+     * Supported asymmetric key types.
+     */
+    enum class KeyTypes : int { RSA = EVP_PKEY_RSA, ECC = EVP_PKEY_EC };
 
+    KeyTypes getType() const  { return static_cast<KeyTypes>(openssl::_EVP_PKEY_type(_key.get())); }
+    openssl::ellipticCurveNid getCurve() const;
```
garw: I have a bad gut feeling about methods that only work sometimes depending on the state of the object. I would like to centralize this somehow. Could we have a method instead to get the KeySpec that is then the respective keyspec depending on the type of key and contains all the context-dependendant information? 

I have the feeling that this is a cleaner general interface for the future.

src/mococrw/key.h
```diff
@@ -77,12 +66,49 @@ class AsymmetricPublicKey
...
+
+    /**
+     * Gets the size of the Asymmetric key in bytes
+     * @return the size of the key in bytes
+     */
+    int getKeySize() const { return EVP_PKEY_bits(_key.internal().get());}
+
     inline bool operator==(const AsymmetricPublicKey &rhs) const
     {
         return openssl::_EVP_PKEY_cmp(internal(), rhs.internal());
```
garw: Did you verify that this still works for ECC keys?

src/mococrw/padding_mode.h
```diff
@@ -46,7 +47,7 @@ class RSAPadding
      * @param key RSA public key that will be used for encryption
      * @return the maximum size of the data that can be encrypted in bytes.
      */
-    virtual int getDataMaxSize(const AsymmetricPublicKey& key) const = 0;
+    virtual int getDataMaxSize(AsymmetricPublicKey& key) const = 0;
```
garw: Why? This should not be a non-const. Otherwise, please use pointers. I don't want non-const refs in parameters because you don't see this when reading code.

src/mococrw/padding_mode.h
```diff
@@ -77,9 +78,9 @@ class NoPadding: public RSAPadding {
     * @param key RSA public key that will be used for encryption
     * @return the maximum size of the data that can be encrypted in bytes.
     */
-    int getDataMaxSize(const AsymmetricPublicKey& key) const override
+    int getDataMaxSize(AsymmetricPublicKey& key) const override
     {
-        return openssl::_RSA_size(key.internal()->pkey.rsa);
+        return key.getKeySize()/8;
```
garw: Maybe a general block size function? It is unclear if the data block size always is exactly the key size?

src/mococrw/padding_mode.h
```diff
@@ -224,7 +225,7 @@ class PSSPadding: public RSAPadding {
      * @param Unused
      * @return -1.
      */
-    int getDataMaxSize(const AsymmetricPublicKey& key) const override
+    int getDataMaxSize(AsymmetricPublicKey& key) const override
     {
         /*Ignore unused parameter*/
         std::ignore = key;
```
garw: side remark. In modern C++, leaving the name of the parameter empty suffices to indicate that you know that the parameter will be unused.

src/openssl_wrap.cpp
```diff
@@ -202,6 +202,57 @@ void _EVP_PKEY_CTX_set_rsa_keygen_bits(EVP_PKEY_CTX* ctx, int mbits)
...
 
+void _EVP_PKEY_paramgen_init(EVP_PKEY_CTX *ctx)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_EVP_PKEY_paramgen_init, ctx);
+}
+
+SSL_EVP_PKEY_Ptr _EVP_PKEY_paramgen(EVP_PKEY_CTX *ctx)
+{
+    EVP_PKEY* ptr{nullptr};
+    try {
```
garw: This construct looks clumsy. Is there really a situation where openssl will instantiate the ptr but then return an error? maybe we can come up with a more elegant solution?

src/openssl_wrap.cpp
```diff
@@ -202,6 +202,57 @@ void _EVP_PKEY_CTX_set_rsa_keygen_bits(EVP_PKEY_CTX* ctx, int mbits)
...
+
+void _EVP_PKEY_CTX_set_ec_param_enc(EVP_PKEY_CTX *ctx, int param_enc)
+{
+    OpensslCallIsNonNegative::callChecked(lib::OpenSSLLib::SSL_EVP_PKEY_CTX_set_ec_param_enc,
+                                          ctx, param_enc);
+}
+
+const EC_GROUP* _EC_KEY_get0_group(const EC_KEY *key)
+{
+    return lib::OpenSSLLib::SSL_EC_KEY_get0_group(key);
```
garw: callPtr?

src/asymmetric_encryption.cpp
fmonteiro93: Done.

src/key.cpp
```diff
@@ -84,4 +102,47 @@ AsymmetricKey RSASpec::generate() const
...
+   return AsymmetricKey{std::move(pkey)};
+}
+
+std::unique_ptr<AsymmetricKey::Spec> AsymmetricKey::getKeySpec() const
+{
+    if(getType() == KeyTypes::ECC) {
+        openssl::ellipticCurveNid keyNid;
+        try {
+            const EC_GROUP *group = _EC_KEY_get0_group(_key->pkey.ec);
+            std::cout << "NID: " << _EC_GROUP_get_curve_name(group) << std::endl;
```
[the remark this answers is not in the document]
fmonteiro93: Done.

fmonteiro93:
Added unit test to check if creating PEMs also work when the key was initially load from PEM and not generated by the library.

PR #15, opened 2018-11-30
 Added Support to ECC certificates

fmonteiro93:
- Removed unnecessary guards to allow ECC keys to be used in addition
to RSA keys. No other change was necessary in order to support ECC certificates.
- The CA unit tests were extended to feature ECC certificates
- Extended CSR unit tests to use ECC keys
- Extended X509/verification tests to use ECC key certificates


tests/unit/test_ca.cpp
```diff
@@ -222,70 +239,162 @@ TEST_F(CATest, testIterateOnExtensionMap)
...
+    EXPECT_EQ(_signParams, _rsaCa->getSignParams());
+}
+
+TEST_F(CATest, testECCCAContents)
+{
+    EXPECT_EQ(_rootEccCert->toPEM(), _eccCa->getRootCertificate().toPEM());
+    EXPECT_EQ(_signParams, _eccCa->getSignParams());
+}
+
+TEST_F(CATest, testSignedCSRHasCorrectFieldsRSA)
```
garw: Maybe we could do this with an INSTANCIATE_TESTCASE as well?
Also for a lot of the above test cases.

fmonteiro93: Agreed. Reworked the test_ca to use parametrized tests.

fmonteiro93:
I agree that it would be nice to have as little replicated tests as possible. But reworking the x509 and verification tests to value parametrized would represent a large amount of work without adding much value right now. Maybe we can do in the future, out of the scope of this PR.

PR #18, opened 2019-02-04
Refactor of asymmetric crypto functions and add support to ECC signatures

fmonteiro93:
- Added crypto context class
- Added prepareContext to padding mode class
- Refactor of asymmetric encryption facility 
- Added support for ECC signatures
- Refactor of SignatureUtils class
- Updated unit tests


src/mococrw/signature.h
```diff
@@ -34,136 +35,36 @@ class SignatureUtils
     SignatureUtils() = delete;
```
garw: This is an all-static member class once again.

Maybe we could attach this to the CryptoContext and not have Encrypt or Decrypt Context but a PubkeyContext and Privkey Context?

src/mococrw/signature.h
```diff
@@ -34,136 +35,36 @@ class SignatureUtils
...
-     * @param signature The signature to be verified
-     * @param messageDigest The message digest the signature is verified with
-     * @throw MoCOCrWException if the verification fails.
-     */
-    static void verify(const X509Certificate &certificate,
-                       const RSAPadding &padding,
-                       const std::vector<uint8_t> signature,
-                       const std::vector<uint8_t> messageDigest);
-
+    static void verify(AsymmetricVerifyCtx &ctx,
```
garw: Why is this here? For Encrypt/Decrypt this is encapsulated in the params class or the context?

src/signature.cpp
```diff
@@ -26,102 +26,51 @@ namespace mococrw {
...
         _EVP_PKEY_sign(keyCtx.get(),
                        nullptr,
-                       &siglen,
+                       &sigLen,
                        reinterpret_cast<const unsigned char *>(messageDigest.data()),
                        messageDigest.size());
 
-        _CRYPTO_malloc_init();
-        sig.reset(static_cast<unsigned char *>(_OPENSSL_malloc(siglen)));
+        signatureData.reset(static_cast<unsigned char *>(_OPENSSL_malloc(sigLen)));
```
garw: is there a reason why this memory needs to be managed by openssl? Otherwise - just use std::vector as buffer.

src/signature.cpp
```diff
@@ -26,102 +26,51 @@ namespace mococrw {
...
-    if(publicKey.getType() != AsymmetricKey::KeyTypes::RSA){
-        throw MoCOCrWException("Functionality is only supported for RSA keys");
-    }
-
     try {
+        auto keyCtx = _EVP_PKEY_CTX_new(ctx.getKey().internal());
         _EVP_PKEY_verify_init(keyCtx.get());
-        setUpContext(keyCtx.get(), padding);
+
+        setupOpenSSLCtx(ctx, keyCtx, RSAPadding::OperationTypes::Verify);
```
garw: Operation type could be more global? Does not seem to be specific to RSAPadding.

src/signature.cpp
```diff
@@ -134,66 +83,22 @@ void SignatureUtils::verify(AsymmetricPublicKey &publicKey,
...
+        if (ctx.getRsaPaddingMode().get() == nullptr){
+            throw mococrw::MoCOCrWException("When using a RSA key the RSAPadding context "
+                                            "member needs to be set");
+        }
+        ctx.getRsaPaddingMode()->prepareOpenSSLContext(keyCtx, op);
     }
-    default:
-        throw MoCOCrWException("Padding mode not supported");
+    else if (ctx.getKeyType() == AsymmetricKey::KeyTypes::ECC){
+        _EVP_PKEY_CTX_set_signature_md(keyCtx.get(),
```
garw: it seems that all this here is known to the context. so why not move it there?

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -0,0 +1,255 @@
...
+     * @param rsaPaddingMode padding mode to be used
+     */
+    void setRsaPaddingMode (std::shared_ptr<RSAPadding> const& rsaPaddingMode)
+    { _paddingMode = rsaPaddingMode; }
+
+    /**
+     * Gets the RSA padding mode object
+     * @return RSAPadding mode
+     */
+    std::shared_ptr<RSAPadding> getRsaPaddingMode() const
```
garw: see above comment on const correctness. While this is syntactically correct, I think this gives the wrong impression.

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -0,0 +1,255 @@
...
+
+    virtual ~AsymmetricCryptoCtx() = default;
+
+    virtual AsymmetricKey::KeyTypes getKeyType() const = 0;
+
+    /**
+     * Sets the RSA padding mode Object
+     * @param rsaPaddingMode padding mode to be used
+     */
+    void setRsaPaddingMode (std::shared_ptr<RSAPadding> const& rsaPaddingMode)
```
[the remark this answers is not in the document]
fmonteiro93: Agreed. I was't 100% sure about this implementation either.

src/mococrw/asymmetric_crypto_ctx.h
fmonteiro93: Removed the method. It's unused and to be honest it doens't serve any specific purpose.

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -0,0 +1,255 @@
...
+     */
+    openssl::DigestTypes getEccHashingFunction() const
+    {
+        if (getKeyType() != AsymmetricKey::KeyTypes::ECC)
+            throw MoCOCrWException{"Hashing function only valid when using an ECC key"};
+        return  _eccMd;
+    }
+
+protected:
+    void setupSignatureOpenSSLCtx(openssl::SSL_EVP_PKEY_CTX_Ptr& keyCtx,
```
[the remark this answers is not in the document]
fmonteiro93: Done.

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * Builder method with private key
+     * @param privateKey private key to be used by the context
+     * @return Asymmetric Private key context
+     */
+    AsymmetricPrivkeyCtx build(const AsymmetricPrivateKey& privateKey);
+
+private:
+    void buildHelper(AsymmetricCryptoCtx& ctx);
```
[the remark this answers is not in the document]
fmonteiro93: Done.

PR #21, opened 2019-02-11
Boost include paths fix for MacOS Mojave 10.14

Fleckol:


CMakeLists.txt
```diff
@@ -62,6 +62,9 @@ if(OPENSSL_FOUND)
 endif()
 
 find_package(Boost REQUIRED)
+if(Boost_FOUND)
+    include_directories (SYSTEM ${Boost_INCLUDE_DIR})
```
neverpanic: What's the error you see without this? Usually, the `FindBoost.cmake` module exports a `Boost::boost` target that should already pull this into the build where required (see https://cmake.org/cmake/help/latest/module/FindBoost.html). Maybe we're actually missing a `target_link_libraries(${target} <PRIVATE|PUBLIC> Boost::boost)` somewhere instead?

Fleckol: I get the following:
MoCOCrW/src/mococrw/openssl_wrap.h:25:10: fatal error: 'boost/format.hpp' file not found
#include "boost/format.hpp"

my cmake command:
cmake -DOPENSSL_ROOT_DIR=/usr/local/Cellar/openssl/1.0.2q/ -DOPENSSL_LIBRARIES=/usr/local/Cellar/openssl/1.0.2q/lib -DBUILD_TESTING=OFF -DBOOST_ROOT=/usr/local/Cellar/boost/1.65.1 ..

i haven't checked it with target_link_libraries yet

neverpanic: Yeah, that actually looks like the target you're building doesn't actually use `target_link_libraries(${target} PUBLIC Boost::boost)` and that should be fixed.

Fleckol: BTW. it used to work on previous releases of XCODE... however since it is version 10, it stopped working for me.

Fleckol: i jus tested it with in the main cmakelist file:

find_package(Boost REQUIRED)
include_directories( ${Boost_INCLUDE_DIRS} ) 

This also works.

Hence i will update the pull request

neverpanic: Yes, that will work, but it's not modern CMake. We should strive to use modern CMake where possible.

Fleckol: What about target_include_directories? that works as well and i think this is the way how it should be done

neverpanic: The modern way to do things is using targets, because those encapsulate all the details and relieve you from caring whether a library needs include directories, compiler flags, preprocessor flags, linker paths, linker flags and libraries required to link.

That way you can put all your dependencies in different targets and not care in which properties you need their various variables.

neverpanic: Can you try https://github.com/neverpanic/MoCOCrW/tree/cal-fix-cmake?

Fleckol: Scanning dependencies of target mococrw
[  5%] Building CXX object src/CMakeFiles/mococrw.dir/asn1time.cpp.o
[ 10%] Building CXX object src/CMakeFiles/mococrw.dir/asymmetric_encryption.cpp.o
[ 15%] Building CXX object src/CMakeFiles/mococrw.dir/basic_constraints.cpp.o
[ 20%] Building CXX object src/CMakeFiles/mococrw.dir/bio.cpp.o
[ 25%] Building CXX object src/CMakeFiles/mococrw.dir/ca.cpp.o
[ 30%] Building CXX object src/CMakeFiles/mococrw.dir/crl.cpp.o
[ 35%] Building CXX object src/CMakeFiles/mococrw.dir/csr.cpp.o
[ 40%] Building CXX object src/CMakeFiles/mococrw.dir/distinguished_name.cpp.o
[ 45%] Building CXX object src/CMakeFiles/mococrw.dir/hash.cpp.o
[ 50%] Building CXX object src/CMakeFiles/mococrw.dir/key.cpp.o
[ 55%] Building CXX object src/CMakeFiles/mococrw.dir/key_usage.cpp.o
[ 60%] Building CXX object src/CMakeFiles/mococrw.dir/openssl_lib.cpp.o
[ 65%] Building CXX object src/CMakeFiles/mococrw.dir/openssl_wrap.cpp.o
[ 70%] Building CXX object src/CMakeFiles/mococrw.dir/signature.cpp.o
[ 75%] Building CXX object src/CMakeFiles/mococrw.dir/subject_key_identifier.cpp.o
[ 80%] Building CXX object src/CMakeFiles/mococrw.dir/util.cpp.o
[ 85%] Building CXX object src/CMakeFiles/mococrw.dir/x509.cpp.o
[ 90%] Linking CXX shared library libmococrw.dylib
Undefined symbols for architecture x86_64:
  "_ASN1_TIME_diff", referenced from:
      mococrw::openssl::lib::OpenSSLLib::SSL_ASN1_TIME_diff(int*, int*, asn1_string_st const*, asn1_string_st const*) in openssl_lib.cpp.o
     (maybe you meant: __ZN7mococrw7openssl3lib10OpenSSLLib18SSL_ASN1_TIME_diffEPiS3_PK14asn1_string_stS6_, __ZN7mococrw7openssl15_ASN1_TIME_diffEPiS1_PK14asn1_string_stS4_ )
ld: symbol(s) not found for architecture x86_64

neverpanic: Interesting, but that looks more like an internal linking failure of MoCOCrW. I'll see what I can do.

Fleckol: Why not using:

target_include_directories?

that worked for me, although with scope PUBLIC

neverpanic: Because it doesn't work correctly for cross-compilation. If your include directory while building is `/home/builder/$project/build/tmp/work/$arch/mococrw-$version/recipe-sysroot/usr/include/boost` and you use that as `PUBLIC target_include_directory`, it will get included in the CMake target exported by MoCOCrW. Obviously this path will not be correct for users, e.g. in an SDK.

That can be worked around by using `target_include_directory(PRIVATE ...)` or `target_include_directory(PUBLIC $<BUILD_INTERFACE:...>)`, but that will just mean that you leave users of this library on their own with finding appropriate boost headers.

I can recommend https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/, which gives a good overview of why and how this should be done with targets.

neverpanic: Can you give https://github.com/neverpanic/MoCOCrW/tree/cal-fix-cmake one more try? It builds fine on my Mac, but so did the last version.

Fleckol: The last version worked.
Thanks a lot

neverpanic: Right, see #22.

flmei: Since we merged [cal-fix-cmake](https://github.com/neverpanic/MoCOCrW/tree/cal-fix-cmake), do we still need this PR?

neverpanic:
Now obsolete due to #22. Closing.

PR #22, opened 2019-02-12
CMake: Use CMake targets to link against Boost and OpenSSL

neverpanic:
This fixes the build in situations where Boost or OpenSSL headers/libraries are not in the default search path.

PR #23, opened 2019-02-12
Check OpenSSL version in CMakeLists

vkrikun:
Since we wrap specific version of OpenSSL it is worth checking which on
is installed. If user has a newer (or older) version of OpenSSL than we
support it may result into compilation errors/warning or misbehaviours.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ if(APPLE)
     find_package(OpenSSL REQUIRED)
 else()
     pkg_search_module(OPENSSL REQUIRED openssl)
+    pkg_check_modules(OPENSSL openssl<=1.1)
 endif()
 
 if(OPENSSL_FOUND)


CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ if(APPLE)
     find_package(OpenSSL REQUIRED)
 else()
     pkg_search_module(OPENSSL REQUIRED openssl)
```
neverpanic: You can do that in the same line, no need to call `pkg_check_modules` a second time. Also, look at #22, which touches this code anyway.

vkrikun: You can do it in one line, but in that case, the output from CMake in case of version mismatch is misleading. It complains that package was not found and does not say why (version mismatch).

neverpanic: But if you do it in a separate call and do not pass the `REQUIRED` flag, it will just print a message during configure, but not actually fail the configure phase…

neverpanic:
This now has a conflict due to #22. You can do the same version check by adding a version number to the `find_package(OpenSSL)` statement, though. The `>=` comparison is then implied.

neverpanic:
Let's revisit this later.

PR #24, opened 2019-02-19
Added user id field to distinguished name.

davidpereiracriticaltechworkscom:


garw:
Looks also good to me

PR #28, opened 2019-07-03
Tests: Fix use after free in KeyHandlingTests

blenk92:
In testGetKeySpec a raw pointer is extracted from a temporary unique_ptr. After extracting the pointer, the temporary unique_ptr is destroyed (and the object it points to is freed). That is, the extracted pointer points to a freed memory location.


neverpanic:
Please also cherry-pick this to the openssl1.1 branch.

PR #31, opened 2019-07-08
Add support for Ed448 and Ed25519

blenk92:
This patch adds support for Ed448 and Ed25519:
 - key generation
 - (eddsa) signatures


src/key.cpp
```diff
@@ -128,16 +147,21 @@ AsymmetricKey ECCSpec::generate() const {
...
-        } catch (const OpenSSLException &e) {
-            throw MoCOCrWException(e.what());
+        if(nid == EVP_PKEY_EC) {
+            try {
+                const EC_GROUP *group = _EC_KEY_get0_group(_EVP_PKEY_get0_EC_KEY(_key.get()));
+                keyNid = static_cast<openssl::ellipticCurveNid>(_EC_GROUP_get_curve_name(group));
+            } catch (const OpenSSLException &e) {
+                throw MoCOCrWException(e.what());
+            }
+        } else {
```
garw: A generic else looks potentially dangerous to me. Please check exactly what we expect here and throw if something unexpected arrives here.

src/key.cpp
```diff
@@ -105,18 +118,24 @@ AsymmetricKey RSASpec::generate() const
 
 AsymmetricKey ECCSpec::generate() const {
```
garw: If these curves are that special, we might want to create a separate Keyspec type? Let's discuss this with Clemens.

src/mococrw/key.h
```diff
@@ -43,9 +43,9 @@ class AsymmetricKey
     /**
      * Supported asymmetric key types.
      */
-    enum class KeyTypes : int { RSA = EVP_PKEY_RSA, ECC = EVP_PKEY_EC };
+    enum class KeyTypes : int { RSA, ECC, UNSUPPORTED };
```
garw: We could create a key type ECC_ED for the new curves? I am not sure about pro and con. 
For better ABI/API stability, UNSUPPORTED should get a high enum value?

src/mococrw/openssl_wrap.h
```diff
@@ -545,6 +545,7 @@ void _X509_REQ_sign_ctx(X509_REQ* req, EVP_MD_CTX* ctx);
  * or hash computations.
  */
 enum class DigestTypes {
+    NONE = NULL,
```
garw: NULL?
What is the reason for this? Looks like a workaround to me.

src/mococrw/openssl_wrap.h
```diff
@@ -1067,6 +1056,64 @@ void _EVP_PKEY_verify(EVP_PKEY_CTX *ctx,
...
+ * @throw OpenSSLException if an error occurs in the underlying OpenSSL function.
+ */
+void _EVP_DigestSign(EVP_MD_CTX* ctx,
+                     unsigned char* sigret,
+                     size_t* siglen,
+                     const unsigned char* tbs,
+                     size_t tbslen);
+
+/**
+ * Initialize the MD_CTX for signing, using a given digest-type and a given
```
garw: signing? Rather verification?

src/mococrw/openssl_wrap.h
```diff
@@ -1067,6 +1056,64 @@ void _EVP_PKEY_verify(EVP_PKEY_CTX *ctx,
...
+ * public key.
+ *
+ * @param ctx The context to be set up for signing.
```
garw: s.a.

src/mococrw/openssl_wrap.h
```diff
@@ -1067,6 +1056,64 @@ void _EVP_PKEY_verify(EVP_PKEY_CTX *ctx,
...
+ * @param ctx The context used for signing.
+ * @param sigret The buffer to write the signature to.
+ * @param siglen The length of the generated signature.
+ * @param tbs Message to be signed.
+ * @param tbslen Length of the message to be signed.
+ *
+ * @throw OpenSSLException if an error occurs in the underlying OpenSSL function.
+ */
+void _EVP_DigestSign(EVP_MD_CTX* ctx,
+                     unsigned char* sigret,
```
garw: Why do parameters have these strange short names instead of good ones?

src/mococrw/openssl_wrap.h
```diff
@@ -1067,6 +1056,64 @@ void _EVP_PKEY_verify(EVP_PKEY_CTX *ctx,
...
+ *
+ * @param ctx The context used for signing.
+ * @param sigret The signature to be verified.
+ * @param siglen The length of the the signature to be verified.
+ * @param tbs Signed message.
+ * @param tbslen Length of the signed message.
+ *
+ * @throw OpenSSLException if an error occurs in the underlying OpenSSL function.
+ */
+void _EVP_DigestVerify(EVP_MD_CTX *ctx,
```
garw: How is the information whether the signature was valid or not transported to the caller?

src/mococrw/padding_mode.h
```diff
@@ -60,6 +60,11 @@ class NoPadding: public RSAPadding {
 public:
     virtual ~NoPadding() = default;
 
+    NoPadding(openssl::DigestTypes hashingFunction = openssl::DigestTypes::SHA256)
```
garw: Why? NoPadding should not have a hash function. This is supposed to configure padding modes. NoPadding has no attributes. Looks a bit like misuse to me.

src/mococrw/util.h
```diff
@@ -30,5 +30,11 @@ template <class T>
 using SharedPtrTypeFromUniquePtr = std::shared_ptr<typename T::element_type>;
 
 std::string toHex(const std::vector<uint8_t> &data);
+
+template<typename T, typename U>
+std::unique_ptr<T> unique_ptr_cast(std::unique_ptr<U>&& ptr) {
```
garw: unique_ptr_dynamic_cast?

src/mococrw/util.h
```diff
@@ -30,5 +30,11 @@ template <class T>
...
+    return std::unique_ptr<T>(dynamic_cast<T*>(ptr.release()));
```
garw: No. This is dangerous. If dynamic_cast decides that the types are incompatible, it returns a nullptr. In this case, the objected pointed to by the U* ptr is leaked.

src/openssl_wrap.cpp
```diff
@@ -414,6 +414,8 @@ void _X509_REQ_sign_ctx(X509_REQ* req, EVP_MD_CTX* ctx)
 const EVP_MD* _getMDPtrFromDigestType(DigestTypes type)
 {
     switch (type) {
+        case DigestTypes::NONE:
```
garw: This function used to return only valid objects and otherwise throw. I think this is the more reasonable behavior than returning nullptrs if you want things to fail loudly.

src/signature.cpp
```diff
@@ -32,8 +32,16 @@ std::vector<uint8_t> SignatureUtils::create(AsymmetricPrivateKey &privateKey,
 {
     std::vector<uint8_t> messageDigest;
 
-    if(privateKey.getType() != AsymmetricKey::KeyTypes::RSA){
-        throw MoCOCrWException("Functionality is only supported for RSA keys");
+    if(privateKey.getType() != AsymmetricKey::KeyTypes::RSA) {
```
garw: I fear all of this was refactored in master and needs to merged here as well before we can do this :(

src/mococrw/openssl_wrap.h
blenk92: these are the original parameter names of the openssl functions that are wrapped inside. I don't have a strong opinion on this, but other functions that do similar things (such as EVP_PKEY_verify) use the same naming scheme. I guess if we change that, we should do that for more than just the new functions.

src/mococrw/openssl_wrap.h
blenk92: An OpenSSLException is thrown. (I added that to the comment)

src/openssl_wrap.cpp
blenk92: I this case nullptr does not indicate an error, instead it is expected by openssl to be passed if no hashing shall be used. (See e.g. https://www.openssl.org/docs/man1.1.1/man3/EVP_DigestSignInit.html)

blenk92:
This is the current status of the ed448 and ed25519 support. It turned out that the api requires some more refactoring before this can be merged. Until this is done, this PR is on hold.

DO NOT MERGE

blenk92:
Refactored the PR and rebased it onto https://github.com/bmwcarit/MoCOCrW/pull/39

src/openssl_lib.cpp
```diff
@@ -242,6 +242,15 @@ int OpenSSLLib::SSL_EVP_DigestSignInit(EVP_MD_CTX* ctx,
     return EVP_DigestSignInit(ctx, pctx, type, e, pkey);
 }
 
+int OpenSSLLib::SSL_EVP_DigestSign(EVP_MD_CTX* ctx,
```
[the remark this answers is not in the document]
blenk92: hmm functions in this file are auto-generated using the `add_function.py` script. Because the functions in here are simply calling the appropriate openssl functions, it might be a good idea to keep the openssl parameter-names here (so they match the openssl documentation).

tests/unit/openssl_lib_mock.h
```diff
@@ -208,6 +208,21 @@ class OpenSSLLibMockInterface
     virtual void SSL_EVP_MD_CTX_destroy(EVP_MD_CTX* ptr) = 0;
     virtual int SSL_EVP_DigestSignInit(
             EVP_MD_CTX* ctx, EVP_PKEY_CTX**, const EVP_MD*, ENGINE*, EVP_PKEY*) = 0;
+    virtual int SSL_EVP_DigestSign(EVP_MD_CTX* ctx,
```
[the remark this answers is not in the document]
blenk92: see below...

src/mococrw/key.h
```diff
@@ -43,9 +43,13 @@ class AsymmetricKey
     /**
      * Supported asymmetric key types.
      */
-    enum class KeyTypes : int { RSA = EVP_PKEY_RSA, ECC = EVP_PKEY_EC };
+    enum class KeyTypes : int {
+        RSA,
```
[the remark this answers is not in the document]
blenk92: To keep ABI stable we could do something like:
```
enum class KeyTypes : int {
    RSA = EVP_PKEY_RSA,
    ECC = EVP_PKEY_EC,
    ECC_ED 
};
```
But ihmo this is weird and we might have to break it in the future (or we assign e.g EVP_PKEY_ED448 to ECC_ED, which is also weird because ECC_ED represents Ed448 and Ed25519).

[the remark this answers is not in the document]
neverpanic: I think we're changing ABI at other places anyway, so let's just bump the SOVERSION and be done with it.

PR #32, opened 2019-07-09
Add attribute order feature for subject

marcoczech:
This allows users to use custom subject attribute order for X509 name.


src/distinguished_name.cpp
```diff
@@ -84,6 +90,11 @@ DistinguishedName DistinguishedName::fromX509Name(X509_NAME *x509)
...
 }
 
+const std::string& DistinguishedName::getSubjectAttributeByNIDAsString(const ASN1_NID id) const
+{
+    return _getSubjectAttributeByNID(id).name;
+}
+
 auto _createTuple(const DistinguishedName &dn)
 {
     return std::tie(dn.commonName(),
```
garw: This seems to be broken since the introduction of givenName(). Should be added here as well.

tests/unit/test_distinguished_name.cpp
```diff
@@ -101,3 +101,32 @@ TEST_F(DistinguishedNameTest, testThatDistinguishedNameIsPopulatedCorrectly)
...
+        x509Name.get(), openssl::ASN1_NID::OrganizationalUnitName)[0], 3);
+    ASSERT_EQ(_X509_NAME_get_index_by_NID(
+        x509Name.get(), openssl::ASN1_NID::LocalityName)[0], 4);
+    ASSERT_EQ(_X509_NAME_get_index_by_NID(
+        x509Name.get(), openssl::ASN1_NID::SerialNumber)[0], 5);
+    ASSERT_EQ(_X509_NAME_get_index_by_NID(
+        x509Name.get(), openssl::ASN1_NID::OrganizationName)[0], 6);
+    ASSERT_EQ(_X509_NAME_get_index_by_NID(
+        x509Name.get(), openssl::ASN1_NID::StateOrProvinceName)[0], 7);
+}
```
garw: There should be a test of the order on the old builder format is unchanged. I think there is no test right now that tests anything w.r.t. the order of elements.

src/mococrw/distinguished_name.h
```diff
@@ -80,19 +82,17 @@ class DistinguishedName
...
-    std::string _organizationName;
-    std::string _pkcs9EmailAddress;
-    std::string _serialNumber;
-    std::string _givenName;
-    std::string _userId;
+private:    
+    const SubjectAttribute& _getSubjectAttributeByNID(const openssl::ASN1_NID id) const;
+    void _addString(openssl::SSL_X509_NAME_Ptr &x509Name, const SubjectAttribute& unit) const;
+    std::deque<SubjectAttribute> _subjectAttributes;
+    SubjectAttribute _emptyAttribute;
```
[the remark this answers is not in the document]
marcoczech: So that we can use const reference and dont have to return a copy of std::string every time.

src/distinguished_name.cpp
marcoczech: We thought it is on purpose left out. I can added no problem.

src/mococrw/distinguished_name.h
```diff
@@ -40,30 +41,31 @@ namespace mococrw
...
-     * Definition is below.
-     *
-     */
     class Builder;
+    class CustomOrderBuilder;
+
+    struct SubjectAttribute {
+        openssl::ASN1_NID id;
+        std::string       name;
+    };
```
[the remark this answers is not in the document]
marcoczech: User cannot read the list of attributes otherwise.

src/distinguished_name.cpp
```diff
@@ -84,6 +90,11 @@ DistinguishedName DistinguishedName::fromX509Name(X509_NAME *x509)
     return builder.build();
 }
 
+const std::string& DistinguishedName::getSubjectAttributeByNIDAsString(const ASN1_NID id) const
```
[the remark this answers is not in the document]
marcoczech: ?

src/mococrw/distinguished_name.h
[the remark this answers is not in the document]
garw: Sorry but this is not good design. Default objects are kind of an anti-pattern and can lead to a multitude of issues. Moving strings around and having empty strings is expensive enough so that you don't need to get into this kind of design nightmares.
Moreover, there is return-value-optimization.  So just return values instead of const refs. What is more, if the attribute is empty, which means that it doesn't exist, you can just return an optional<> none.

src/distinguished_name.cpp
```diff
@@ -94,12 +105,39 @@ auto _createTuple(const DistinguishedName &dn)
...
 {
     return _createTuple(*this) == _createTuple(other);
 }
 
+void DistinguishedName::_addAttributeByIndex(const ASN1_NID id, const std::string &attribute, const size_t index)
+{
+    if (index >= _subjectAttributes.size()) {
+        _subjectAttributes.resize(index+1);
+    }
+    _subjectAttributes[index] = std::move(SubjectAttribute{id, attribute});
```
garw: There is no reason to invoke std::move on objects that are already an r-value

src/distinguished_name.cpp
```diff
@@ -71,19 +77,24 @@ std::string _getEntryByNIDAsString(X509_NAME *x509, ASN1_NID nid)
...
-    builder.countryName(_getEntryByNIDAsString(x509, ASN1_NID::CountryName));
-    builder.localityName(_getEntryByNIDAsString(x509, ASN1_NID::LocalityName));
-    builder.stateOrProvinceName(_getEntryByNIDAsString(x509, ASN1_NID::StateOrProvinceName));
-    builder.organizationalUnitName(_getEntryByNIDAsString(x509, ASN1_NID::OrganizationalUnitName));
-    builder.organizationName(_getEntryByNIDAsString(x509, ASN1_NID::OrganizationName));
-    builder.pkcs9EmailAddress(_getEntryByNIDAsString(x509, ASN1_NID::Pkcs9EmailAddress));
-    builder.serialNumber(_getEntryByNIDAsString(x509, ASN1_NID::SerialNumber));
-    builder.givenName(_getEntryByNIDAsString(x509, ASN1_NID::GivenName));
-    builder.userId(_getEntryByNIDAsString(x509, ASN1_NID::UserId));
+    builder.addAttributeByIndex(ASN1_NID::CommonName, x509);
```
garw: This whole process looks weird to me. We search for all of the attributes and then try to place them in the right location. What happens if there is an item in the Subject of a type that is not supported? It will end up in the list with an uninitialized SubjectAttribute object. That sounds potentially dangerous. It rather looks to me like the reasonable strategy here would be to iterate over the DN and add them 1:1 to the internal data vector.

marcoczech: I dont see how I can iterate over the X509 attributes.

marcoczech: This behaves like before. When the attribute cannot be found we leave it empty.

src/mococrw/distinguished_name.h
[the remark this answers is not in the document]
marcoczech: DONE

src/distinguished_name.cpp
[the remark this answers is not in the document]
marcoczech: Return an empty string in case it cannot be found. The old code detected invalid data like that. We rather stick to it.

tests/unit/test_distinguished_name.cpp
```diff
@@ -101,3 +101,32 @@ TEST_F(DistinguishedNameTest, testThatDistinguishedNameIsPopulatedCorrectly)
...
+    customOrderBuilder.commonName("ImATeapot")
+            .countryName("DE")
+            .pkcs9EmailAddress("support@example.com")
+            .organizationalUnitName("Linux Support")
+            .localityName("oben")
+            .serialNumber("08E36DD501941432358AFE8256BC6EFD")
+            .organizationName("Linux AG")
+            .stateOrProvinceName("nebenan");
+    auto dn = customOrderBuilder.build();
+    SSL_X509_NAME_Ptr x509Name = _X509_NAME_new();
```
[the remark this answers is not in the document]
marcoczech: DONE

src/distinguished_name.cpp
```diff
@@ -25,30 +24,38 @@ namespace mococrw
...
-    _addString(subject, stateOrProvinceName(), ASN1_NID::StateOrProvinceName);
-    _addString(subject, organizationalUnitName(), ASN1_NID::OrganizationalUnitName);
-    _addString(subject, organizationName(), ASN1_NID::OrganizationName);
-    _addString(subject, pkcs9EmailAddress(), ASN1_NID::Pkcs9EmailAddress);
-    _addString(subject, serialNumber(), ASN1_NID::SerialNumber);
-    _addString(subject, givenName(), ASN1_NID::GivenName);
-    _addString(subject, userId(), ASN1_NID::UserId);
+    if (_customAttributeOrderFlag) {
+        for (const auto& it : _subjectAttributes) {
+            _addString(subject, it.id);
```
garw: This should also put in the string to avoid re-searching the vector...

src/mococrw/distinguished_name.h
```diff
@@ -122,82 +115,87 @@ class DistinguishedName::Builder
...
+{
+public:
+    CustomOrderBuilder(): Builder() { _dn._customAttributeOrderFlag = true; }
+};
+
 template <class T>
 DistinguishedName::Builder &DistinguishedName::Builder::commonName(T &&name)
 {
-    _dn._commonName = std::forward<T>(name);
+    _dn._subjectAttributes.emplace_back(Attribute{openssl::ASN1_NID::CommonName, name});
```
garw: here without forward ... in the others with forward?

src/distinguished_name.cpp
```diff
@@ -25,30 +24,38 @@ namespace mococrw
...
         return;
     }
-    auto vec = std::vector<unsigned char>{str.begin(), str.end()};
+    auto vec = std::vector<unsigned char>{name.begin(), name.end()};
     _X509_NAME_add_entry_by_NID(x509Name.get(), nid, ASN1_Name_Entry_Type::ASCIIString, vec);
 }
 
 void DistinguishedName::populateX509Name(SSL_X509_NAME_Ptr &subject) const
 {
-    _addString(subject, commonName(), ASN1_NID::CommonName);
```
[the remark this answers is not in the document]
garw: Moreover, it looks to me like it still has issues w.r.t. the old order if the builder function for one entry was called more than once. In the old behavior, the latest value would have been serialized. In the new order, the first value is serialized. I guess this still needs to be fixed.

src/distinguished_name.cpp
```diff
@@ -16,39 +16,47 @@
...
 #include "mococrw/error.h"
 
+#include <tuple>
+
 namespace mococrw
 {
 using namespace openssl;
 
-void _addString(SSL_X509_NAME_Ptr &x509Name, const std::string &str, ASN1_NID nid)
+void _addString(SSL_X509_NAME_Ptr &x509Name, const std::string& name, const ASN1_NID nid)
```
neverpanic: Wouldn't it be better to refactor this into

```c++
void _addAttribute(SSL_X509_NAME_Ptr& x509Name, const DistinguishedName::Attribute& attribute)
{
    // ...
    if (attribute.name.empty()) {
        return;
    }
    std::vector<unsigned char> vec{attribute.name.begin(), attribute.name.end()};
    _X509_NAME_add_entry_by_NID(x509Name.get(), attribute.nid, ASN1_Name_Entry_Type::ASCIIString, vec);
}
```

Of course you'd then have to support a way of returning "unspecified" from whatever you'd replace `_getAttributeByNIDAsString()` with, potentially using `std::optional` or `boost::optional`. This would make the API much cleaner, IMHO, and would also allow you to replace the series of `_addString` calls below with a loop that loops over the NIDs (plus you wouldn't have to specify the NID twice per line).

src/distinguished_name.cpp
```diff
@@ -84,22 +92,32 @@ DistinguishedName DistinguishedName::fromX509Name(X509_NAME *x509)
     return builder.build();
 }
 
-auto _createTuple(const DistinguishedName &dn)
+auto _createTuple(const DistinguishedName& dn)
 {
-    return std::tie(dn.commonName(),
+    return std::make_tuple(dn.commonName(),
```
neverpanic: This is used in the comparison operator, but the comparison operator now does not take the order into account. I would expect the following behavior:

 - Comparing a non-order-aware with a non-order-aware DN: previous behavior
 - Comparing an order-aware with a non-order-aware DN (and vice-versa): Only equal if the order-aware entries are in the default order and the contents match.
 - Comparing two order-aware DNs: Only equal if the order and the contents match.

src/distinguished_name.cpp
```diff
@@ -84,22 +92,32 @@ DistinguishedName DistinguishedName::fromX509Name(X509_NAME *x509)
...
 }
 
 bool DistinguishedName::operator==(const DistinguishedName &other) const
 {
     return _createTuple(*this) == _createTuple(other);
 }
 
+std::string DistinguishedName::_getAttributeByNIDAsString(const ASN1_NID id) const
+{
+    const auto it = std::find_if(_subjectAttributes.rbegin(), _subjectAttributes.rend(), [&id] (const auto& attribute) { return attribute.id == id; } );
```
neverpanic: You've done this multiple times now, so I'd like to point out that there's very little point in declaring a ASN1_NID const or including it by reference in a lambda.

An ASN1_NID is an enum, which is probably just an int. For the same reason that you usually don't see `const int` function parameters (they are copied anyway), you don't really need a `const ASN1_NID`. Additionally, including a reference in the lambda is actually more expensive, because you copy a pointer, which is likely larger than an integer.

This is definitely a nitpick, though, because it really don't make any difference whatsoever at runtime.

src/mococrw/distinguished_name.h
```diff
@@ -40,30 +41,25 @@ namespace mococrw
...
-    const std::string &countryName() const { return _countryName; }
-    const std::string &localityName() const { return _localityName; }
-    const std::string &stateOrProvinceName() const { return _stateOrProvinceName; }
-    const std::string &organizationalUnitName() const { return _organizationalUnitName; }
-    const std::string &organizationName() const { return _organizationName; }
-    const std::string &pkcs9EmailAddress() const { return _pkcs9EmailAddress; }
-    const std::string &serialNumber() const { return _serialNumber; }
-    const std::string &givenName() const { return _givenName; }
-    const std::string &userId() const { return _userId; }
+    std::string commonName() const { return _getAttributeByNIDAsString(openssl::ASN1_NID::CommonName); }
```
neverpanic: Since this changes the ABI of the program, we should probably increase the SOVERSION of the generated library. Currently, the SOVERSION handling is tied to the MoCOCrW version number (see `src/CMakeLists.txt:5`), which is not ideal.

I would propose to change the SOVERSION to `1.0` for this change, since the SOVERSION should only change when the ABI changes, and not when we release an ABI-compatible update with a newer version number.

src/distinguished_name.cpp
marcoczech: In my opinion its good practice to make a read only input parameter also readonly by design. The compiler is then complaining in case someone modifies it within the function. Sure this function is simple but there might be more complex once. Further down the lines someone sets the value to something else accidently -> boom. Sure we can the reference passing can be removed if you really dont like it.

src/mococrw/distinguished_name.h
marcoczech: Sure

src/distinguished_name.cpp
neverpanic: I don't have a very strong opinion about this. I figured I'd mention it since you actually do change it from the previous state in a couple of places.

src/distinguished_name.cpp
marcoczech: I had it like this. But then it will break when using duplicate NIDs when using the CustomOrderBuilder. It will then always return the last entry. Now it returns the name that was actually given.

src/distinguished_name.cpp
marcoczech: Sure we can check the order using the CustomOrderBuilder. However if the DN is created from a X509 then the order is not preserved and will then also not matter for matching. Martin and I talked about it and it is quite an effort to add the missing API from openssl to iterate over the attributes and therefore record the actual order. This is can be done in a second step which is out of the scope from the pull request.

src/mococrw/distinguished_name.h
marcoczech: These are currently two variables. Do you want to hard code 1.0 at line 5 in the cmake?

src/distinguished_name.cpp
neverpanic: I don't understand how this changes things for the custom order case. You're either going to do
```c++
for (const auto& it : _subjectAttributes) {
    _addString(subject, it.name, it.id);
}
```
or
```c++
for (const auto& it: _subjectAttributes) {
    _addAttribute(subject, it);
}
```
which doesn't change behavior when your `_subjectAttributes` contains duplicate NIDs. What am I missing?

src/mococrw/distinguished_name.h
neverpanic: Yes. It was a mistake to set the library SOVERSION to the version of the library. Those two are separate numbers.

src/distinguished_name.cpp
marcoczech: Do you like it better now?

src/mococrw/distinguished_name.h
marcoczech: DONE. What about the mococrw version itself?

src/mococrw/distinguished_name.h
```diff
@@ -79,20 +75,18 @@ class DistinguishedName
...
-    std::string _serialNumber;
-    std::string _givenName;
-    std::string _userId;
+private:    
+    struct Attribute {
+        openssl::ASN1_NID id;
+        std::string       name;
+    };
+    std::string _getAttributeByNIDAsString(const openssl::ASN1_NID id) const;
+    boost::optional<Attribute> _getAttributeByNID(const openssl::ASN1_NID id) const;
```
neverpanic: This requires that you add `#include <boost/optional.hpp>`.

src/distinguished_name.cpp
```diff
@@ -84,22 +92,52 @@ DistinguishedName DistinguishedName::fromX509Name(X509_NAME *x509)
...
+                     dn.pkcs9EmailAddress(),
+                     dn.serialNumber(),
+                     dn.userId(),
+                     dn.givenName());
 }
 
 bool DistinguishedName::operator==(const DistinguishedName &other) const
 {
-    return _createTuple(*this) == _createTuple(other);
+    if (_customAttributeOrderFlag) {
```
neverpanic: This breaks commutativity of the `==` operator when comparing an ordered and an unordered DN.

Either you also take `other._customAttributeOrderFlag` into account or we only do the unordered comparison as suggested by you.

src/distinguished_name.cpp
```diff
@@ -84,22 +92,52 @@ DistinguishedName DistinguishedName::fromX509Name(X509_NAME *x509)
...
+        // order matters now
+        return std::equal(_attributes.begin(), _attributes.end(), other._attributes.begin(), other._attributes.end()
```
neverpanic: Doesn't the struct have an auto-generated equality operator that does the same thing? Also, doesn't `std::vector` have an equality operator, which would make this entire block equal to `_attributes == other._attributes`?

src/mococrw/distinguished_name.h
neverpanic: I'd leave that as-is for now. We can do a commit to increase it later and push a tag on that commit.

neverpanic:
You probably also want to open a PR that cherry-picks this into the openssl1.1 branch.

PR #35, opened 2019-07-25
Remove weird api that will be refactored soonish

blenk92:
This PR removes the asymmetric crypto api. This is done because it will be refactored soon and we want to avoid that it is already used.


neverpanic:
Unfortunately this drops the tests, which I think we should keep. Let's revert the commit instead, since we should be doing a better job of documenting API changes anyway and the change was an API & ABI break.

PR #36, opened 2019-07-25
Add title attribute to the subject.

marcoczech:


marcoczech:
Added test for the new feature.

marcoczech:
Added title to standard builder.
Added title for comparator.

PR #40, opened 2019-08-19
Refactor asymmetric crypto API further and add examples

blenk92:
This PR adds virtual interfaces to the current API, moves the specification of the hash function from the paddings to the contexts (and does some adaptions that are required with to the API). Additionally a (doxygen) example page for the asymmetric crypto interface is added.


tests/unit/test_signature.cpp
```diff
@@ -30,14 +30,18 @@ using namespace std::string_literals;
 
 using namespace mococrw;
 
-class SignatureTest : public ::testing::Test
-{
+class SignatureTestDataSet {
```
vkrikun: It does not look like a "DataSet". More like a "SignatureStrategy".

src/mococrw/padding_mode.h
```diff
@@ -25,116 +25,87 @@
...
-      * @return The hash function
-      */
-    openssl::DigestTypes getHashFunction() const;
-
-protected:
-    /**
-     * @brief Hash function to be used for signing
-     */
-    openssl::DigestTypes _hashFunction;
+    bool checkMessageSize(const AsymmetricPublicKey &key, size_t messageSize) const override;
```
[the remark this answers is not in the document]
blenk92: Well, lets assume there are three files
test.h:
```
class A {
public:
    virtual void func() = 0;
};

class B : public A {
// No declaration of func
};
```
test.cpp:
```
#include "test.h"

void B::func() {};
```
main.cpp:
```
#include "test.h"

int main () {
    A *a = new B();
    return 0;
}
```
That is what you are suggesting, right? At least for me this doesn't compile:
```
test.cpp:3:6: error: no declaration matches ‘void B::func()’
    3 | void B::func() {
      |      ^
test.cpp:3:6: note: no functions named ‘void B::func()’
In file included from test.cpp:1:
test.h:6:7: note: ‘class B’ defined here
    6 | class B : public A {
      |       ^
main.cpp: In function ‘int main()’:
main.cpp:4:18: error: invalid new-expression of abstract class type ‘B’
    4 |     A *a = new B();
      |                  ^
In file included from main.cpp:1:
test.h:6:7: note:   because the following virtual functions are pure within ‘B’:
    6 | class B : public A {
      |       ^
test.h:3:18: note: 	‘virtual void A::func()’
    3 |     virtual void func() = 0;
      |  
```
If you add the declaration though the code compiles just fine. Which means to me that `func()` has to be declared in `B` in order to implement it. Or am I missing something here?

src/mococrw/padding_mode.h
```diff
@@ -144,19 +115,16 @@ class PKCSSignaturePadding: public RSASignaturePadding {
...
      * @param hashFunction Hash function to be used
      */
     MGF1(openssl::DigestTypes hashFunction = openssl::DigestTypes::SHA256);
 
-    /**
-     * @brief Prepares the given openssl context with the masking function specific parameters.
-     * @param ctx OpenSSL context
-     */
-    void prepareOpenSSLContext(openssl::SSL_EVP_PKEY_CTX_Ptr& ctx) const;
+    void prepareOpenSSLContext(openssl::SSL_EVP_PKEY_CTX_Ptr& ctx) const override;
```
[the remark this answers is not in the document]
blenk92: see above

src/mococrw/padding_mode.h
```diff
@@ -165,52 +133,50 @@ class MGF1 {
...
      * @brief Copy Assignment
      */
-    PSSPadding& operator=(const PSSPadding& other);
+    OAEPPadding& operator=(const OAEPPadding& other);
 
     /**
      * @brief Destructor
      */
-    ~PSSPadding();
+    ~OAEPPadding();
```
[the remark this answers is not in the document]
blenk92: Well, I think the destructor is required because `Impl` is forward declared here. That is the destructor has to be declared once `Impl` is defined (in the source file). But you are right, probably it should be virtual...

src/mococrw/padding_mode.h
```diff
@@ -226,59 +192,43 @@ class PSSPadding: public RSASignaturePadding {
...
      */
-    ~OAEPPadding();
-
-    /**
-     * @brief Get maximum data size that can encrypted using the OAEP padding
-     * @param key RSA public key that will be used for encryption
-     * @return the maximum size of the data that can be encrypted in bytes.
-     */
-    int getDataBlockSize(const AsymmetricPublicKey &key) const;
+    ~PSSPadding();
```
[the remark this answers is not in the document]
blenk92: see above

blenk92:
> I have two major complains:
> 
>     * declaration of implemented interface methods in a declaration of the classes which implement those interfaces. It does not add any value but makes code harder to read.
> 
>     * Tests files are too big to reason about them.  Single test file includes several god-classes, some test methods checks several things. We can do a better job making them more focused and concise.

Hey,

i've changed a few things:
* I hope the tests are a bit cleaner now...
* Regarding your first point: As stated in the comments i don't think it is possible. And even if, I'd argue against that. Because from only reading the header you couldn't really tell if the class is abstract or if all virtual methods have been implemented.

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+
+## Padding Modes
+
+The current RSA encryption interface supports the following types of paddings for encryption:
+* No Padding
+* PKCS v1.5
+* OAEP
+
+### No Padding
+The following code shows how to create that de-/encrypts without padding the content. Please note that this entails the message size equals the size of the RSA key. Thus this mode should only be used with care:
```
garw: The comment is not really correct. The message size in the case can be actually smaller than the key size and this is the source of some problems. In contrary, with padding, the message size equals the size of the key.

Even though this is not critical to the code, it might be confusing to the reader and should be corrected.

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+std::vector<uint8_t> message = {...};
+std::vector<uint8_t> signature = {...};
+
+mococrw::AsymmetricPublicKey key = mococrw::AsymmetricKeypair::readPublicKeyFromPEM(pubKey);
+
+/*
+ * ...
+ */
+
+mococrw::ECDSASignaturePublicKeyCtx ctx = mococrw::RSASignaturePublicKeyCtx(key);
```
garw: EC = RSA?

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+std::vector<uint8_t> message = {...};
+std::vector<uint8_t> signature;
+
+mococrw::AsymmetricPrivateKey key = mococrw::AsymmetricKeypair::readPrivateKeyFromPEM(privKey, "");
+
+/*
+ * ...
+ */
+
+mococrw::ECDSASignaturePrivateKeyCtx ctx = mococrw::RSASignaturePrivateKeyCtx(key);
```
garw: again EC = RSA?

src/asymmetric_crypto_ctx.cpp
```diff
@@ -373,29 +311,22 @@ namespace mococrw {
...
+                                                         openssl::DigestTypes hashFunction,
+                                                         std::shared_ptr<RSASignaturePadding> padding)
+        : _impl(std::make_unique<RSASignaturePrivateKeyCtx::Impl>(key, hashFunction, padding)) {}
 
     RSASignaturePrivateKeyCtx::RSASignaturePrivateKeyCtx(const RSASignaturePrivateKeyCtx& other)
         : _impl(std::make_unique<RSASignaturePrivateKeyCtx::Impl>(*(other._impl))) {}
 
     RSASignaturePrivateKeyCtx& RSASignaturePrivateKeyCtx::operator=(const RSASignaturePrivateKeyCtx& other) {
-        _impl = std::make_unique<RSASignaturePrivateKeyCtx::Impl>(*(other._impl));
+        *this = RSASignaturePrivateKeyCtx(other);
```
garw: this is self-recursive. How is this supposed to work?

src/asymmetric_crypto_ctx.cpp
```diff
@@ -414,6 +345,10 @@ namespace mococrw {
         }
     }
 
+    std::vector<uint8_t> RSASignaturePrivateKeyCtx::signMessage(const std::vector<uint8_t> &message) {
+        return signMessageHelper(this, _impl->hashFunction, message);
```
garw: what is the rationale of these one-line helpers in a one liner function?

src/asymmetric_crypto_ctx.cpp
```diff
@@ -423,40 +358,29 @@ namespace mococrw {
...
+                                                       openssl::DigestTypes hashFunction,
+                                                       std::shared_ptr<RSASignaturePadding> padding)
+        : RSASignaturePublicKeyCtx(cert.getPublicKey(), hashFunction, padding) {}
 
     RSASignaturePublicKeyCtx::RSASignaturePublicKeyCtx(const RSASignaturePublicKeyCtx& other)
         : _impl(std::make_unique<RSASignaturePublicKeyCtx::Impl>(*(other._impl))) {}
 
     RSASignaturePublicKeyCtx& RSASignaturePublicKeyCtx::operator=(const RSASignaturePublicKeyCtx& other) {
-        _impl = std::make_unique<RSASignaturePublicKeyCtx::Impl>(*(other._impl));
+        *this = RSASignaturePublicKeyCtx(other);
```
garw: s.a.

src/asymmetric_crypto_ctx.cpp
```diff
@@ -519,16 +448,12 @@ namespace mococrw {
         : _impl(std::make_unique<ECDSASignaturePrivateKeyCtx::Impl>(*(other._impl))) {}
 
     ECDSASignaturePrivateKeyCtx& ECDSASignaturePrivateKeyCtx::operator=(const ECDSASignaturePrivateKeyCtx& other) {
-        _impl = std::make_unique<ECDSASignaturePrivateKeyCtx::Impl>(*(other._impl));
+        *this = ECDSASignaturePrivateKeyCtx(other);
```
garw: s.a.

src/asymmetric_crypto_ctx.cpp
```diff
@@ -566,18 +495,13 @@ namespace mococrw {
         : _impl(std::make_unique<ECDSASignaturePublicKeyCtx::Impl>(*(other._impl))) {}
 
     ECDSASignaturePublicKeyCtx& ECDSASignaturePublicKeyCtx::operator=(const ECDSASignaturePublicKeyCtx& other) {
-        _impl = std::make_unique<ECDSASignaturePublicKeyCtx::Impl>(*(other._impl));
+        *this = ECDSASignaturePublicKeyCtx(other);
```
garw: s.a.

src/mococrw/padding_mode.h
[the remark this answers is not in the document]
garw: imho, if you define one, you need to declare one

src/mococrw/padding_mode.h
```diff
@@ -25,116 +25,97 @@
...
+
+    void prepareOpenSSLContext(openssl::SSL_EVP_PKEY_CTX_Ptr& ctx) const override;
 };
 
 /**
- * @brief PKCSSignaturePadding
+ * @brief PKCSPadding
  *
- * This class defines the parameters specific to the PKCS1 padding mode.
+ * This class defines the parameters specific to the PKCS v1.5 padding mode.
```
garw: knitpicking: It's PKCS#1 v1.5

src/mococrw/padding_mode.h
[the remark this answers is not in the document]
garw: `override`?

src/padding_mode.cpp
```diff
@@ -33,27 +33,20 @@ using namespace openssl;
...
 /*
  * NoPadding
  */
 
-int NoPadding::getDataBlockSize(const AsymmetricPublicKey &key) const
+bool NoPadding::checkMessageSize(const AsymmetricPublicKey &key, size_t messageSize) const
 {
-    return key.getKeySize()/8;
+    size_t keySizeInBytes = key.getKeySize()/8;
+    return messageSize == keySizeInBytes;
```
garw: Is this enforced by OpenSSL? There is no mathematical reason for this? messageSize < keySizeInBytes should be just fine. In particular, if the size in bytes matches but the message is one bit too long, data ist lost?

src/padding_mode.cpp
```diff
@@ -62,43 +55,25 @@ void NoPadding::prepareOpenSSLContext(openssl::SSL_EVP_PKEY_CTX_Ptr& ctx) const
...
-void PKCSEncryptionPadding::prepareOpenSSLContext(openssl::SSL_EVP_PKEY_CTX_Ptr& ctx) const
-{
-    _EVP_PKEY_CTX_set_rsa_padding(ctx.get(), static_cast<int>(openssl::RSAPaddingMode::PKCS1));
-}
-
-int PKCSEncryptionPadding::getDataBlockSize(const AsymmetricPublicKey &key) const
+bool PKCSPadding::checkMessageSize(const AsymmetricPublicKey &key, size_t messageSize) const
 {
-    return key.getKeySize()/8 - c_pkcsMaxSizeSOverhead;
+    int remainingSize = key.getKeySize()/8 - c_pkcsMaxSizeSOverhead;
```
garw: c_
please remove this weird remnant of hungarian notation. This library doesn't use it. Constants are usually uppercased with underscores. so PKCS_MAX_SIZE_OVERHEAD

tests/unit/test_signature.cpp
```diff
@@ -341,48 +536,25 @@ TEST_F(SignatureTest, testUnsuccessfulRsaSignatureWithInvalidDigestLengthPKCS1)
...
-
 /**
  * @brief Verification fails due to an invalid public key.
  */
 TEST_F(SignatureTest, testUnsuccessfulRsaVerificationWithWrongPublicKey)
 {
     std::vector<uint8_t> signature;
 
-    auto signCtx256 = RSASignaturePrivateKeyCtx(*_keyPairRsa, _sha256PSSPadding);
+    auto signCtx256 = RSASignaturePrivateKeyCtx(_keyPairRsa, DigestTypes::SHA256,_PSSPadding);
```
neverpanic: Missing space after comma.

tests/unit/test_signature.cpp
```diff
@@ -222,102 +178,341 @@ const std::vector<uint8_t> SignatureTest::_testMessageDigestSHA1{
...
-    EXPECT_NO_THROW(signature = ctx256.signDigest(_testMessageDigestSHA256));
-    EXPECT_EQ(_validPKCS1SignatureSHA256, signature);
+};
 
-    auto ctx512 = RSASignaturePrivateKeyCtx(key, _sha512PKCSSignaturePadding);
-    EXPECT_NO_THROW(signature = ctx512.signDigest(_testMessageDigestSHA512));
-    EXPECT_EQ(_validPKCS1SignatureSHA512, signature);
-}
+const std::vector<uint8_t>& SignatureTest::_validPKCS1SignatureSHA256 = SignatureTestData[1].validSignature;
+const std::vector<uint8_t>& SignatureTest::_validPKCS1SignatureSHA512 = SignatureTestData[2].validSignature;
```
neverpanic: I couldn't find a use of `_validPKCS1SignatureSHA512`. Am I missing something or should it be removed?

tests/unit/test_padding_mode.cpp
```diff
@@ -0,0 +1,299 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2018 BMW Car IT GmbH
```
neverpanic: 2019?

tests/unit/test_padding_mode.cpp
```diff
@@ -0,0 +1,299 @@
...
+public:
+    static const std::vector<MessageSizeDataSet> messageSizeDataSet;
+    static const AsymmetricPublicKey rsa1024bit;
+    static const AsymmetricPublicKey rsa2048bit;
+};
+
+const AsymmetricPublicKey PaddingModeTest::rsa1024bit = mococrw::AsymmetricKeypair::generate(RSASpec(1024));
+const AsymmetricPublicKey PaddingModeTest::rsa2048bit = mococrw::AsymmetricKeypair::generate(RSASpec(2048));
+
+/// \brief Data set used to test the message size limits use-cases, encrytion functionality
```
neverpanic: Did you mean: "encryption functionality"?

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
+# Asymmetric Crypto API {#example5}
+
+# Common Interface
+
+## Encryption Interface
+
+All contexts that support en- / decryption are implementing an `encrypt()` or `decrypt()` method resprectively. The classes `mococrw::EncryptionCtx` and `mococrw::DecryptionCtx` provide a pure virtual interface to these methods:
+<!-- TODO for when refactoring Examples: Generate such code snippets from code-->
+\code{.cpp}
```
neverpanic: As a general question about this documentation: These are markdown files, but we're using doxygen tags `\code{.cpp}` in them. Is there a rendered version of these available somewhere? Otherwise I would probably to use markdown code blocks instead so these examples are at least readable on GitHub.

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+    ...
+    virtual std::vector<uint8_t> encrypt(const std::vector<uint8_t>& message) = 0;
+};
+
+class DecryptionCtx {
+    ...
+    virtual std::vector<uint8_t> decrypt(const std::vector<uint8_t>& message) = 0;
+}
+\endcode
+That is, these classes can be used to access a en-/decryption context in a generic manner.
```
neverpanic: It's **an** encryption context.

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+
+class MessageVerificationCtx {
+    ...
+    virtual void verifyMessage(const std::vector<uint8_t> &signature,
+                               const std::vector<uint8_t> &message) = 0;
+};
+\endcode
+
+#### Example
+If a context supports both, signing messages and message digest, the following code snippets are equal. For simplicity it is assumes that SHA256 is used to create the signed hash.
```
neverpanic: "it is assumes" is not good Engrish. I propose we turn this into "it is assumed" or "we assume".

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+    * Empty label
+
+## Encryption (Default Padding)
+
+\code{.cpp}
+std::string pubKey= R"(-----BEGIN PUBLIC KEY-----
+...
+-----END PUBLIC KEY-----)";
+
+std::vector<uint8_t> encryptedMessage = {...};
```
neverpanic: This should have been `message`, right? You're declaring `std::vector<uint8_t> encryptedMessage` below with the result of the encryption operation.

doc/examples/example5.md
neverpanic: "shows how to create that de-/encrypts" Create what? There seems to be an object missing in this sentence.

"this entails the message size equals" Is this missing a "that", e.g. "entails **that** the message size"?

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+    std::shared_ptr<mococrw::MaskGenerationFunction> mgf = std::make_shared<MGF1>(mococrw::openssl::DigestTypes::SHA512);
+
+    std::shared_ptr<mococrw::RSAEncryptionPadding> padding = std::make_shared<mococrw::OAEPPadding>(mococrw::openssl::DigestTypes::SHA256, mgf);
+
+    mococrw::RSAEncryptionPublicKeyCtx encCtx = mococrw::RSAEncryptionPublicKeyCtx(key, padding);
+
+    mococrw::RSAEncryptionPrivateKeyCtx decCtx = mococrw::RSAEncryptionPrivateKeyCtx(key, padding);
+
+\endcode
+Please note that for the moment beeing only the mask generation function MGF1 is available. The example above uses SHA256 as OEAP hash function and SHA512 as hash function for MGF1.
```
neverpanic: `s/beeing/being/`, we're not dealing with 🐝s here ;-)

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+// Hash of message is calculated automatically
+signature = ctx.signMessage(message); 
+
+// Alternative (manual hashing)
+std::vector<uint8_t> messageDigest = mococrw::Hash::sha256(message);
+signature = ctx.signDigest(messageDigest);
+\endcode
+
+## Using Different Hash Function
+To use different hash function to create the hash to be signed (e.g. SHA512) the contexts has to be created as follows:
```
neverpanic: The third person plural form of "(to) have" is "have", thus: `s/has/have/`

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+
+## Padding Modes
+
+Currently, the following padding modes are supported:
+* PKCS v1.5 Padding
+* PSS Padding
+
+### PKCS v1.5
+
+The following example shows how to create signing / verification contexts using the PKCS v1.5 Padding.
```
neverpanic: Strictly speaking, "PKCS#1 v1.5".

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+
+If the salt length shall be a non-default value, `nullptr` can be passed as padding which keeps the default behaviour of the Mask Generation Function. The following example sets the salt length to 20, while keeping the default behaviour for the Mask Generation Function:
+\code{.cpp}
+std::shared_ptr<RSASignaturePadding> padding = std::make_shared<PSSPadding>(nullptr, 20);
+\endcode
+
+
+# ECDSA
+
+ECDAS only allows to choose the hash function that is used to create the signed hash. The default value is `SHA256`.
```
neverpanic: `s/ECDAS/ECDSA/`

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+// Hash of message is calculated automatically
+signature = ctx.signMessage(message); 
+
+// Alternative (manual hashing)
+std::vector<uint8_t> messageDigest = mococrw::Hash::sha256(message);
+signature = ctx.signDigest(messageDigest);
+\endcode
+
+## Using Different Hash Function
+To use different hash function to create the hash to be signed (e.g. SHA512) the contexts has to be created as follows:
```
neverpanic: `s/has/have/`

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+\code{.cpp}
+mococrw::ECDSASignaturePublicKeyCtx verifyCtx = mococrw::RSASignaturePublicKeyCtx(key, mococrw::openssl::DigestTypes::SHA512);
```
neverpanic: `s/RSA/ECDSA/`

doc/examples/example5.md
```diff
@@ -0,0 +1,397 @@
...
+
+mococrw::ECDSASignaturePrivateKeyCtx signCtx = mococrw::RSASignaturePrivateKeyCtx(key, mococrw::openssl::DigestTypes::SHA512);
```
neverpanic: Ditto.

doc/examples/example5.md
blenk92: I was already talking to @vkrikun about this. The contents of these md files are embedded into the generated doxygen documentation. We agreed on looking if there are better solutions to this as part of a separate task (as far as i remember there is already such a ticket somewhere in the backlog)

src/mococrw/padding_mode.h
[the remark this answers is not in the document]
blenk92: Virtual destructors don't work like virtual functions. afaik they are chained which means that the destructor of the base  as well as the destructor of the derived class are executed. That is, there is no override for destructors.

src/mococrw/padding_mode.h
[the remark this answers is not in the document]
blenk92: see above

src/padding_mode.cpp
blenk92: Yes this is enforced by openssl:
```
flen must not be more than RSA_size(rsa) - 11 for the PKCS #1 v1.5 based padding 
modes, not more than RSA_size(rsa) - 42 for RSA_PKCS1_OAEP_PADDING and
exactly RSA_size(rsa) for RSA_NO_PADDING.
```
See https://www.openssl.org/docs/man1.1.1/man3/RSA_public_encrypt.html

Message size shoud imho always be a multiple of 8 Bit in this lib because the lib only offers to en-/decrypt vector<uint8_t>. But indeed, it is possible that the key size is not a multiple of 8 bits. (Will fix this)...

doc/examples/example5.md
blenk92: Well, I think it is correct in the context of this library. See comment to NoPadding below...

PR #42, opened 2019-10-17
tests: Disable installing googletest & googlemock

neverpanic:
With version 1.8, googletest and googlemock gained support for CMake install rules, which causes MoCOCrW to install `/usr/lib/libgtest.a` when not explicitly disabled.

Since we only need this library for testing, disable the installation.


neverpanic:
It seems this doesn't work yet, so please don't merge as-is.

PR #51, opened 2020-01-17
CSR: Additional features

garw:
- Add interfaces to read/write DER in addition to PEM format
- Add option to choose the digest function that used in the
  signature of the CSR.

[ This change has currently breaking tests due to an expired certificate in another test. However, it can be individually reviewed ]

### DIFF
--- a/src/csr.cpp
+++ b/src/csr.cpp
@@ -26,7 +26,8 @@ namespace mococrw
 using namespace openssl;
 
 CertificateSigningRequest::CertificateSigningRequest(const DistinguishedName &dn,
-                                                     const AsymmetricKeypair &keypair)
+                                                     const AsymmetricKeypair &keypair,
+                                                     const openssl::DigestTypes digestFunction)
         : _req{openssl::_X509_REQ_new()}
 {
     /* setup x509 version number */
@@ -40,13 +41,19 @@ CertificateSigningRequest::CertificateSigningRequest(const DistinguishedName &dn
 
     auto mctx = _EVP_MD_CTX_create();
 
-    auto digestType = DigestTypes::SHA256;
+    auto digestType = digestFunction;
     if (keypair.getType() == AsymmetricKey::KeyTypes::ECC_ED) {
         digestType = DigestTypes::NONE;
     }
     _EVP_DigestSignInit(mctx.get(), digestType, const_cast<EVP_PKEY*>(keypair.internal()));
 
     _X509_REQ_sign_ctx(_req.get(), mctx.get());
+}
+
+CertificateSigningRequest::CertificateSigningRequest(const DistinguishedName &dn,
+                                                     const AsymmetricKeypair &keypair)
+        : CertificateSigningRequest{dn, keypair, DigestTypes::SHA256}
+{
 }
 
 AsymmetricPublicKey CertificateSigningRequest::getPublicKey() const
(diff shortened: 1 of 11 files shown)


src/mococrw/csr.h
```diff
@@ -38,8 +38,30 @@ class CertificateSigningRequest
...
+
+    /**
+     * Return the DER for this CSR as a string.
+     */
+    std::vector<uint8_t> toDER() const;
+
+
     /**
      * Return the PEM for this CSR as a string.
+     * @depricated - inconsistenting capitalization with rest of lib
```
ricleite: I guess you mean `@deprecated`

ricleite (approved):
LGTM (part of JC-3 SF)

Example files might need some updating..

src/csr.cpp
```diff
@@ -92,6 +111,19 @@ CertificateSigningRequest CertificateSigningRequest::fromPEMFile(const std::stri
...
 }
 
+CertificateSigningRequest CertificateSigningRequest::fromDER(const std::vector<uint8_t> &derData)
+{
+    BioObject bio{BioObject::Types::MEM};
+    bio.write(derData);
+    return CertificateSigningRequest{_d2i_X509_REQ_bio(bio.internal())};
+}
+
+CertificateSigningRequest CertificateSigningRequest::fromDERFile(const std::string &filename)
```
donatellob, line 121: Why no "toDERFile" method?

donatellob (approved):

src/csr.cpp
garw: No particular reason. This doesn't exist for the other objects as well. Do you need this? Then I can add it.

src/mococrw/csr.h
garw: yes. Thanks.

src/mococrw/openssl_wrap.h
```diff
@@ -528,13 +528,28 @@ int _BIO_read(BIO* bio, std::vector<uint8_t> &buf, std::size_t numBytes);
  * @throw OpenSSLException if an OpenSSL internal error occurs.
  */
 SSL_X509_Ptr _d2i_X509_bio(BIO* bp);
+
+
```
blenk92: Nitpicking: move one empty line below declaration

src/mococrw/openssl_wrap.h
```diff
@@ -528,13 +528,28 @@ int _BIO_read(BIO* bio, std::vector<uint8_t> &buf, std::size_t numBytes);
...
+SSL_X509_REQ_Ptr _d2i_X509_REQ_bio(BIO* bp);
 /*
  * Write an X509 certificate in DER encoded form to a bio
  *
  * @throw OpenSSLException if an OpenSSL internal error occurs.
  */
 void _i2d_X509_bio(BIO* bp, X509* x);
 
+/*
+ * Write an X509 certificatfication request in DER encoded form to a bio
```
blenk92: certification

src/mococrw/csr.h
```diff
@@ -38,8 +38,30 @@ class CertificateSigningRequest
...
+
+    /**
+     * Return the DER for this CSR as a string.
+     */
+    std::vector<uint8_t> toDER() const;
+
+
     /**
      * Return the PEM for this CSR as a string.
+     * @deprecated - inconsistenting capitalization with rest of lib
```
blenk92: Typo: inconsisting

blenk92 (commented):
Some typos, otherwise looks good to me

src/csr.cpp
donatellob: Not needed, I was wondering if it was forgotten as you added "fromDERFile"

donatellob (approved):
Thanks! LGTM

src/mococrw/csr.h
```diff
@@ -38,8 +38,30 @@ class CertificateSigningRequest
...
+                                       const AsymmetricKeypair &key,
+                                       const openssl::DigestTypes digestFunction);
+
+    /**
+     * Return the PEM for this CSR as a string.
+     */
+    std::string toPEM() const;
+
+    /**
+     * Return the DER for this CSR as a string.
```
neverpanic: Technically, it's not a string, since it's a `std::vector<uint8_t>`, but I guess we could still stay that's a binary string.

src/mococrw/csr.h
neverpanic: Did you mean: "inconsistent"?

Also, should we use the C++14 `[[deprecated]]` attribute (see https://en.cppreference.com/w/cpp/language/attributes/deprecated)?

neverpanic (approved):
LG2M!

PR #52, opened 2020-01-17
tests: Replace expired CRL test CAs and certs

neverpanic:
Tests started failing because the root CA used to test CRL functionality (as well as a few CRLs) expired beginning of 2020. Switch to a set of certificates generated similar to our other test certificates where we keep the keys around so we can prevent having to re-generate the entire set next time (although the expiry dates are now all 100y into the future, so this is unlikely to happen again during my and your lifetime).

The old test certificates used for the CRL testing have expired and the private keys are no longer available. Set up a CA infrastructure similar to `tests/certs/rootCA{1,2}` that should be used to test CRLs in the future.

Generate a number of intermediate CAs and enduser certificates and provide CRLs that mark those certificates and CAs as valid, revoked or are expired.

Switch to SHA256 signatures for all CRLs over the SHA1 signatures that were previously used. Ensure expiration dates for all certificates, CAs and CRLs that we consider valid are 100 years into the future.

Extend the openssl.cnf used with a section for CRLs as outlined in [1].

[1] https://jamielinux.com/docs/openssl-certificate-authority/index.html


tests/unit/test_crl.cpp
```diff
@@ -158,10 +158,10 @@ TEST_F(CRLTest, testHasCorrectValues)
     EXPECT_EQ(rootCert->getSubjectDistinguishedName(), rootCrl->getIssuerName());
 
     // Last update: Feb  2 10:35:56 2018 GMT
```
blenk92: update time in comment

tests/unit/test_crl.cpp
```diff
@@ -158,10 +158,10 @@ TEST_F(CRLTest, testHasCorrectValues)
...
-    EXPECT_EQ(rootCrl->getLastUpdateAsn1(),  Asn1Time::fromString("20180202103556Z"));
+    EXPECT_EQ(rootCrl->getLastUpdateAsn1(),  Asn1Time::fromString("20200117164847Z"));
 
     // Next update: Jan  9 10:35:56 2118 GMT
```
blenk92: update time in comment

garw (approved):

PR #55, opened 2020-02-17
symmetricencryption: Add howto document with example code

vkrikun:


doc/examples/example_symm.md
```diff
@@ -0,0 +1,206 @@
+# Symmetric Crypto API {#example_sym}
```
neverpanic, line 1: The GitHub rendering does not support `{#example_sym}`. Should we just drop it?

doc/examples/example_symm.md
```diff
@@ -0,0 +1,206 @@
...
+//
+// Encryption
+//
+
+auto encryptor = AESCipherBuilder{SymmetricCipherMode::GCM, SymmetricCipherKeySize::S_256, secretKey}.buildAuthenticatedEncryptor();
+
+std::vector<uint8_t> ciphertext;
+
+// Encrypt plaintext messages in 1K blocks.
+// NOTE: If `update()`-s and the following `read()` operations use the same chunk sizes you can
```
neverpanic: This is actually only the case if the cipher is a stream cipher, or a block cipher and the input size produces an equivalent output size, i.e. if the input size is a multiple of the block size.

I've modified the implementation so that this also works correctly when using readAll() inside of the loop, and I think we should recommend that.

doc/examples/example_symm.md
```diff
@@ -0,0 +1,206 @@
...
+
+//
+// Decryption
+//
+
+auto decryptor = AESCipherBuilder{SymmetricCipherMode::GCM, SymmetricCipherKeySize::S_256, secretKey}.setIV(iv).buildAuthenticatedDecryptor();
+
+
+// You can mix `update()`, `read()` and `readAll()`. For example, in the code below we first
+// read and decrypt several packets, then accumulated some decrypted blocks in the cipher
```
neverpanic:
```suggestion
// read and decrypt several packets, then accumulate some decrypted blocks in the cipher
```

doc/examples/example_symm.md
```diff
@@ -0,0 +1,206 @@
...
+// internal memory and finally reading all the remaining plaintext from the cipher.
```
neverpanic:
```suggestion
// internal memory and finally read all the remaining plaintext from the cipher.
```

doc/examples/example_symm.md
```diff
@@ -0,0 +1,206 @@
...
+//
+// NOTE: `update()` and sequential `read()` of the same chunk size is still might be the most
```
neverpanic: The same applies here, `readAll()` now also works here.

neverpanic (changes requested):

doc/examples/example_symm.md
vkrikun: It is ugly but consistent. We have a separate ticket to convert all Doxygen-formatted Markdown documentation into format Github understands.

doc/examples/example_symm.md
vkrikun: Agree, fixed.

doc/examples/example_symm.md
vkrikun: Done

doc/examples/example_symm.md
```diff
@@ -0,0 +1,211 @@
...
+
+std::vector<uint8_t> ciphertext;
+
+// Encrypt plaintext messages in 1K blocks.
+// NOTE: Note about performance: default memory model for streaming-mode is optimized for the
+// following use-cases:
+// - `update()` followed by `readAll()`. `readAll` will return one chunk of encrypted/decrypted
+// data with zero-copy.
+// - `update()` followed by `read()` with the same chunk size (only if the cipher is a stream
+// cipher, or a block cipher and the input size produces an equivalent output size )
```
neverpanic:
```suggestion
// cipher, or a block cipher and the input size produces an equivalent output size)
```

doc/examples/example_symm.md
```diff
@@ -0,0 +1,211 @@
...
+//
+
+auto decryptor = AESCipherBuilder{SymmetricCipherMode::GCM, SymmetricCipherKeySize::S_256, secretKey}.setIV(iv).buildAuthenticatedDecryptor();
+
+
+// You can mix `update()`, `read()` and `readAll()`. For example, in the code below we first
+// read and decrypt several packets, then accumulate some decrypted blocks in the cipher
+// internal memory and finally read all the remaining plaintext from the cipher.
+//
+// NOTE: `update()` and sequential `readAll`  (or`read()` of the same chunk size) is still might be
```
neverpanic:
```suggestion
// NOTE: `update()` and sequential `readAll()` (or `read()` of the same chunk size) is still
```

neverpanic (changes requested):

neverpanic (approved):

PR #56, opened 2020-03-18
ECIES: Adding interfaces

wusto:
This commit introduces interfaces for ECIES (Elliptic Curve
Integrated Encryption Scheme).
For this purpose additional classes for key derivation and
message authentication code are added.

### DIFF
--- a/src/mococrw/ecies.h
+++ b/src/mococrw/ecies.h
@@ -244,7 +244,7 @@ public:
      * @param iv The initialisation vector. The length shall be the same as the key length
      * @return builder instance
      */
-    ECIESCtxBuilder &setSymmetricCipherIv(std::vector<uint8_t> iv);
+    ECIESCtxBuilder &setSymmetricCipherIv(const std::vector<uint8_t>& iv);
 
     /**
      * @brief Sets the salt for the key derivation function (KDF). (Default: empty vector -> not used)
--- a/src/mococrw/ecies.h
+++ b/src/mococrw/ecies.h
@@ -290,7 +290,7 @@ public:
     ECIESEncryptionCtxBuilder(const AsymmetricPublicKey& bobsKey);
 
     /**
-     * @brief Constructor if the public key is inside a certificate
+     * @brief Constructor for certificates
      * @param bobsCert The certificate of the receiver which contains his/her public key.
      */
     ECIESEncryptionCtxBuilder(const X509Certificate& bobsCert);
--- a/src/mococrw/ecies.h
+++ b/src/mococrw/ecies.h
@@ -237,7 +237,7 @@ public:
      * @param padding The padding scheme to be used.
      * @return builder instance
      */
-    ECIESCtxBuilder &setSymmetircCipherPadding(mococrw::SymmetricCipherPadding padding);
+    ECIESCtxBuilder &setSymmetricCipherPadding(mococrw::SymmetricCipherPadding padding);
 
     /**
      * @brief Sets the initialisation vector (IV) for the symmetric encryption. (Default: zeros)
--- a/src/mococrw/ecies.h
+++ b/src/mococrw/ecies.h
@@ -58,7 +58,7 @@ public:
      * 2. Encrypt the plaintext
      * 3. Calculate the tag/mac (use getMac() for getting it)
      *
-     * After invoking this function the ephemeral key and the authentication key can obtained using the getter
+     * After invoking this function, the ephemeral key and the authentication key can obtained using the getter
      * functions.
      * @return The encrypted data.
      */
--- a/src/mococrw/ecies.h
+++ b/src/mococrw/ecies.h
@@ -69,7 +69,7 @@ public:
      *
      * The ephemeral key is generated during encryption. This public key (in combination with the private key of the
      * receiver) is necessary for decrypting the ciphertext.
-     * @throws MoCOCrWException if finish() hasn't invoked beforehand
+     * @throws MoCOCrWException if finish() hasn't been invoked beforehand
      * @return The ephemeral key generated during encryption
      */
     AsymmetricPublicKey getEphemeralKey();
--- a/src/mococrw/ecies.h
+++ b/src/mococrw/ecies.h
@@ -79,7 +79,7 @@ public:
      *
      * The authentication tag is calculated as the last step during encryption. This tag verifies the
      * authenticity of the message.
-     * @throws MoCOCrWException if finish() hasn't invoked beforehand
+     * @throws MoCOCrWException if finish() hasn't been invoked beforehand
      * @return The tag/mac
      */
     std::vector<uint8_t> getAuthTag();
--- a/src/mococrw/ecies.h
+++ b/src/mococrw/ecies.h
@@ -1,0 +1,332 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2018 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+#include "symmetric_crypto.h"
+#include "mac.h"
+#include "kdf.h"
+#include "asymmetric_crypto_ctx.h"
+
+namespace mococrw {
+
+class ECIESEncryptionCtxBuilder;
+
+/**
+ * @brief ECIESEncryptionCtx
+ *
+ * This class supports encryption of plain texts using ECIES
+ *
+ */
+class ECIESEncryptionCtx {
+public:
+    /**
+     * @brief Destructor
+     */
+    ~ECIESEncryptionCtx();
+
+    /**
+     * Encrypt a chunk of data.
+     *
+     * This function stores the data in the internal buffer.
+     * The data in the internal buffer is encrypted, when the finish method is invoked.
+     *
+     * @param message chunk of data to encrypt.
+     * @throws MoCOCrWException if this function is invoked after finish was called.
+     */
+    void update(const std::vector<uint8_t>& message);
+
+    /**
+     * @brief Finalize the encryption.
+     *
+     * This includes:
+     * 1. Creating the ephemeral key used for encryption (use getEphemeralKey() for getting it)
+     * 2. Encrypt the plaintext
+     * 3. Calculate the tag/mac (use getMac() for getting it)
+     *
+     * After invoking this function the ephemeral key and the authentication key can obtained using the getter
+     * functions.
+     * @return The encrypted data.
+     */
+    std::vector<uint8_t> finish();
+
+    /**
+     * @brief Returns the ephemeral key
+     *
+     * The ephemeral key is generated during encryption. This public key (in combination with the private key of the
+     * receiver) is necessary for decrypting the ciphertext.
+     * @throws MoCOCrWException if finish() hasn't invoked beforehand
+     * @return The ephemeral key generated during encryption
+     */
+    AsymmetricPublicKey getEphemeralKey();
+
+    /**
+     * @brief Return the authentication tag
+     *
+     * The authentication tag is calculated as the last step during encryption. This tag verifies the
+     * authenticity of the message.
+     * @throws MoCOCrWException if finish() hasn't invoked beforehand
+     * @return The tag/mac
+     */
+    std::vector<uint8_t> getAuthTag();
+
+private:
+    ECIESEncryptionCtx(const AsymmetricPublicKey& pubKey,
+                       const std::shared_ptr<mococrw::KeyDerivationFunction> kdf,
+                       const std::shared_ptr<mococrw::MessageAuthenticationCode> mac,
+                       const mococrw::SymmetricCipherMode mode,
+                       const mococrw::SymmetricCipherKeySize keySize,
+                       const mococrw::SymmetricCipherPadding padding,
+                       const std::vector<uint8_t>& S1,
+                       const std::vector<uint8_t>& S2,
+                       const std::vector<uint8_t>& iv);
+
+    ECIESEncryptionCtx(const X509Certificate& cert,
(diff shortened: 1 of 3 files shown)


src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * @brief Finalize the encryption.
+     *
+     * This includes:
+     * 1. Creating the ephemeral key used for encryption (use getEphemeralKey() for getting it)
+     * 2. Encrypt the plaintext
+     * 3. Calculate the tag/mac (use getMac() for getting it)
+     *
+     * After invoking this function the ephemeral key and the authentication key can obtained using the getter
```
blenk92:
```suggestion
     * After invoking this function, the ephemeral key and the authentication key can obtained using the getter
```
I'd add a comma here :-D

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+     * @return The encrypted data.
+     */
+    std::vector<uint8_t> finish();
+
+    /**
+     * @brief Returns the ephemeral key
+     *
+     * The ephemeral key is generated during encryption. This public key (in combination with the private key of the
+     * receiver) is necessary for decrypting the ciphertext.
+     * @throws MoCOCrWException if finish() hasn't invoked beforehand
```
blenk92:
```suggestion
     * @throws MoCOCrWException if finish() hasn't been invoked beforehand
```

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+     * @return The ephemeral key generated during encryption
+     */
+    AsymmetricPublicKey getEphemeralKey();
+
+    /**
+     * @brief Return the authentication tag
+     *
+     * The authentication tag is calculated as the last step during encryption. This tag verifies the
+     * authenticity of the message.
+     * @throws MoCOCrWException if finish() hasn't invoked beforehand
```
blenk92:
```suggestion
     * @throws MoCOCrWException if finish() hasn't been invoked beforehand
```

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2018 BMW Car IT GmbH
```
neverpanic, line 4: 2020?

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+    /**
+     * @brief Destructor
+     */
+    ~ECIESEncryptionCtx();
+
+    /**
+     * Encrypt a chunk of data.
+     *
+     * This function stores the data in the internal buffer.
+     * The data in the internal buffer is encrypted, when the finish method is invoked.
```
neverpanic, line 46: I think we should write the documentation as if we're implementing streaming encryption right away. I'd consider the documentation part of the interface, and the interface shouldn't change when we implement the streaming support.

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+     * @return builder instance
+     */
+    ECIESCtxBuilder &setSymmetircCipherPadding(mococrw::SymmetricCipherPadding padding);
+
+    /**
+     * @brief Sets the initialisation vector (IV) for the symmetric encryption. (Default: zeros)
+     * @param iv The initialisation vector. The length shall be the same as the key length
+     * @return builder instance
+     */
+    ECIESCtxBuilder &setSymmetricCipherIv(std::vector<uint8_t> iv);
```
blenk92:
```suggestion
    ECIESCtxBuilder &setSymmetricCipherIv(const std::vector<uint8_t>& iv);
```
any reason to no use const-refs here (and in the other places)?

src/mococrw/ecies.h
neverpanic:
```suggestion
     * @throws MoCOCrWException if finish() was not invoked beforehand
```

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+{
+public:
+    /**
+     * \brief Constructor for public keys
+     * \param bobsKey The public key of the receiver.
+     */
+    ECIESEncryptionCtxBuilder(const AsymmetricPublicKey& bobsKey);
+
+    /**
+     * @brief Constructor if the public key is inside a certificate
```
blenk92:
```suggestion
     * @brief Constructor for certificates
```
That sounds odd... maybe just say certificates?

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * @brief Return the authentication tag
+     *
+     * The authentication tag is calculated as the last step during encryption. This tag verifies the
+     * authenticity of the message.
+     * @throws MoCOCrWException if finish() hasn't invoked beforehand
+     * @return The tag/mac
+     */
+    std::vector<uint8_t> getAuthTag();
```
neverpanic, line 85: Your documentation for `finish()` says this is called `getMac()`, and I'd prefer that wording since "auth tag" is mostly used in symmetric AEAD.

Additionally, I'd capitalize `getMac()` to `getMAC()`, since that matches the style we've used in the X509 functionality with `fromPEM()` and `toPEM()`.

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+    void setEphemeralKey(const AsymmetricPublicKey& ephKey);
+
+    /**
+     * @brief Set the authentication tag
+     *
+     * As the last step during encryption the authentication tag is calculated. Set this tag to verify the authenticity
+     * of the received message.
+     * @param tag The authentication tag
+     */
+    void setAuthTag( const std::vector<uint8_t>& tag);
```
neverpanic, line 172: I'd also use `setMAC()` here.

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+    ~ECIESDecryptionCtx();
+
+    /**
+     * Decrypt a chunk of data.
+     *
+     * This function stores the data in the internal buffer.
+     * The data in the internal buffer is decrypted, when the finish method is invoked.
+     *
+     * @param message chunk of data to decrypt.
+     * @throws MoCOCrWException if this function is invoked after finish was called.
```
neverpanic, line 144: This is another case where the documentation should already reflect the streaming use-case. For example, for streaming decryption, `setEphemeralKey()` must have been called before the first `update()`, and we should both document and enforce this in the implementation to prevent situations where users didn't do this and we can't change the implementation later on without breaking their software.

src/mococrw/kdf.h
```diff
@@ -0,0 +1,119 @@
...
+{
+
+/**
+ * @brief Abstract class for key derivation functions
+ */
+class KeyDerivationFunction
+{
+public:
+    /**
+     * destructor
```
blenk92, line 34:
```suggestion
     * @brief destructor
```

src/mococrw/kdf.h
```diff
@@ -0,0 +1,119 @@
...
+     */
+    ~KeyDerivationFunction();
```
blenk92, line 36: default destructor is not enough? or did want a virtual destructor?

src/mococrw/kdf.h
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     */
+    PBKDF2(openssl::DigestTypes hashFunction);
+
+    /**
+     * @brief destructor
+     */
+    ~PBKDF2() {}
```
blenk92, line 63: i guess the implementation should move into the source files somewhen (and may be replaced by `= default`)

src/mococrw/mac.h
```diff
@@ -0,0 +1,72 @@
...
+    HMAC(openssl::DigestTypes hashFunction, uint32_t tagBitsLength);
+
+    /**
+     * @brief destructor
+     */
+    ~HMAC();
+
+    /**
+     * @brief Calculates the tag/mac based on the key, the message and the parameters provided to the constructor
+     * @param key The which shall be used for calculation
```
blenk92:
```suggestion
     * @param key The key which shall be used for calculation
```
?

src/mococrw/ecies.h
blenk92, line 172:
```suggestion
    void setAuthTag(const std::vector<uint8_t>& tag);
```

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * @brief Sets the key size for the symmetric encryption. (Default: mococrw::SymmetricCipherKeySize::S_256)
+     * @param keySize The key size to be used
+     * @return builder instance
+     */
+    ECIESCtxBuilder &setSymmetricCipherKeySize(mococrw::SymmetricCipherKeySize keySize);
+
+    /**
+     * @brief Sets the padding for the symmetric encryption. (Default: mococrw::SymmetricCipherPadding::NO)
```
neverpanic, line 236: Is the default really AES-CBC without padding? IIRC CBC is a block cipher, which requires padding if you want to support plaintexts that are not multiples of the cipher block size.

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+     * @param padding The padding scheme to be used.
+     * @return builder instance
+     */
+    ECIESCtxBuilder &setSymmetircCipherPadding(mococrw::SymmetricCipherPadding padding);
```
neverpanic:
```suggestion
    ECIESCtxBuilder &setSymmetricCipherPadding(mococrw::SymmetricCipherPadding padding);
```

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * @brief Sets the initialisation vector (IV) for the symmetric encryption. (Default: zeros)
```
neverpanic, line 243:
```suggestion
     * @brief Sets the initialisation vector (IV) for the symmetric encryption. (Default: zeroes)
```

src/mococrw/ecies.h
neverpanic: Actually, if this is a copy, it could be `std::move()`d, and that may not be the worst idea.

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+public:
+    /**
+     * \brief Constructor for public keys
+     * \param bobsKey The public key of the receiver.
+     */
+    ECIESEncryptionCtxBuilder(const AsymmetricPublicKey& bobsKey);
+
+    /**
+     * @brief Constructor if the public key is inside a certificate
+     * @param bobsCert The certificate of the receiver which contains his/her public key.
```
neverpanic, line 294:
```suggestion
     * @param bobsCert The certificate of the receiver which contains their public key.
```
is a common approach to this issue.

src/mococrw/ecies.h
```diff
@@ -0,0 +1,332 @@
...
+    AsymmetricPublicKey pubKey;
+    X509Certificate cert;
+};
+
+class ECIESDecryptionCtxBuilder : public ECIESCtxBuilder
+{
+public:
+    /**
+     * @brief Constructor
+     * @param bobsKey The private key of the receiver. This key is needed for decrypting the ciphertext
```
neverpanic, line 316:
```suggestion
     * @param bobsKey The receiver's private key. This key is needed for decrypting the ciphertext
```
My English teacher didn't stop mentioning this.

src/mococrw/kdf.h
```diff
@@ -0,0 +1,119 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2018 BMW Car IT GmbH
```
neverpanic, line 4:
```suggestion
 * Copyright (C) 2020 BMW Car IT GmbH
```

src/mococrw/kdf.h
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * This function returns a derived key based on the arguments to this function and the arguments provided to
+     * the constructor
+     * @param password The password/secret key which shall be used for key derivation
+     * @param outputLength The length of the derived key in bits.
+     * @param salt The salt used for the key derivation
+     * @throws MoCOCrWException if the required output length is greater than the length of the raw output.
+     * @return The derived key
+     */
+    virtual std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const uint32_t outputLength,
```
neverpanic, line 49:
```suggestion
    virtual std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
```
Lengths are usually `size_t`s.

src/mococrw/mac.h
```diff
@@ -0,0 +1,72 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2018 BMW Car IT GmbH
```
```suggestion
 * Copyright (C) 2020 BMW Car IT GmbH
```

src/mococrw/mac.h
```diff
@@ -0,0 +1,72 @@
...
+#include <memory>
+#include <vector>
+#include "openssl_wrap.h"
+
+namespace mococrw
+{
+
+/**
+ * @brief The MessageAuthenticationCode class is the abstract base class for the different implementation of
+ * message authentication code.
```
```suggestion
 * message authentication codes.
```

src/mococrw/mac.h
```diff
@@ -0,0 +1,72 @@
...
+public:
+    ~MessageAuthenticationCode() {}
+
+    /**
+     * @brief Calculates the tag/mac based on the key, the message and the parameters provided to the constructor
+     * @param key The which shall be used for calculation
+     * @param message The message which shall be authenticated
+     * @return The message authentication tag
+     */
+    virtual std::vector<uint8_t> calculate(const std::vector<uint8_t>& key, const std::vector<uint8_t>& message) = 0;
```
neverpanic, line 41: That interface won't work. Consider the streaming case, where we get the ciphertext in blocks and return it to the user without caching it until finalize is called. We need a similar `update()` and `finish()` based API here.

Luckily, OpenSSL will force you to do this approach anyway in the implementation, so it shouldn't be complicated to just pass this along to the user.

src/mococrw/mac.h
```diff
@@ -0,0 +1,72 @@
...
+};
+
+class HMAC : public MessageAuthenticationCode{
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     * @param tagBitsLength The length of the tag/mac in number of bits
+     */
+    HMAC(openssl::DigestTypes hashFunction, uint32_t tagBitsLength);
```
neverpanic, line 51:
```suggestion
    HMAC(openssl::DigestTypes hashFunction, size_t tagBitsLength);
```

neverpanic (changes requested):

wusto:
No needed any longer.

PR #58, opened 2020-03-23
Add KDF schemes PBKDF2 and X9.63

wusto:
Adding an C++ interface for PBKDF2 and X9.63.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -22,6 +22,7 @@ set(LIBRARY_SOURCES
     csr.cpp
     distinguished_name.cpp
     hash.cpp
+    kdf.cpp
     key.cpp
     key_usage.cpp
     openssl_lib.cpp
@@ -45,6 +46,7 @@ set(LIBRARY_PUBLIC_HEADERS
     mococrw/error.h
     mococrw/extension.h
     mococrw/hash.h
+    mococrw/kdf.h
     mococrw/key.h
     mococrw/key_usage.h
     mococrw/openssl_lib.h
--- a/src/kdf.cpp
+++ b/src/kdf.cpp
@@ -1,0 +1,136 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2020 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include "mococrw/kdf.h"
+#include "mococrw/openssl_wrap.h"
+
+namespace mococrw {
+
+KeyDerivationFunction::~KeyDerivationFunction() = default;
+
+
+
+
+class PBKDF2::Impl
+{
+public:
+    Impl(openssl::DigestTypes hashFunction, uint32_t iterations):
+        _hashFunction(hashFunction), _iterations(iterations) {}
+
+    ~Impl() = default;
+
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password,
+                                   const size_t outputLength,
+                                   const std::vector<uint8_t> &salt)
+    {
+        std::vector<uint8_t> derivedKey(outputLength);
+        const EVP_MD* digestFn = _getMDPtrFromDigestType(_hashFunction);
+        openssl::_PKCS5_PBKDF2_HMAC(password, salt, _iterations, digestFn, derivedKey);
+
+        return derivedKey;
+    }
+
+private:
+    openssl::DigestTypes _hashFunction;
+    uint32_t _iterations;
+};
+
+PBKDF2::PBKDF2(openssl::DigestTypes hashFunction, uint32_t iterations)
+{
+    _impl = std::make_unique<PBKDF2::Impl>(hashFunction, iterations);
+}
+
+PBKDF2::~PBKDF2() = default;
+
+std::vector<uint8_t> PBKDF2::deriveKey(const std::vector<uint8_t> &password,
+                                       const size_t outputLength,
+                                       const std::vector<uint8_t> &salt)
+{
+    return _impl->deriveKey(password, outputLength, salt);
+}
+
+PBKDF2::PBKDF2(PBKDF2&& other) = default;
+
+PBKDF2& PBKDF2::operator=(PBKDF2&& other) {
+    this->_impl = std::move(other._impl);
+    return *this;
+}
+
+PBKDF2::PBKDF2(const PBKDF2& other)
+    : _impl(std::make_unique<PBKDF2::Impl>(*other._impl)) {}
+
+PBKDF2& PBKDF2::operator=(const PBKDF2& other) {
+    this->_impl = std::make_unique<PBKDF2::Impl>(*other._impl);
+    return *this;
+}
+
+
+
+
+class X963KDF::Impl
+{
+public:
+    Impl(openssl::DigestTypes hashFunction) : _hashFunction(hashFunction) {}
+
+    ~Impl() = default;
+
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password,
+                                   const size_t outputLength,
+                                   const std::vector<uint8_t> &salt)
+    {
+        std::vector<uint8_t> derivedKey(outputLength);
+        const EVP_MD* digestFn = _getMDPtrFromDigestType(_hashFunction);
+        openssl::_ECDH_KDF_X9_63(derivedKey, password, salt, digestFn);
+        return derivedKey;
+    }
+
+private:
+    openssl::DigestTypes _hashFunction;
+};
+
+X963KDF::X963KDF(openssl::DigestTypes hashFunction)
+{
+    _impl = std::make_unique<X963KDF::Impl>(hashFunction);
+}
+
+X963KDF::~X963KDF() = default;
+
+std::vector<uint8_t> X963KDF::deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                        const std::vector<uint8_t> &salt)
+{
+    return _impl->deriveKey(password, outputLength, salt);
+}
+
+X963KDF::X963KDF(X963KDF&& other) = default;
+
+X963KDF::X963KDF(const X963KDF &other)
+    : _impl(std::make_unique<X963KDF::Impl>(*other._impl)) {}
+
+X963KDF& X963KDF::operator=(X963KDF&& other)
+{
+    this->_impl = std::move(other._impl);
+    return *this;
+}
+
+X963KDF& X963KDF::operator=(const X963KDF &other)
+{
+    this->_impl = std::make_unique<X963KDF::Impl>(*other._impl);
+    return *this;
+}
+
+}
--- a/src/mococrw/kdf.h
+++ b/src/mococrw/kdf.h
@@ -1,0 +1,171 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2020 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+#include <memory>
+#include <vector>
+#include <cstdint>
+#include "hash.h"
+
+namespace mococrw
+{
+
+/**
+ * @brief Abstract class for key derivation functions
+ */
+class KeyDerivationFunction
+{
+public:
+    /**
+     * @brief destructor
+     */
+    virtual ~KeyDerivationFunction();
+    /**
+     * @brief Provides a derived key.
+     *
+     * This function returns a derived key based on the arguments passed to this function and the arguments provided to
+     * the constructor
+     * @param password The password/secret key which shall be used for key derivation
+     * @param outputLength The length of the derived key in bytes.
+     * @param salt The salt used for the key derivation
+     * @throws OpenSSLException This method may throw an OpenSSLException if OpenSSL indicates an error.
+     * @return The derived key
+     */
+    virtual std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                           const std::vector<uint8_t> &salt) = 0;
+};
+
+class PBKDF2: public KeyDerivationFunction {
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     * @param iterations Number of iterations performed during deriveKey
+     */
+    PBKDF2(openssl::DigestTypes hashFunction, uint32_t iterations);
+
+    /**
+     * @brief destructor
+     */
+    ~PBKDF2();
+
+    /**
+     * @see KeyDerivationFunction::deriveKey
+     */
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                   const std::vector<uint8_t> &salt) override;
+
+    /**
+     * @brief The move constructor
+     * @param other The object to move
+     */
+    PBKDF2(PBKDF2 &&other);
+
+    /**
+     * @brief The copy constructor
+     * @param other The object to copy
+     */
+    PBKDF2(const PBKDF2 &other);
+
+    /**
+     * @brief The move assignment operator
+     * @param other The object to move
+     * @return The result of the assignment
+     */
+    PBKDF2 &operator=(PBKDF2 &&other);
+
+    /**
+     * @brief The copy assignment operator
+     * @param other The object to copy
+     * @return The result of the assignment
+     */
+    PBKDF2& operator=(const PBKDF2 &other);
+
+private:
+    /**
+     * Internal class for applying the PIMPL design pattern
+     * (to hide the details of storing the padding objects from the client)
+     */
+    class Impl;
+
+    /**
+     * Pointer for PIMPL design pattern
+     */
+    std::unique_ptr<Impl> _impl;
+};
+
+class X963KDF : public KeyDerivationFunction
+{
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     */
+    X963KDF(openssl::DigestTypes hashFunction);
+
+    /**
+     * @brief destructor
+     */
+    ~X963KDF();
+
+    /**
+     * @see KeyDerivationFunction::deriveKey
+     */
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                   const std::vector<uint8_t> &salt) override;
+
+    /**
+     * @brief The move constructor
+     * @param other The object to move
+     */
+    X963KDF(X963KDF &&other);
+
+    /**
+     * @brief The copy constructor
+     * @param other The object to copy
+     */
+    X963KDF(const X963KDF &other);
+
+    /**
+     * @brief The move assignment operator
+     * @param other The object to move
+     * @return The result of the assignment
+     */
+    X963KDF &operator=(X963KDF &&other);
+
+    /**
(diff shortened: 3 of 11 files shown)


src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+{
+public:
+    /**
+     * @brief destructor
+     */
+    virtual ~KeyDerivationFunction();
+    /**
+     * @brief Provides a derived key.
+     *
+     * This function returns a derived key based on the arguments to this function and the arguments provided to
```
blenk92:
```suggestion
     * This function returns a derived key based on the arguments passed to this function and the arguments provided to
```

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+     * the constructor
+     * @param password The password/secret key which shall be used for key derivation
+     * @param outputLength The length of the derived key in bytes.
+     * @param salt The salt used for the key derivation
+     * @throws This method may throw an OpenSSLException if OpenSSL indicates an error.
```
blenk92:
```suggestion
     * @throws OpenSSLException if OpenSSL indicates an error.
```
see syntax of `@throws` at http://www.doxygen.nl/manual/commands.html#cmdthrows

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+
+class PBKDF2: public KeyDerivationFunction {
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     */
+    PBKDF2(openssl::DigestTypes hashFunction, uint32_t iterations);
+
+    ~PBKDF2();
```
blenk92, line 65: doxygen (at least it's there in X963KDF) :-D

src/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ set(LIBRARY_SOURCES
     csr.cpp
```
neverpanic, line 22: Since I can't comment on commit messages: Our commit message guidelines recommend using imperative form, i.e. "Add KDF schemes PBKDF2 and ANSI X9.63", not "Adding KDF schemes".

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * @brief Provides a derived key.
+     *
+     * This function returns a derived key based on the arguments to this function and the arguments provided to
```
blenk92:
```suggestion
     * This function returns a derived key based on the arguments passed to this function and the arguments provided to
```

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+     * the constructor
+     * @param password The password/secret key which shall be used for key derivation
+     * @param outputLength The length of the derived key in bytes.
+     * @param salt The salt used for the key derivation
+     * @throws This method may throw an OpenSSLException if OpenSSL indicates an error.
```
blenk92: same as above, won't mark in the following (if there are more)

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+                                           const std::vector<uint8_t> &salt) = 0;
+};
+
+class PBKDF2: public KeyDerivationFunction {
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     */
+    PBKDF2(openssl::DigestTypes hashFunction, uint32_t iterations);
```
neverpanic, line 60: Please document the iterations parameter.

Please `#include <cstdint>` for `uint32_t`.

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+
+    ~PBKDF2();
+
+    /**
+     * @brief Provides a derived key.
```
neverpanic: In previous cases where we had documentation in the superclass, we documented implementations using a `@see` comment rather than reproducing the entire documentation. Maybe we should do the same thing here?

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+     * @param password The password/secret key which shall be used for key derivation
+     * @param outputLength The length of the derived key in bytes.
+     * @param salt The salt used for the key derivation
+     * @throws This method may throw an OpenSSLException if OpenSSL indicates an error.
+     * @return The derived key
+     */
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                   const std::vector<uint8_t> &salt) override;
+
+private:
```
neverpanic, line 99: Do you want to PIMPL this?

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+    openssl::DigestTypes hashFunction;
+    uint32_t iterations;
+    std::vector<uint8_t>_derivedKey;
```
neverpanic:
```suggestion
    std::vector<uint8_t> _derivedKey;
```

Also, why is this a member variable at all? I think it may be enough that we're returning this from `deriveKey()`.

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+     */
+    X963KDF(openssl::DigestTypes hashFunction);
+
+    /**
+     * @brief destructor
+     */
+    ~X963KDF();
+
+    /**
+     * @brief Provides a derived key.
```
neverpanic: Ditto.

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+     * @param password The password/secret key which shall be used for key derivation
+     * @param outputLength The length of the derived key in bytes.
+     * @param salt The salt used for the key derivation
+     * @throws This method may throw an OpenSSLException if OpenSSL indicates an error.
+     * @return The derived key
+     */
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                   const std::vector<uint8_t> &salt) override;
+
+private:
```
neverpanic, line 158: Ditto.

src/mococrw/kdf.h
```diff
@@ -0,0 +1,117 @@
...
+    openssl::DigestTypes hashFunction;
+    uint32_t iterations;
+    std::vector<uint8_t>_derivedKey;
```
neverpanic:
```suggestion
    std::vector<uint8_t> _derivedKey;
```

Ditto, do we really need this?

src/kdf.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+KeyDerivationFunction::~KeyDerivationFunction() = default;
+
+PBKDF2::PBKDF2(openssl::DigestTypes hashFunction, uint32_t iterations) :
+    hashFunction(hashFunction), iterations(iterations) {}
+
+PBKDF2::~PBKDF2() = default;
+
+std::vector<uint8_t> PBKDF2::deriveKey(const std::vector<uint8_t> &password,
+                                                const size_t outputLength,
```
neverpanic: Please fix indentation.

src/openssl_lib.cpp
```diff
@@ -910,6 +910,31 @@ int OpenSSLLib::SSL_EVP_CIPHER_CTX_set_padding(EVP_CIPHER_CTX * c, int pad) noex
...
+{
+    return PKCS5_PBKDF2_HMAC(pass, passlen, salt, saltlen, iter, digest, keylen, out);
+}
+
+
+int OpenSSLLib::SSL_ECDH_KDF_X9_63(unsigned char *out, size_t outlen, const unsigned char *Z, size_t Zlen,
+                                   const unsigned char *sinfo, size_t sinfolen, const EVP_MD *md) noexcept
+{
+    /*
+     * The old name for ecdh_KDF_X9_63 (include/openssl/ec.h:1110)
```
blenk92, line 925: still not sure how to handle this, maybe we should really use https://www.openssl.org/docs/manmaster/man3/EVP_KDF_derive.html instead using a (sonnish to be) deprecated function. On the other hand this is what ecdh_KDF_X9_63/2 uses anyway...

neverpanic (changes requested):

tests/unit/test_kdf.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+void testX963kdf(struct inputData data)
+{
+    auto x963kdf = X963KDF(data.hashFunction);
+    ASSERT_THAT(utility::toHex(x963kdf.deriveKey(data.password, data.olen, data.salt)),
+                Eq(data.expectedRestult));
+}
+
+TEST(pbkdf2_sha1, tests)
```
blenk92: please use TEST_P to avoid these lengthy tests that test 3 times the same thing

tests/unit/test_kdf.cpp
```diff
@@ -0,0 +1,197 @@
...
+        4096,
+        64,
+        "d197b1b33db0143e018b12f3d1d1479e6cdebdcc97c5c0f87f6902e072f457b5143f30602641b3d55cd335988cb36b84376060ecd532"\
+        "e039b742a239434af2d5"
+    };
+
+    testPbkfd2(test);
+}
+
+TEST(x963kdf, tests)
```
blenk92: same here

blenk92 (dismissed):

tests/unit/test_kdf.cpp
```diff
@@ -0,0 +1,197 @@
...
+TEST(pbkdf2_sha512, tests)
+{
+    /* Test vectors taken from here: https://stackoverflow.com/questions/15593184/pbkdf2-hmac-sha-512-test-vectors */
+    struct inputData test = {
+         openssl::DigestTypes::SHA512,
+        {'p', 'a', 's', 's', 'w', 'o', 'r', 'd'},
+        {'s', 'a', 'l', 't'},
+        1,
+        64,
+        "867f70cf1ade02cff3752599a3a53dc4af34c7a669815ae5d513554e1c8cf252c02d470a285a0501bad999bfe943c08f050235d7d68b"\
```
neverpanic: To concatenate static strings in C/C++, you do not need the line continuation character `\`.

neverpanic (commented):
Hadn't looked at the tests.

src/mococrw/kdf.h
wusto: done

src/mococrw/kdf.h
wusto: done

src/mococrw/kdf.h
wusto: Don't know if it makes sense. The implementation is 4 lines of code.

src/mococrw/kdf.h
wusto: done

src/mococrw/kdf.h
wusto: done

src/mococrw/kdf.h
wusto: done

src/mococrw/kdf.h
wusto: done

src/kdf.cpp
wusto: done

src/mococrw/kdf.h
wusto: done

src/mococrw/kdf.h
wusto: done

src/mococrw/kdf.h
wusto: Done

src/mococrw/kdf.h
wusto: It's pimpled :)

tests/unit/test_kdf.cpp
wusto: done

tests/unit/test_kdf.cpp
wusto: That was easy :)

src/kdf.cpp
```diff
@@ -0,0 +1,125 @@
...
+    Impl(openssl::DigestTypes hashFunction, uint32_t iterations):
+        _hashFunction(hashFunction), _iterations(iterations) {}
+
+    ~Impl() = default;
+
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password,
+                                   const size_t outputLength,
+                                   const std::vector<uint8_t> &salt)
+    {
+        std::vector<uint8_t> _derivedKey;
```
neverpanic: Now that `derivedKey` is no longer a class member, I'd prefer if its name didn't start with an underscore.

src/kdf.cpp
```diff
@@ -0,0 +1,125 @@
...
+public:
+    Impl(openssl::DigestTypes hashFunction) : _hashFunction(hashFunction) {}
+
+    ~Impl() = default;
+
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password,
+                                   const size_t outputLength,
+                                   const std::vector<uint8_t> &salt)
+    {
+        std::vector<uint8_t> _derivedKey;
```
neverpanic: Ditto.

tests/unit/test_kdf.cpp
```diff
@@ -0,0 +1,201 @@
...
+TEST_P(Pbkdf2Tests, tests)
+{
+    testPbkfd2(GetParam());
+}
+
+
+
+
+
+
```
neverpanic: Let's be a bit more swabian with the whitespace between those two functions?

neverpanic (commented):
Nitpicks, other than that lg2m. Oh, and please don't forget the commit message.

src/CMakeLists.txt
wusto: I will do this in the final commit, which will be a force push.

neverpanic (approved):
Please squash.

tests/unit/test_kdf.cpp
wusto: Sure. :)
Just thought they would help to differentiate between the PBKDF2 and X9.63 tests.

src/mococrw/kdf.h
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * @see KeyDerivationFunction::deriveKey
+     */
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                   const std::vector<uint8_t> &salt) override;
+
+    /**
+     * @brief The move constructor
+     * @param other
```
neverpanic:
```suggestion
     * @param other The object to move
```

src/kdf.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    Impl(Impl&) = default;
+
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password,
+                                   const size_t outputLength,
+                                   const std::vector<uint8_t> &salt)
+    {
+        std::vector<uint8_t> derivedKey;
+        const EVP_MD* digestFn = _getMDPtrFromDigestType(_hashFunction);
+        derivedKey.resize(outputLength);
```
neverpanic:
```suggestion
        std::vector<uint8_t> derivedKey(outputLength);
        const EVP_MD* digestFn = _getMDPtrFromDigestType(_hashFunction);
```

This saves a realloc.

src/kdf.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    ~Impl() = default;
+
+    std::vector<uint8_t> deriveKey(const std::vector<uint8_t> &password,
+                                   const size_t outputLength,
+                                   const std::vector<uint8_t> &salt)
+    {
+        std::vector<uint8_t> derivedKey;
+        const EVP_MD* digestFn = _getMDPtrFromDigestType(_hashFunction);
+        derivedKey.resize(outputLength);
```
neverpanic:
```suggestion
        std::vector<uint8_t> derivedKey(outputLength);
        const EVP_MD* digestFn = _getMDPtrFromDigestType(_hashFunction);
```

neverpanic (commented):
There's always some other nitpick to find that I didn't see in the first round. As a courtesy, I'll give you patches.

src/openssl_lib.cpp
wusto: As discussed we will resolve this issue.

src/kdf.cpp
wusto: done

src/mococrw/kdf.h
wusto: done

src/kdf.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+public:
+    Impl(openssl::DigestTypes hashFunction, uint32_t iterations):
+        _hashFunction(hashFunction), _iterations(iterations) {}
+
+    ~Impl() = default;
+
+    Impl(Impl&&) = default;
+
+    Impl(Impl&) = default;
```
garw: This is not the copy constructor?

src/kdf.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+class PBKDF2::Impl
+{
+public:
+    Impl(openssl::DigestTypes hashFunction, uint32_t iterations):
+        _hashFunction(hashFunction), _iterations(iterations) {}
+
+    ~Impl() = default;
+
+    Impl(Impl&&) = default;
```
garw: Why do we even need this here? From the looks of it, we never move the Impl class but only the unique_ptr that holds it?

src/kdf.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+PBKDF2::PBKDF2(PBKDF2&& other) = default;
+
+PBKDF2& PBKDF2::operator=(PBKDF2&& other) {
+    this->_impl = std::move(other._impl);
+    return *this;
+}
+
+PBKDF2::PBKDF2(const PBKDF2& other)
+    : _impl(std::make_unique<PBKDF2::Impl>(*(other._impl))) {}
```
garw: Noise: The brackets around other._impl should not be needed. The dot binds stronger than *

src/kdf.cpp
garw: We are also missing these in the other Impl class for X963…

src/kdf.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+X963KDF::X963KDF(const X963KDF &other)
+    : _impl(std::make_unique<X963KDF::Impl>(*(other._impl))) {}
+
+X963KDF& X963KDF::operator=(X963KDF&& other) {
+    this->_impl = std::move(other._impl);
+    return *this;
+}
+
+X963KDF &X963KDF::operator=(const X963KDF &other)
```
garw: Style: & close to type and not to function name

src/kdf.cpp
```diff
@@ -0,0 +1,142 @@
...
+std::vector<uint8_t> X963KDF::deriveKey(const std::vector<uint8_t> &password, const size_t outputLength,
+                                        const std::vector<uint8_t> &salt)
+{
+    return _impl->deriveKey(password, outputLength, salt);
+}
+
+X963KDF::X963KDF(X963KDF&& other) = default;
+
+X963KDF::X963KDF(const X963KDF &other)
+    : _impl(std::make_unique<X963KDF::Impl>(*(other._impl))) {}
```
garw: s.a.

src/kdf.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+X963KDF& X963KDF::operator=(X963KDF&& other) {
+    this->_impl = std::move(other._impl);
+    return *this;
+}
+
+X963KDF &X963KDF::operator=(const X963KDF &other)
+{
+    this->_impl = std::make_unique<X963KDF::Impl>(*(other._impl));
```
garw: s.a.

src/mococrw/openssl_wrap.h
```diff
@@ -1337,5 +1337,15 @@ enum class ellipticCurveNid
 };
 
 EC_KEY *_EVP_PKEY_get0_EC_KEY(EVP_PKEY *pkey);
+
+void _PKCS5_PBKDF2_HMAC(const char *pass, int passlen,
```
garw: The intend of this library layer is to provide as good memory management as possible in the situation with C++ and OpenSSL. This is, e.g., visible in all the places where it returns smart pointers to OpenSSL objects.

Consequently, I would prefer this function to already manage the output buffer and return a vector<uint_8>
I know that this pattern is not a 100% correct applied but no reason to make the overall situation worse.
Same for the other KDF

garw (dismissed):
A few minor things but one general observation regarding this library's design: openssl_wrap is supposed to provide not only error handling but also as good as possible memory management and memory safety.

src/kdf.cpp
wusto: done

src/kdf.cpp
wusto: You are right. They are not necessary as the class members have an implicit copy and move constructor, so that the compiler is able to generate one for Impl.

src/kdf.cpp
wusto: Done

src/kdf.cpp
wusto: Done

src/kdf.cpp
wusto: done

src/mococrw/openssl_wrap.h
wusto: done

src/kdf.cpp
wusto: done

neverpanic (approved):
lg2m

neverpanic (approved):

PR #59, opened 2020-03-23
Add MAC scheme HMAC

wusto:
This change adds HMAC as a MAC scheme.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,6 +25,7 @@ set(LIBRARY_SOURCES
     kdf.cpp
     key.cpp
     key_usage.cpp
+    mac.cpp
     openssl_lib.cpp
     openssl_wrap.cpp
     subject_key_identifier.cpp
@@ -49,6 +50,7 @@ set(LIBRARY_PUBLIC_HEADERS
     mococrw/kdf.h
     mococrw/key.h
     mococrw/key_usage.h
+    mococrw/mac.h
     mococrw/openssl_lib.h
     mococrw/openssl_wrap.h
     mococrw/padding_mode.h
--- a/src/mac.cpp
+++ b/src/mac.cpp
@@ -1,0 +1,122 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2020 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include "mococrw/mac.h"
+#include "mococrw/error.h"
+#include "mococrw/openssl_wrap.h"
+
+#include <openssl/hmac.h>
+#include <openssl/crypto.h>
+
+namespace mococrw
+{
+
+MessageAuthenticationCode::~MessageAuthenticationCode() = default;
+
+class HMAC::Impl
+{
+public:
+    Impl(openssl::DigestTypes hashFunction, const std::vector<uint8_t> key)
+    {
+        const EVP_MD* digestFn = _getMDPtrFromDigestType(hashFunction);
+
+        if (key.empty()) {
+            throw MoCOCrWException("Key for HMAC is empty.");
+        }
+
+        _ctx = openssl::_HMAC_CTX_new();
+        openssl::_HMAC_Init_ex(_ctx.get(), key, digestFn, NULL);
+    }
+
+    ~Impl() = default;
+
+    Impl(Impl&&) = default;
+
+    void update(const std::vector<uint8_t> &message)
+    {
+        if (_isFinished) {
+            throw MoCOCrWException("update() can't be called after finish()");
+        }
+        openssl::_HMAC_Update(_ctx.get(), message);
+    }
+
+    std::vector<uint8_t> finish()
+    {
+        if (_isFinished) {
+            throw MoCOCrWException("finish() can't be called twice.");
+        }
+
+        _result = openssl::_HMAC_Final(_ctx.get());
+
+        _isFinished = true;
+
+        return _result;
+    }
+
+    void verify(const std::vector<uint8_t> &hmacValue)
+    {
+        if (!_isFinished) {
+            finish();
+        }
+
+        if (hmacValue.size() != _result.size()) {
+            throw MoCOCrWException("HMAC verification failed. Length differs.");
+        }
+
+        if (CRYPTO_memcmp(hmacValue.data(), _result.data(), hmacValue.size())) {
+            throw MoCOCrWException("HMAC verification failed. Calculated value: " + utility::toHex(_result)
+                                   + ". Received value: " + utility::toHex(hmacValue));
+        }
+    }
+
+private:
+    openssl::SSL_HMAC_CTX_SharedPtr _ctx = nullptr;
+    bool _isFinished = false;
+    std::vector<uint8_t> _result;
+};
+
+HMAC::HMAC(mococrw::openssl::DigestTypes hashFunction, const std::vector<uint8_t>& key)
+{
+    _impl = std::make_unique<HMAC::Impl>(hashFunction, key);
+}
+
+HMAC::~HMAC() = default;
+
+HMAC::HMAC(HMAC&& other) = default;
+
+HMAC& HMAC::operator=(HMAC&& other) {
+    this->_impl = std::move(other._impl);
+    return *this;
+}
+
+void HMAC::update(const std::vector<uint8_t> &message)
+{
+    _impl->update(message);
+}
+
+std::vector<uint8_t> HMAC::finish()
+{
+    return _impl->finish();
+}
+
+void HMAC::verify(const std::vector<uint8_t> &hmacValue)
+{
+    _impl->verify(hmacValue);
+}
+
+}
--- a/src/mococrw/mac.h
+++ b/src/mococrw/mac.h
@@ -1,0 +1,149 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2020 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+#include <memory>
+#include <vector>
+#include "openssl_wrap.h"
+
+namespace mococrw
+{
+
+/**
+ * @brief The MessageAuthenticationCode class is the abstract base class for the different implementation of
+ * message authentication codes.
+ */
+class MessageAuthenticationCode {
+public:
+
+    /**
+     * @brief ~MessageAuthenticationCode
+     */
+    virtual ~MessageAuthenticationCode();
+
+    /**
+     * Adds the message to the hash. This function may be invoked multiple times.
+     * For finishing the calculation of the MAC invoke finish().
+     *
+     * @param message chunk of data used for MAC.
+     * @throws MoCOCrWException if this function is invoked after finish was called.
+     */
+    virtual void update(const std::vector<uint8_t>& message) = 0;
+
+    /**
+     * @brief Finalize the MAC.
+     *
+     * This function calculates the message authentication code.
+     *
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The calculated message authentication code.
+     */
+    virtual std::vector<uint8_t> finish() = 0;
+
+    /**
+     * @brief Verifies the MAC
+     *
+     * This function compares the given value (macValue), with the value calculated during finish().
+     * Invoke finish() before calling verify().
+     * The length of the calculated value and the given value has to be the same!
+     * This function also finishes the MAC if finish is not invoked in advance.
+     *
+     * The comparison happens in constant time.
+     * @throws MoCOCrWException if the verification fails because the values or their lengths differ.
+     * @param hmacValue The value which shall be compared to the calculated value
+     */
+    virtual void verify(const std::vector<uint8_t> &macValue) = 0;
+};
+
+class HMAC : public MessageAuthenticationCode {
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     * @param key The key used for HMAC
+     */
+    HMAC(mococrw::openssl::DigestTypes hashFunction, const std::vector<uint8_t> &key);
+
+    /**
+     * @brief destructor
+     */
+    ~HMAC();
+
+    /**
+     * Adds the chunk of data to the hash. H(i-key || message1 [ || message2])
+     * For getting the HMAC invoke finish()
+     *
+     * @param message chunk of data to used for HMAC.
+     * @throws MoCOCrWException if this function is invoked after finish was called.
+     */
+    void update(const std::vector<uint8_t>& message) override;
+
+
+    /**
+     * @brief Finalize the HMAC.
+     *
+     * This function calculates the message authentication code.
+     * H(o-key || H(i-key || message [ || message]))
+     *
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The hashed message authentication code
+     */
+    std::vector<uint8_t> finish() override;
+
+    /**
+     * @see MessageAuthenticationCode::verify
+     */
+    void verify(const std::vector<uint8_t> &hmacValue) override;
+
+    /**
+     * @brief The move constructor
+     * @param other The other HMAC to be moved
+     */
+    HMAC(HMAC &&other);
+
+    /**
+     * @brief The assignment operator
+     * @param other The other HMAC to be assigned
+     * @return The result of the assignment
+     */
+    HMAC &operator=(HMAC &&other);
+
+    /**
+     * @brief Delete the copy constructor.
+     */
+    HMAC(const HMAC &other) = delete;
+
+    /**
+     * @brief Delete the copy assignment
+     */
+    HMAC & operator=(const HMAC&) = delete;
+
+private:
+    /**
+     * Internal class for applying the PIMPL design pattern
+     * (to hide the details of storing the padding objects from the client)
+     */
+    class Impl;
+
+    /**
+     * Pointer for PIMPL design pattern
+     */
+    std::unique_ptr<Impl> _impl;
+};
+
+}
--- a/src/openssl_wrap.cpp
+++ b/src/openssl_wrap.cpp
@@ -601,6 +601,12 @@ X509_CRL *createOpenSSLObject<X509_CRL>()
 }
 
 template<>
+HMAC_CTX *createOpenSSLObject<HMAC_CTX>()
+{
+    return OpensslCallPtr::callChecked(lib::OpenSSLLib::SSL_HMAC_CTX_new);
+}
+
+template<>
 void addObjectToStack<STACK_OF(X509), X509>(STACK_OF(X509) *stack, const X509 *obj)
 {
     OpensslCallIsPositive::callChecked(lib::OpenSSLLib::SSL_sk_X509_push, stack, obj);
@@ -1260,6 +1266,30 @@ void _ECDH_KDF_X9_63(std::vector<uint8_t> &out, const std::vector<uint8_t> &Z, c
     const unsigned char *sinfo_ = reinterpret_cast<const unsigned char*>(sinfo.data());
     OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_ECDH_KDF_X9_63, out_, out.size(), Z_, Z.size(),
                                   sinfo_, sinfo.size(), md);
+}
+
+void _HMAC_Init_ex(HMAC_CTX *ctx, const std::vector<uint8_t> &key, const EVP_MD *md, ENGINE *impl)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Init_ex, ctx, reinterpret_cast<const void *>(key.data()),
+                                  key.size(), md, impl);
+}
+
+std::vector<uint8_t> _HMAC_Final(HMAC_CTX* ctx) {
+    unsigned int length = EVP_MAX_MD_SIZE;
+    std::vector<uint8_t> md(length);
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Final, ctx, md.data(), &length);
+    md.resize(length);
+    return md;
+}
+
+void _HMAC_Update(HMAC_CTX *ctx, const std::vector<uint8_t> &data)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Update, ctx, data.data(), data.size());
+}
+
+SSL_HMAC_CTX_Ptr _HMAC_CTX_new()
+{
+    return createManagedOpenSSLObject<SSL_HMAC_CTX_Ptr>();
 }
 
 }  // ::openssl
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -112,6 +112,11 @@ if(BUILD_TESTING) #set by Ctest. Also set in the integration build environment.
 	"${SRC_DIR}/kdf.cpp"
 	${REAL_SOURCES})
 
+    add_executable(mactests test_mac.cpp
+	"${SRC_DIR}/mac.cpp"
+	"${SRC_DIR}/hash.cpp"
+	${REAL_SOURCES})
+
     target_link_libraries(openssltest ${GMOCK_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
 
     target_link_libraries(asn1timetests
@@ -148,7 +153,8 @@ if(BUILD_TESTING) #set by Ctest. Also set in the integration build environment.
         ${GMOCK_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT}  Boost::boost)
     target_link_libraries(kdftests
 	${GMOCK_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} OpenSSL::Crypto OpenSSL::SSL Boost::boost)
-
+    target_link_libraries(mactests
+	${GMOCK_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} OpenSSL::Crypto OpenSSL::SSL Boost::boost)
 
     add_test(
         NAME OpenSSLTest
@@ -230,6 +236,10 @@ if(BUILD_TESTING) #set by Ctest. Also set in the integration build environment.
     add_test(
 	NAME KdfSchemesTest
 	COMMAND kdftests
+    )
+    add_test(
+	NAME MacSchemesTest
+	COMMAND mactests
     )
 
     # Configure valgrind
--- a/tests/unit/test_mac.cpp
+++ b/tests/unit/test_mac.cpp
@@ -1,0 +1,195 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2020 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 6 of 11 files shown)


src/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ set(LIBRARY_SOURCES
     kdf.cpp
```
neverpanic, line 25: Same as in #58: The commit message should use imperative present tense: "Add MAC [...]"

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * @brief The MessageAuthenticationCode class is the abstract base class for the different implementation of
+ * message authentication codes.
+ */
+class MessageAuthenticationCode {
+public:
+    /**
+     * Adds the message to the hash. This function may be invoked multiple times.
+     * For finishing the calculation of the MAC-tag invoke finish().
```
```suggestion
     * For finishing the calculation of the MAC invoke finish().
```

I think "message authentication code" is the common term for both the algorithm and it's bytestring return value, so we should probably use MAC here. "tag" is only used in AES-GCM (and potentially other symmetric AEAD schemes) to my knowledge.

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param message chunk of data used for MAC.
+     * @throws MoCOCrWException if this function is invoked after finish was called.
+     */
+    void update(const std::vector<uint8_t>& message);
+
+    /**
+     * @brief Finalize the HMAC.
+     *
+     * This function calculates the hashed messaged authentication code.
```
```suggestion
     * This function calculates the message authentication code.
```

"hashed message authentication code" may be mis-interpreted as calculating the MAC and hashing the generated MAC. Let's avoid this ambiguity and just call it "message authentication code".

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The hashed message authentication code tag (MAC-tag)
```
```suggestion
     * @return The calculated message authentication code
```

For the same reasons as outlined above.

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     */
+    std::vector<uint8_t> finish();
+};
+
+class HMAC : public MessageAuthenticationCode{
```
```suggestion
class HMAC : public MessageAuthenticationCode {
```

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     * @param tagBytesLength The length of the tag/mac in number of bytes (multiple of 8 bits)
```
```suggestion
     * @param tagBytesLength The length of the MAC in bytes
```

If you specify the length in bytes, there is no way you could specify a value that isn't a multiple of 8 bits, so I'd just leave this out of the documentation. We're targeting software devs with experience in crypto functionality with this interface, so I think we can assume the bits-to-bytes relation as given.

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     */
+    HMAC(mococrw::openssl::DigestTypes hashFunction, const std::vector<uint8_t> &key, size_t outputLength);
```
neverpanic: Please document the `outputLength` parameter.

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+#include "openssl_wrap.h"
+
+namespace mococrw
+{
+
+/**
+ * @brief The MessageAuthenticationCode class is the abstract base class for the different implementation of
+ * message authentication codes.
+ */
+class MessageAuthenticationCode {
```
neverpanic, line 31: Should this have a virtual d'tor?

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     * @param hashFunction The hash function which shall be used
+     * @param tagBytesLength The length of the tag/mac in number of bytes (multiple of 8 bits)
+     */
+    HMAC(mococrw::openssl::DigestTypes hashFunction, const std::vector<uint8_t> &key, size_t outputLength);
+
+    ~HMAC();
+
+    /**
+     * Adds the chunk of data to the hash. H(i-key || message1 [ || message2])
+     * For getting the HMAC-tag invoke finish()
```
```suggestion
     * For getting the HMAC invoke finish()
```

And everywhere below where you've used "tag".

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     * @param other
+     */
+    HMAC(const HMAC &other);
+
+    /**
+     * @brief The assignment operator
+     * @param other
+     * @return The result of the assignment
+     */
+    HMAC &operator =(const HMAC &other);
```
```suggestion
    HMAC &operator=(const HMAC &other);
```

is the common way of writing this that I'm used to, but maybe that's just me.

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     * H(o-key || H(i-key || message [ || message]))
+     *
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The hashed message authentication code tag (HMAC-tag)
+     */
+    std::vector<uint8_t> finish();
+
+    /**
+     * @brief The copy constructor
+     * @param other
```
neverpanic: Please add a description for the `other` parameter, e.g. "The other HMAC to be copied".

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     */
+    HMAC(const HMAC &other);
+
+    /**
+     * @brief The assignment operator
+     * @param other
```
neverpanic: Ditto.

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The hashed message authentication code tag (HMAC-tag)
+     */
+    std::vector<uint8_t> finish();
+
+    /**
+     * @brief The copy constructor
+     * @param other
+     */
+    HMAC(const HMAC &other);
```
neverpanic: Do we need (and/or want) this to be copyable?

src/mac.cpp
```diff
@@ -0,0 +1,112 @@
...
+            }
+            _result.resize(outputLength);
+        }
+
+        _isFinished = true;
+
+        return _result;
+    }
+
+private:
```
neverpanic, line 87: Why do some of the members start with an underscore, while others don't? Shouldn't we be consistent with naming the members?

IIRC this applied to #58 as well.

src/mac.cpp
```diff
@@ -0,0 +1,112 @@
...
+    Impl(openssl::DigestTypes hashFunction, const std::vector<uint8_t> key, size_t outputLength) :
+        outputLength(outputLength)
+    {
+        const EVP_MD* digestFn = _getMDPtrFromDigestType(hashFunction);
+
+        if (key.empty()) {
+            throw MoCOCrWException("Key for HMAC is empty.");
+        }
+
+        ctx = openssl::_HMAC_CTX_new();
```
neverpanic: I don't have a good overview of the advantages and disadvantages, but have you considered using the `EVP_*` digest interface to do this? See https://wiki.openssl.org/index.php/EVP_Message_Digests and https://wiki.openssl.org/index.php/EVP_Signing_and_Verifying#Calculating_HMAC for more information.

src/mac.cpp
```diff
@@ -0,0 +1,112 @@
...
+        _result.resize(EVP_MAX_MD_SIZE);
+
+        unsigned int tagLength;
+        openssl::_HMAC_Final(ctx.get(), _result.data(), &tagLength);
+
+        if (outputLength == 0) {
+            _result.resize(tagLength);
+        } else {
+            if (outputLength > tagLength) {
+                throw MoCOCrWException("output length shall be greater than the actual tag length.");
```
neverpanic: I think using `cannot be greater than` would make the problem easier to understand.

src/mac.cpp
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        _isFinished = true;
+
+        return _result;
+    }
+
+private:
+    openssl::SSL_HMAC_CTX_SharedPtr ctx = nullptr;
+    std::vector<uint8_t> _result;
```
neverpanic: I don't think `_result` needs to be a member here. You can create it in `finish()`.

src/openssl_wrap.cpp
```diff
@@ -1257,5 +1257,25 @@ void _ECDH_KDF_X9_63(unsigned char *out, size_t outlen, const unsigned char *Z,
...
+}
+
+void _HMAC_Final(HMAC_CTX *ctx, unsigned char *md, unsigned int *len)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Final, ctx, md ,len);
+}
+
+void _HMAC_Update(HMAC_CTX *ctx, const unsigned char *data, int len)
+{
+        OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Update, ctx, data, len);
```
neverpanic: Indentation differs from the other functions.

tests/unit/CMakeLists.txt
```diff
@@ -148,7 +152,8 @@ if(BUILD_TESTING) #set by Ctest. Also set in the integration build environment.
         ${GMOCK_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT}  Boost::boost)
     target_link_libraries(kdftests
 	${GMOCK_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} OpenSSL::Crypto OpenSSL::SSL Boost::boost)
-
+    target_link_libraries(mactests
+    ${GMOCK_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} OpenSSL::Crypto OpenSSL::SSL Boost::boost)
```
neverpanic: Indentation differs from the other lines.

tests/unit/test_mac.cpp
```diff
@@ -0,0 +1,119 @@
...
+using namespace mococrw;
+
+using testing::Eq;
+
+struct inputData
+{
+    std::string key;
+    std::string data;
+    size_t outputLength;
+    std::string expectedRestultSha256;
```
neverpanic: `s/Restult/Result/g`

tests/unit/test_mac.cpp
```diff
@@ -0,0 +1,119 @@
...
+        "b0344c61d8db38535ca8afceaf0bf12b881dc200c9833da726e9376c2e32cff7",
+        "afd03944d84895626b0825f4ab46907f15f9dadbe4101ec682aa034c7cebc59cfaea9ea9076ede7f4af152e8b2fa9cb6",
+        "87aa7cdea5ef619d4ff0b4241a1d6cb02379f4e2ce4ec2787ad0b30545e17cdedaa833b7d6b8a702038b274eaea3f4e4be9d914eeb61f1702e696c203a126854"
+    };
+
+    testHmac(testData);
+
+    /* Truncate output */
+    testData = {
+        "0c0c0c0c0c0c0c0c0c0c0c0c0c0c0c0c"\
```
neverpanic: The line continuations aren't necessary.

tests/unit/test_mac.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+void testHmac(struct inputData testData)
+{
+    testHmacSha(openssl::DigestTypes::SHA256, testData, testData.expectedRestultSha256);
+    testHmacSha(openssl::DigestTypes::SHA384, testData, testData.expectedRestultSha384);
+    testHmacSha(openssl::DigestTypes::SHA512, testData, testData.expectedRestultSha512);
+}
+
+TEST(hmac, tests)
```
neverpanic: I think this might also be a good place to use parametrized tests using `TEST_P` if you move the parameters and expected result into a struct.

neverpanic (changes requested):

src/mococrw/mac.h
wusto: yes, this should :)

src/mococrw/mac.h
wusto: If there is no copy operator something like 
auto hmac = mococrw::HMAC(hashFunction, utility::fromHex(testData.key), testData.outputLength);

Fails with:
error: use of deleted function ‘mococrw::HMAC::HMAC(const mococrw::HMAC&)

MoCOCrW/tests/unit/test_mac.cpp:44: error: call to implicitly-deleted copy constructor of 'mococrw::HMAC'

garw: But
`mococrw::HMAC hmac{hashFunction, utility::fromHex(testData.key};`
works.
If you want to support almost-always-auto a move constructor should suffice. All copy operations, I would honestly delete on these kind of objects.

src/mococrw/mac.h
```diff
@@ -0,0 +1,103 @@
...
+class MessageAuthenticationCode {
+public:
+    /**
+     * Adds the message to the hash. This function may be invoked multiple times.
+     * For finishing the calculation of the MAC-tag invoke finish().
+     *
+     * @param message chunk of data used for MAC.
+     * @throws MoCOCrWException if this function is invoked after finish was called.
+     */
+    void update(const std::vector<uint8_t>& message);
```
garw: If this is supposed to be an abstract base, class, these need to be virtual?

src/mac.cpp
wusto: I haven't considered it.
I checked the openssl tests for hmac (test/hmactest.c) and there the HMAC_CTX_new and HMAC_Init_ex interface is used. So I decided to use this, too.

src/mac.cpp
wusto: I have named all private members starting with an underscore.

src/mococrw/mac.h
wusto: it should be a pure virtual.
Will be fixed.

src/CMakeLists.txt
wusto: Will be done after squashing. :+1:

src/mococrw/mac.h
wusto: done

src/mac.cpp
wusto: I checked it again.
Internally the HMAC_* interface is invoking the EVP_* interface. So this is a "higher" level interface for calculating the MAC.

src/mococrw/mac.h
wusto: Removed the copy constructors and added move constructors and assignment operators.

src/mococrw/mac.h
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * This function calculates the messaged authentication code.
+     *
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The calculated message authentication code.
+     */
+    virtual std::vector<uint8_t> finish() = 0;
+
+    /**
+     * @brief Verifies the HMAC
```
garw: MAC vs HMAC?

src/mococrw/mac.h
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * This function compares the given value (hmacValue), with the value calculated during finish().
+     * Invoke finish() before calling verify().
+     * Only the characters of the shorter value (hmacValue or calculated value) are checked for equality with the
```
garw: Do we really want this or should verify() just implicitly cause a finish()? I think this is more natural to use in the verification scenario.
Then we have
update()
update()
finish() for creating
and
update()
update()
verify()
for the verification

src/mococrw/mac.h
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The hashed message authentication code
+     */
+    std::vector<uint8_t> finish() override;
+
+    /**
+     * @see MessageAuthenticationCode::verify
+     */
+    void verify(const std::vector<uint8_t> hmacValue) override;
```
garw: & missing?

src/mococrw/mac.h
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * This function compares the given value (hmacValue), with the value calculated during finish().
+     * Invoke finish() before calling verify().
+     * Only the characters of the shorter value (hmacValue or calculated value) are checked for equality with the
+     * longer one if the length differs.
+     * @throws MoCOCrWException if verify is invoked before finish was invoked.
+     * @throws MoCOCrWException if the verification fails.
+     * @param hmacValue The value which shall be compared to the calculated value
+     */
+    virtual void verify(const std::vector<uint8_t> hmacValue) = 0;
```
garw: & missing?

src/mococrw/mac.h
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * @see MessageAuthenticationCode::verify
+     */
+    void verify(const std::vector<uint8_t> hmacValue) override;
+
+    /**
+     * @brief The move constructor
+     * @param other The other HMAC to be moved
+     */
+    HMAC(HMAC &&other);
```
garw, line 117: Should we explicitly delete copying? I don't think that copying such a context with state makes a lot of sense. We should better forbid it in the first place.

src/mac.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+        _isFinished = true;
+
+        return _result;
+    }
+
+    void verify(const std::vector<uint8_t> hmacValue)
+    {
+        if (!_isFinished) {
+            throw MoCOCrWException("finish() has to be invoked before verify()");
```
garw: As mentioned in the header. I would rather just finish it here now. This would feel more natural for the verification flow.

src/mac.cpp
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        if (hmacValue.size() < _result.size()) {
+            if (!std::equal(hmacValue.begin(), hmacValue.end(), _result.begin())) {
+                throw MoCOCrWException("HMAC verification failed. Calculated value: . Received value: ");
+            }
+        } else {
+            if (!std::equal(_result.begin(), _result.end(), hmacValue.begin())) {
```
garw: This is definetely no constant time compare.

src/mac.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    void verify(const std::vector<uint8_t> hmacValue)
+    {
+        if (!_isFinished) {
+            throw MoCOCrWException("finish() has to be invoked before verify()");
+        }
+
+        if (hmacValue.size() < _result.size()) {
+            if (!std::equal(hmacValue.begin(), hmacValue.end(), _result.begin())) {
+                throw MoCOCrWException("HMAC verification failed. Calculated value: . Received value: ");
```
garw: If the sizes missmatch, this is a format problem? Why do you check of <. Wouldn't it make more sense to check for !=?

src/openssl_wrap.cpp
```diff
@@ -1243,5 +1243,39 @@ EC_KEY *_EVP_PKEY_get0_EC_KEY(EVP_PKEY *pkey)
...
+}
+
+void _HMAC_Update(HMAC_CTX *ctx, const unsigned char *data, int len)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Update, ctx, data, len);
+}
+
+SSL_HMAC_CTX_Ptr _HMAC_CTX_new()
+{
+    return SSL_HMAC_CTX_Ptr{OpensslCallPtr::callChecked(lib::OpenSSLLib::SSL_HMAC_CTX_new)};
```
garw: There is some infrastructure with the createOpenSSLObject and createManagedOpenSSLObject function(templates). It unfortunately doesn't work for everything but should be imho the preferred pattern in this library.

src/mac.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+        if (key.empty()) {
+            throw MoCOCrWException("Key for HMAC is empty.");
+        }
+
+        _ctx = openssl::_HMAC_CTX_new();
+        openssl::_HMAC_Init_ex(_ctx.get(), key.data(), key.size(), digestFn, NULL);
+    }
+
+    ~Impl() = default;
```
garw, line 46: Secure clearing of key memory?

src/mac.cpp
```diff
@@ -0,0 +1,137 @@
...
+        openssl::_HMAC_Update(_ctx.get(), message.data(), message.size());
+    }
+
+    std::vector<uint8_t> finish()
+    {
+        if (_isFinished) {
+            throw MoCOCrWException("finish() can't be called twice.");
+        }
+
+        /* Currently SHA512 has the greatest hash size */
```
garw: Why the comment? Imho the comment should (if at all) refer to EVP_MAX_MD_SIZE being a constant maintained by OpenSSL and thus there is no maintainence effort on our end?

src/mac.cpp
```diff
@@ -0,0 +1,137 @@
...
+        _result.resize(EVP_MAX_MD_SIZE);
+
+        unsigned int tagLength;
+        openssl::_HMAC_Final(_ctx.get(), _result.data(), &tagLength);
+
+        if (_outputLength == 0) {
```
garw: This seems a weird default? It is not documented anywhere. Do we really want to support this? Are there real use-cases for truncating the HMAC?

garw (dismissed):

src/mococrw/mac.h
wusto: done

src/mococrw/mac.h
wusto: Done

src/mococrw/mac.h
wusto: jep. Done

src/mococrw/mac.h
wusto: Done

src/mac.cpp
wusto: OpenSSL does this checking in its example.
https://wiki.openssl.org/index.php/EVP_Signing_and_Verifying#Calculating_HMAC

What's your preference?
It make also sense to abort the verification if the string size doesn't match.

src/openssl_wrap.cpp
wusto: Done

src/mac.cpp
wusto: This should be done by the std::unique_ptr<HMAC_CTX>:
using SSL_HMAC_CTX_Ptr = std::unique_ptr<HMAC_CTX, SSLDeleter<HMAC_CTX, lib::OpenSSLLib::SSL_HMAC_CTX_free>>;

This is invoking OPENSSL_cleanse(ctx->key, sizeof(ctx->key));
As we don't store the key anywhere in between that should be save.

garw: ah nice

src/mac.cpp
wusto: Done

src/mac.cpp
wusto: Truncating the HMAC is the default API for the test-vectors and from openssl. I would prefer to not change it.

wusto: Removed.

src/mac.cpp
wusto: Adapated.

src/mococrw/mac.h
wusto: done

src/openssl_wrap.cpp
```diff
@@ -1262,5 +1268,26 @@ void _ECDH_KDF_X9_63(std::vector<uint8_t> &out, const std::vector<uint8_t> &Z, c
...
 
+void _HMAC_Init_ex(HMAC_CTX *ctx, const std::vector<uint8_t> &key, const EVP_MD *md, ENGINE *impl)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Init_ex, ctx, reinterpret_cast<const void *>(key.data()),
+                                  key.size(), md, impl);
+}
+
+void _HMAC_Final(HMAC_CTX* ctx, std::vector<uint8_t> &md, unsigned int* len)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Final, ctx, md.data(), len);
```
neverpanic: I'm not sure this is safe: Users could pass `md` with a size smaller than `EVP_MAX_MD_SIZE` here. We should make this interface impossible to misuse by allocating the vector internally and returning it by value.

I guess a good interface might be this:
```c++
std::vector<uint8_t> _HMAC_Final(HMAC_CTX* ctx) {
    unsigned int length = EVP_MAX_MD_SIZE;
    std::vector<uint8_t> md(length);
    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_HMAC_Final, ctx, md.data(), &length);
    md.resize(length);
    return md;
}
```

src/mococrw/mac.h
```diff
@@ -0,0 +1,148 @@
...
+ * @brief The MessageAuthenticationCode class is the abstract base class for the different implementation of
+ * message authentication codes.
+ */
+class MessageAuthenticationCode {
+public:
+
+    /**
+     * @brief ~MessageAuthenticationCode
+     */
+    virtual ~MessageAuthenticationCode();
```
neverpanic, line 37:
```suggestion
    virtual ~MessageAuthenticationCode();

```

src/mococrw/mac.h
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @param message chunk of data used for MAC.
+     * @throws MoCOCrWException if this function is invoked after finish was called.
+     */
+    virtual void update(const std::vector<uint8_t>& message) = 0;
+
+    /**
+     * @brief Finalize the MAC.
+     *
+     * This function calculates the messaged authentication code.
```
```suggestion
     * This function calculates the message authentication code.
```

src/mococrw/mac.h
```diff
@@ -0,0 +1,148 @@
...
+     */
+    virtual std::vector<uint8_t> finish() = 0;
+
+    /**
+     * @brief Verifies the MAC
+     *
+     * This function compares the given value (macValue), with the value calculated during finish().
+     * Invoke finish() before calling verify().
+     * Only the characters of the shorter value (macValue or calculated value) are checked for equality with the
+     * longer one if the length differs.
```
neverpanic: I'm not sure we want that behavior. It would allow attackers to pass a 0-length MAC to disable MAC verification.

src/mococrw/mac.h
```diff
@@ -0,0 +1,148 @@
...
+    virtual void verify(const std::vector<uint8_t> &macValue) = 0;
+};
+
+class HMAC : public MessageAuthenticationCode {
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
+     * @param key The key used for HMAC
+     *
```
```suggestion
```

src/mococrw/mac.h
```diff
@@ -0,0 +1,148 @@
...
+     * This function calculates the messaged authentication code.
+     *
+     * @throws MoCOCrWException if this function is invoked twice.
+     * @return The calculated message authentication code.
+     */
+    virtual std::vector<uint8_t> finish() = 0;
+
+    /**
+     * @brief Verifies the MAC
```
neverpanic, line 59: This should mention that the comparison happens in constant time.

src/mac.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    void verify(const std::vector<uint8_t> &hmacValue)
+    {
+        if (!_isFinished) {
+            finish();
+        }
+
+        const size_t m = (hmacValue.size() < _result.size() ? hmacValue.size() : _result.size());
+        if (!!CRYPTO_memcmp(hmacValue.data(), _result.data(), m)) {
```
neverpanic: Why the double-!! here? Wouldn't this work just fine without it? If it doesn't, I'd prefer adding `!= 0`, since that's clearer to read.

neverpanic (changes requested):

src/mococrw/mac.h
wusto: Done

src/mococrw/mac.h
wusto: garw also asked this.
I have taken this from the openssl implementation (https://wiki.openssl.org/index.php/EVP_Signing_and_Verifying#Calculating_HMAC). If you want I can check for equality of both string before comparing them.
I have not strong opinion here.

src/mococrw/mac.h
wusto: Done

src/mococrw/mac.h
wusto: Done

src/openssl_wrap.cpp
wusto: That's a good idea.
Somehow I sticked to the openssl approach. But this one is definitely the better one.
Thx.

src/mococrw/mac.h
wusto: Done

src/mac.cpp
wusto: Also taken from here:
https://wiki.openssl.org/index.php/EVP_Signing_and_Verifying#Calculating_HMAC
As I don't know if this triggers a compiler non-optimisation I have copied it. 

What do  you think?
Shall I remove it?

src/mococrw/mac.h
neverpanic: The OpenSSL example also checks for a zero-length string in the macValue.

I'd really prefer if we didn't attempt any magic here. If the strings aren't the same length, make this return false in any case. You can still trigger the `CRYPTO_memcmp` to keep the constant time behavior, i.e. maybe use the same code, but add the equivalent of
```
result &= size1 == size2;
```

src/mac.cpp
neverpanic: I think the double-`!!` is in the OpenSSL code to ensure that it either returns 0 or 1, regardless of whether `CRYPTO_memcmp` returns any value other than those two. Since we're not storing the result of the `CRYPTO_memcmp` in an `int`, we won't need it. Please remove it.

neverpanic (changes requested):
See my answers for the two remaining comments, other than that this lg2m.

src/mococrw/mac.h
wusto: I will return false (throw an exception), if the length differs.

wusto: Done

src/mac.cpp
wusto: Done

neverpanic (approved):

src/mac.cpp
wusto: Done

src/mac.cpp
wusto: Done

PR #62, opened 2020-04-07
Add workaround for OpenSSL PEM parsing bug

blenk92:
OpenSSL currently cannot parse PEM certificates that contain the
base64 encoded content in a single line and the length of the line
is a multiple of 254. Due to this, a workaround that splits the base64
encoded string into lines of 64 chars has been introduced.

OpenSSL issue: https://github.com/openssl/openssl/issues/9187

### DIFF
--- a/src/x509.cpp
+++ b/src/x509.cpp
@@ -16,6 +16,11 @@
  * limitations under the License.
  * #L%
  */
+
+#include <fstream>
+#include <boost/range/algorithm_ext/erase.hpp>
+#include <boost/format.hpp>
+
 #include "mococrw/x509.h"
 #include "mococrw/bio.h"
 #include "mococrw/error.h"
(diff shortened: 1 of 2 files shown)


src/format_utils.h
```diff
@@ -22,13 +22,85 @@
...
+ * @param pem String containing the certificate to be reformatted
+ * @throws MoCOCrWException if pem doesn't contain the proper begin and end markers
+ * @returns Certificate with reformatted base64
+ */
+inline std::string reformatPEMCertificate(const std::string& pem) {
+    // search for begin-marker and extract everything from beginning of the string
+    // until the begin-markers end
+    std::string prevBase64;
+    size_t endBeginMarker = findNextMarker(pem, "-----BEGIN CERTIFICATE-----", true);
+    std::copy(pem.begin(), pem.begin() + endBeginMarker, std::back_inserter(prevBase64));
```
garw: substr probably easier here as well?

src/format_utils.h
```diff
@@ -22,13 +22,85 @@
...
+ * Helper function for reformatPEMCertificate:
+ * Searches for the next occurence of the given marker in pem starting at start
+ * @param pem String to search in
+ * @param marker Marker to search
+ * @param endPos Inidicates if start or end position of the marker shall be returned
+ * @param start Position to start searching at
+ * @throws MoCOCrWException if marker couldn't be found
+ * @return Start/End of next occurence of marker
+ */
+inline size_t findNextMarker(const std::string& pem, const std::string& marker, bool endPos, size_t start = 0)
```
garw: if this wasn't a temporary utility function I would complain about the boolean parameter. This is hard to read in function calls. If you read the calls in 2 weeks, you probably know why ;)

src/x509.cpp
```diff
@@ -31,17 +34,26 @@ using namespace openssl;
...
 {
-    FileBio bio{filename, FileBio::FileMode::READ, FileBio::FileType::TEXT};
-    auto cert = _PEM_read_bio_X509(bio.internal());
-    return X509Certificate{std::move(cert)};
+    // Change to be reverted to previous solution after OpenSSL parsing
+    // bug for PEM certificates has been fixed.
+    std::ifstream pemFile(filename);
+    std::stringstream buffer;
+    if (pemFile) {
+        buffer << pemFile.rdbuf();
```
garw, line 145: I'm not sure if we should add error handling here.
pemFile should be at eof after this but the bad bit should not be set?

garw (dismissed):
Mostly knitpicking. You might want to look at the error handling for fromPEMFile

src/format_utils.h
blenk92: agree

src/format_utils.h
blenk92: i guess you are right, should be better now

src/x509.cpp
blenk92: right.. but for some reason error handling is a bit weird http://www.cplusplus.com/reference/ostream/ostream/operator%3C%3C/:

(2) stream buffers
    Retrieves as many characters as possible from the input sequence controlled by the stream buffer object pointed by sb (if any) and inserts them into the stream, until either the input sequence is exhausted or the function fails to insert into the stream.
    Internally, the function accesses the output sequence by first constructing a sentry object. Then (if good), it inserts characters into its associated stream buffer object as if calling its member function sputc, and finally destroys the sentry object before returning.

failbit | The function failed to format val (it may also be set if the construction of sentry failed). For (2), it is set if no characters could be extracted from the object pointed by sb.
-- | --
badbit | Either  the insertion on the stream failed, or some other error happened (such  as when this function catches an exception thrown by an internal  operation). When set, the integrity of the stream may have been affected. For (2), it is also set if sb is a null pointer.

So i guess checking for fail() on the buffer should be sufficient, or am I missing something

src/format_utils.h
```diff
@@ -22,13 +22,88 @@
...
+ */
+inline std::string reformatPEMCertificate(const std::string& pem) {
+    // search for begin-marker and extract everything from beginning of the string
+    // until the begin-markers end
+    std::string prevBase64;
+    size_t endBeginMarker = findEndOfNextMarker(pem, "-----BEGIN CERTIFICATE-----");
+    prevBase64 = pem.substr(0, endBeginMarker);
+
+    // search for end-marker and extract everything from the beginning of the marker
+    // until end of the the string
```
wusto: one "the" should be enough

src/format_utils.h
```diff
@@ -22,13 +22,88 @@
...
+    postBase64 = pem.substr(startEndMarker);
+
+    // extract base64 encoded content
+    std::string base64 = pem.substr(endBeginMarker, startEndMarker - endBeginMarker - 1);
+    boost::range::remove_erase_if(base64, [](char x) { return std::isspace(x); });
+
+    // Insert newline after each 64 chars into base64 encoded content
+    std::string splittedBase64;
+    auto iter = base64.begin();
+    for (; std::distance(iter, base64.end()) >=64; iter+=64) {
```
wusto: I don't know the actual coding style, but I guess we are adding spaces between operators and integers. If not just ignore this comment.

wusto (changes requested):

src/x509.cpp
garw: unclear. It reads to me like both indicate different error scenarios but in principle I agree that error handling with the stream APIs is weird. Maybe we should better be safe than sorry and check both?

blenk92: `fail()` actually checks both bits:
>  Returns true if either (or both) the failbit or the badbit error state flags is set for the stream.
http://www.cplusplus.com/reference/ios/ios/fail/

blenk92: or did you mean check on both, buffer and pemFile?

wusto (approved):

src/x509.cpp
```diff
@@ -31,17 +34,29 @@ using namespace openssl;
...
     return X509Certificate{std::move(cert)};
 }
 
 X509Certificate X509Certificate::fromPEMFile(const std::string &filename)
 {
-    FileBio bio{filename, FileBio::FileMode::READ, FileBio::FileType::TEXT};
-    auto cert = _PEM_read_bio_X509(bio.internal());
-    return X509Certificate{std::move(cert)};
+    // Change to be reverted to previous solution after OpenSSL parsing
+    // bug for PEM certificates has been fixed.
```
neverpanic, line 140: Can you link the OpenSSL bug here?

src/x509.cpp
```diff
@@ -31,17 +34,29 @@ using namespace openssl;
...
+    std::ifstream pemFile(filename);
+    std::stringstream buffer;
+    if (pemFile.is_open()) {
+        buffer << pemFile.rdbuf();
+        pemFile.close();
+    } else {
+        throw MoCOCrWException("Couldn't open certificate at: " + filename);
```
neverpanic: Are the MoCOCrWExceptions here an ABI change, i.e. did this throw MoCOCrWExceptions before?

src/format_utils.h
```diff
@@ -22,13 +22,98 @@
...
 #include <boost/format.hpp>
+#include <boost/range/algorithm_ext/erase.hpp>
 
 #include "mococrw/error.h"
 
 namespace mococrw {
 
 namespace util {
 
+
```
neverpanic: Not sure I like that this is in a header file. Can you move this to the cpp into an anonymous namespace?

src/format_utils.h
```diff
@@ -22,13 +22,98 @@
...
+ */
+inline size_t findEndOfNextMarker(const std::string& pem, const std::string& marker, size_t start = 0)
+{
+    return findStartOfNextMarker(pem, marker, start) + marker.size();
+}
+
+/**
+ * Reformats a PEM encoded certificate to contain only base64 encoded lines that consist
+ * of 64 char at max. This had to be introduced to workaround a bug in OpenSSL that prevents
+ * from parsing PEM certificates that contain the base64 encoded parts in only one line and
```
neverpanic:
```suggestion
 * parsing of PEM certificates that contain the base64 encoded parts in only one line and
```

src/x509.cpp
blenk92: hmm, right.. i guess not. It guess std::runtime_error will also do (that actually could have been thrown earlier)

neverpanic: @garw What's your opinion on that? Not sure std::runtime_error is a good solution here either. Maybe we should just invoke the OpenSSL functions with empty data and let that throw an error?

neverpanic (approved):

neverpanic (approved):

PR #66, opened 2020-04-17
Various small fixes for issues reported by clang-tidy

neverpanic:
- Avoid calling `getNid()` on `std::move()`'d objects
- Initialize KeyUsageExtension member variables
- bio: throw exceptions, not pointers to exceptions


garw (approved):
lg2m

PR #67, opened 2020-04-17
ECDSA: Support verification from IEEE1363 format

garw:
ECDSA signatures are a tuple of two (larged) unsigned integers
(r,s). Most libraries and protocols encode ECDSA signature in
DER serialized ASN.1 of SEQUENCE (int r, int s). This is also
what the current verifyDigest() and verifyMessage() functions
in the ECDSASignaturePublicKeyCtx expect.

In addition to this format, there is also a serialization format
specified in IEEE 1363. This serializes (r,s) in the form
padToKeyLength(r) || padToKeyLength(s). As this is binary
incompatible, it can't be processed with the current verification
functions.

This patch adds the verifyDigestIEEE1363() and verifyMessageIEEE1363()
functions that take an ECDSA signature in IEEE 1363 serialized form,
then convert it into the format that OpenSSL expects and then uses the
already provided functionality to validate the signature.

### DIFF
--- a/src/asymmetric_crypto_ctx.cpp
+++ b/src/asymmetric_crypto_ctx.cpp
@@ -1,7 +1,7 @@
 /*
  * #%L
  * %%
- * Copyright (C) 2018 BMW Car IT GmbH
+ * Copyright (C) 2018-2020 BMW Car IT GmbH
  * %%
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -512,10 +512,36 @@ namespace mococrw {
             throw MoCOCrWException(e.what());
         }
     }
-    
+
+    void ECDSASignaturePublicKeyCtx::verifyDigestIEEE1363(const std::vector<uint8_t> &signature,
+                                                          const std::vector<uint8_t> &messageDigest) {
+        /* We reformat the signature to the ASN.1 format understood by OpenSSL and then pass it on
+         * to the standard verification function.
+         */
+        size_t keySizeBytes = (_impl->key.getKeySize() + 7) / 8;
+        if (signature.size() != 2*keySizeBytes) {
+            throw MoCOCrWException("Invalid signature size.");
+        }
+        auto r = _BN_bin2bn(signature.data(), keySizeBytes);
+        auto s = _BN_bin2bn(signature.data() + keySizeBytes, keySizeBytes);
+        if (r == nullptr || s == nullptr) {
+            throw MoCOCrWException("Cannot extract ECSDA signature components");
+        }
+        auto ecdsa = createManagedOpenSSLObject<SSL_ECDSA_SIG_Ptr>();
+        _ECDSA_SIG_set0(ecdsa.get(), std::move(r), std::move(s));
+        auto asn1Signature = _i2d_ECSDA_SIG(ecdsa.get());
+        return verifyDigest(asn1Signature, messageDigest);
+    }
+
+
     void ECDSASignaturePublicKeyCtx::verifyMessage(const std::vector<uint8_t> &signature,
                                                    const std::vector<uint8_t> &message) {
         verifyDigest(signature, createHash(_impl->hashFunction, message));
+    }
+
+    void ECDSASignaturePublicKeyCtx::verifyMessageIEEE1363(const std::vector<uint8_t> &signature,
+                                                           const std::vector<uint8_t> &message) {
+        verifyDigestIEEE1363(signature, createHash(_impl->hashFunction, message));
     }
 
     /* ###########
--- a/src/mococrw/asymmetric_crypto_ctx.h
+++ b/src/mococrw/asymmetric_crypto_ctx.h
@@ -501,11 +501,40 @@ public:
      */
     ~ECDSASignaturePublicKeyCtx();
 
+    /**
+     * Verifies an ECSDA signature against the given digest hash.
+     * It expects the ECSDA signature in the common ASN.1 encoding
+     * format SEQUENCE (INT r , INT s)
+     */
     void verifyDigest(const std::vector<uint8_t> &signature,
                       const std::vector<uint8_t> &digest) override;
 
+    /**
+     * Verifies an ECSDA signature against the given message
+     * It expects the ECSDA signature in the common ASN.1 encoding
+     * format SEQUENCE (INT r , INT s)
+     */
     void verifyMessage(const std::vector<uint8_t> &signature,
                        const std::vector<uint8_t> &message) override;
+
+
+    /**
+     * Verifies an ECSDA signature against the given digest hash.
+     * It expects the ECSDA signature in the IEEE 1363 format in which
+     * the two integers r and s are padded to the length of the ECC key
+     * and then just (binary)-concatenated together
+     */
+    void verifyDigestIEEE1363(const std::vector<uint8_t> &signature,
+                              const std::vector<uint8_t> &digest);
+
+    /**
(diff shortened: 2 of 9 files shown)


src/asymmetric_crypto_ctx.cpp
```diff
@@ -1,7 +1,7 @@
 /*
  * #%L
  * %%
- * Copyright (C) 2018 BMW Car IT GmbH
+ * Copyright (C) 2018-2020 BMW Car IT GmbH
```
neverpanic, line 4: First line of the commit message, what's "larged"? Did you mean "large"?

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -501,12 +501,41 @@ class ECDSASignaturePublicKeyCtx : public DigestVerificationCtx, public MessageV
      */
     ~ECDSASignaturePublicKeyCtx();
 
+    /**
+     * Verifies a ECSDA signature against the given digest hash.
```
neverpanic: Should this be
```suggestion
     * Verifies an ECSDA signature against the given digest hash.
```

If so, the same applies to all comments below.

src/mococrw/openssl_wrap.h
```diff
@@ -533,7 +536,7 @@ int _BIO_read(BIO* bio, std::vector<uint8_t> &buf, std::size_t numBytes);
 SSL_X509_Ptr _d2i_X509_bio(BIO* bp);
 
 /*
- * Read a DER encoded X509 certificate request from a bio
+ * Read a DER encoded X509 certificate request from o
```
neverpanic: This change seems unintentional?

src/openssl_wrap.cpp
```diff
@@ -130,6 +130,35 @@ using OpensslCallIsPositive =
 using OpensslCallPtr = ::cppc::CallCheckContext<::cppc::IsNotNullptrReturnCheckPolicy,
                                                 OpenSSLExceptionErrorPolicy>;
 
+
+template<typename T>
+class OpenSSLGuardedOutputBuffer {
+public:
+    OpenSSLGuardedOutputBuffer() : _ptr{nullptr} {}
+    OpenSSLGuardedOutputBuffer(const OpenSSLGuardedOutputBuffer<T>&) = delete;
+    OpenSSLGuardedOutputBuffer(OpenSSLGuardedOutputBuffer<T>&&) = delete;
```
neverpanic, line 139: Shouldn't you delete copy-assignment too to prevent double free of the pointer?

src/openssl_wrap.cpp
```diff
@@ -1012,6 +1047,13 @@ std::vector<uint8_t> Asn1IntegerToBinary(ASN1_INTEGER *number)
     return serialNumber;
 }
 
+SSL_BIGNUM_Ptr _BN_bin2bn(const uint8_t* data, size_t dataLen) {
+    if (dataLen > std::numeric_limits<int>::max()) {
+        throw OpenSSLException("INT_MAX is the maximal supported size for BIGNUM");
```
neverpanic:
```suggestion
        throw OpenSSLException("INT_MAX is the maximum supported size for BIGNUM");
```

src/openssl_wrap.cpp
```diff
@@ -1250,6 +1292,23 @@ EC_KEY *_EVP_PKEY_get0_EC_KEY(EVP_PKEY *pkey)
     return OpensslCallPtr::callChecked(lib::OpenSSLLib::SSL_EVP_PKEY_get0_EC_KEY,pkey);
 }
 
+void _ECDSA_SIG_set0(ECDSA_SIG* sig, SSL_BIGNUM_Ptr r, SSL_BIGNUM_Ptr s) {
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_ECDSA_SIG_set0, sig, r.release(), s.release());
+}
+
+std::vector<uint8_t> _i2d_ECSDA_SIG(const ECDSA_SIG *sig) {
+    OpenSSLGuardedOutputBuffer<unsigned char> outputBuffer;
+    int result = lib::OpenSSLLib::SSL_i2d_ECDSA_SIG(sig, &outputBuffer.get());
```
neverpanic, line 1312: `outputBuffer.get()` is a `nullptr` here, you're passing the address of that `nullptr` in the `outputBuffer` object. How is the write operation that follows not a null pointer dereference?

The documentation says
> i2d_ECDSA_SIG() creates the DER encoding of the ECDSA signature sig and writes the encoded signature to *pp (note: if pp is NULL i2d_ECDSA_SIG() returns the expected length in bytes of the DER encoded signature). i2d_ECDSA_SIG() returns the length of the DER encoded signature (or 0 on error).

The way I see this you'd have to call `SSL_i2d_ECDSA_SIG(sig, nullptr)` and use the length to allocate the output buffer, then call it again. What am I missing?

src/asymmetric_crypto_ctx.cpp
```diff
@@ -512,12 +512,38 @@ namespace mococrw {
             throw MoCOCrWException(e.what());
         }
     }
-    
+
+    void ECDSASignaturePublicKeyCtx::verifyDigestIEEE1363(const std::vector<uint8_t> &signature,
+                                                          const std::vector<uint8_t> &messageDigest) {
+        /* We reformat the signature to the OpenSSL understood ASN.1 format and then pass it on the
```
neverpanic:
```suggestion
        /* We reformat the signature to the ASN.1 format understood by OpenSSL format and then pass it on to the
```

neverpanic (changes requested):

src/openssl_wrap.cpp
garw: should be implicitly deleted by user-delcared (deleted) move constructor but I will make it explicit for better readability anyways.

src/openssl_wrap.cpp
garw: The function expects `unsigned char **pp`, which is a classical pointer output parameter in which memory is reserved for you. This is actually also what happens. I'll add a comment w.r.t. confusing OpenSSL documentation

neverpanic (approved):
LG2M

wusto (approved):
LG2M2

src/asymmetric_crypto_ctx.cpp
```diff
@@ -512,12 +512,38 @@ namespace mococrw {
...
+        return verifyDigest(asn1Signature, messageDigest);
+    }
+
+
     void ECDSASignaturePublicKeyCtx::verifyMessage(const std::vector<uint8_t> &signature,
                                                    const std::vector<uint8_t> &message) {
         verifyDigest(signature, createHash(_impl->hashFunction, message));
     }
 
+    void ECDSASignaturePublicKeyCtx::verifyMessageIEEE1363(const std::vector<uint8_t> &signature,
```
ricleite, line 542: Should `verifyMessageIEEE1363` not be part of `DigestVerificationCtx`?
I get why it is declared here, but this also means that `DigestVerificationCtx` cannot be used to check this type of signatures.

Same problem applies to `verifyMessageIEEE1363` and `MessageVerificationCtx`

ricleite (commented):
See comments.

I think maybe the constructor of `ECDSASignaturePublicKeyCtx` should take in an argument that defines message encoding, so that `verifyMessage` and `verifyDigest` could be used to accept IEEE1363 format.
That would be coherent with `RSASignaturePublicKeyCtx`, which also has a constructor with arguments that control part of the message encoding (well, padding).

Otherwise, this LGTM

PR #68, opened 2020-04-20
ECDSASignaturePublicKeyCtx: IEEE1363 support

garw:
Improve IEEE1636 support so that it is actually compatible
with the abstract verification context interfaces that are
implemented by ECDSASignaturePublicKeyCtx.

In order to do so, the constructor now offers to specify which
signature serialization format should be used by the verifyDigest()
and verifyMessage() functions (that are then also callable via
the abstract interfaces).

### DIFF
--- a/src/asymmetric_crypto_ctx.cpp
+++ b/src/asymmetric_crypto_ctx.cpp
@@ -463,20 +463,96 @@ namespace mococrw {
         return signDigest(createHash(_impl->hashFunction, message));
     }
 
+namespace {
+    std::vector<uint8_t> _IEEE1363EcSignatureToAsn1ECSignature(const std::vector<uint8_t> &signature,
+                                                               size_t keySizeBytes) {
+        if (signature.size() != 2*keySizeBytes) {
+            throw MoCOCrWException("Invalid signature size.");
+        }
+        auto r = _BN_bin2bn(signature.data(), keySizeBytes);
+        auto s = _BN_bin2bn(signature.data() + keySizeBytes, keySizeBytes);
+        if (r == nullptr || s == nullptr) {
+            throw MoCOCrWException("Cannot extract ECSDA signature components");
+        }
+        auto ecdsa = createManagedOpenSSLObject<SSL_ECDSA_SIG_Ptr>();
+        _ECDSA_SIG_set0(ecdsa.get(), std::move(r), std::move(s));
+        return _i2d_ECSDA_SIG(ecdsa.get());
+    }
+}
+
     /*
      * PIMPL-Class for ECDSASignaturePublicKeyCtx
      */
     class ECDSASignaturePublicKeyCtx::Impl : public ECDSAImpl<AsymmetricPublicKey> {
-        using ECDSAImpl<AsymmetricPublicKey>::ECDSAImpl;
+    public:
+        Impl(const AsymmetricPublicKey& key, openssl::DigestTypes hashFunction, ECSDASignatureFormat format)
+            : ECDSAImpl{key, hashFunction},
+              _sigFormat{format}
+            {}
+
+        void verifyDigest(const std::vector<uint8_t> &signature,
+                          const std::vector<uint8_t> &messageDigest) {
+            if (_sigFormat == ECSDASignatureFormat::IEEE1363) {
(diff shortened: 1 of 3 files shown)


src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -477,6 +483,17 @@ class ECDSASignaturePublicKeyCtx : public DigestVerificationCtx, public MessageV
     ECDSASignaturePublicKeyCtx(const AsymmetricPublicKey& key,
                                openssl::DigestTypes hashFunction = openssl::DigestTypes::SHA256);
 
+    /**
+     * @brief Constructor
+     *
+     * @param key The public key to be used
+     * @param sigFormat The format that in which signatures are provided
```
ricleite: Doxygen params are in the wrong order, same with the `X509Certificate` version below

src/asymmetric_crypto_ctx.cpp
```diff
@@ -463,16 +463,87 @@ namespace mococrw {
...
+                                                           openssl::DigestTypes hashFunction,
+                                                           ECSDASignatureFormat format)
+        : _impl(std::make_unique<ECDSASignaturePublicKeyCtx::Impl>(key, hashFunction, format)) {}
+
     ECDSASignaturePublicKeyCtx::ECDSASignaturePublicKeyCtx(const AsymmetricPublicKey& key,
                                                            openssl::DigestTypes hashFunction)
-        : _impl(std::make_unique<ECDSASignaturePublicKeyCtx::Impl>(key, hashFunction)) {}
+        : ECDSASignaturePublicKeyCtx(key, hashFunction, ECSDASignatureFormat::ASN1_SEQUENCE_OF_INTS) {}
 
     ECDSASignaturePublicKeyCtx::ECDSASignaturePublicKeyCtx(const X509Certificate& cert,
```
ricleite, line 548: One of the constructor definitions is missing here:

```
    ECDSASignaturePublicKeyCtx(const X509Certificate& cert, openssl::DigestTypes hashFunction,
                               ECSDASignatureFormat sigFormat);
```

Edit: This can cause a linking problem if a user tries to use this declaration

garw: good catch.

src/asymmetric_crypto_ctx.cpp
```diff
@@ -463,21 +463,97 @@ namespace mococrw {
         return signDigest(createHash(_impl->hashFunction, message));
     }
 
+namespace {
```
neverpanic, line 466: Commit message: `s/so that t now/so that it now/`, `s/sanetization/sanitization/`.

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -458,6 +458,12 @@ class ECDSASignaturePrivateKeyCtx : public DigestSignatureCtx, public MessageSig
     std::unique_ptr<Impl> _impl;
 };
 
+/** @brief Serialization formats for ECDSA signatures
+ */
+enum class ECSDASignatureFormat {
```
neverpanic, line 463: Should we add more documentation than just the names of the enum values? A pointer to the relevant standards, maybe?

tests/unit/test_signature.cpp
```diff
@@ -771,35 +771,39 @@ TEST_F(SignatureTest, testUnsuccessfulEccVerificationWithWrongPublicKey)
  */
 TEST_F(SignatureTest, testSuccessfulEccVerificationInIEEE1363Format)
```
neverpanic, line 772: Should we also add a test that verifies that the default serialization format is the ASN.1 sequence of integers by creating a signature with this encoding and attempting to verify it with the default encoding?

neverpanic (commented):
A couple of minor points.

tests/unit/test_signature.cpp
garw: This is what
`TEST_P(SignatureTest, testSuccessfulSignatureAndVerification)`
implicitly already does. We also check that the default verification format is the ASN.1 one by the tests that check against the known good signatures (which are ASN.1 encoded).

neverpanic (approved):

ricleite (approved):
LGTM! Thanks

PR #69, opened 2020-05-04
Allow setting additional data for AEAD

miroshko:
Extend AuthenticatedAESCipher so it is possible to associate
additional data with the message.

### DIFF
--- a/doc/examples/example_symm.md
+++ b/doc/examples/example_symm.md
@@ -57,12 +57,19 @@ assert(std::equal(plaintext.begin(), plaintext.end(), decryptedtext.begin()));
 
 Authenticated ciphers not only encrypt data but also compute authentication tag over it. It allows to detect if data was modified after it was encrypted.
 
+It is also possible to associate additional unencrypted data with the message. The authentication tag will ensure the integrity of both encrypted and unencrypted data in that case. This is, e.g., used to protect the integrity of header fields that contain metadata that need to be visible on transport.
+
 Example below is similar to the previous one but uses authenticated cipher mode GCM. After encryption we take computed authentication tag and carry it along with the message for further authentication.
 
 ```c++
 
 std::string message{"This is a message we going to encrypt and then, hopefully, decrypt."};
 std::vector<uint8_t> plaintext{message.begin(), message.end()};
+
+// Optional: Associated data
+std::string data{"This is a message we are not going to encrypt but want to make sure arrived unchanged."};
+std::vector<uint8_t> associatedData{data.begin(), data.end()};
+
 
 auto secretKey = utility::cryptoRandomBytes(256/8);
 
@@ -78,6 +85,9 @@ auto secretKey = utility::cryptoRandomBytes(256/8);
 // function will throw.
 auto encryptor = AESCipherBuilder{SymmetricCipherMode::GCM, SymmetricCipherKeySize::S_256, secretKey}.buildAuthenticatedEncryptor();
 
+// Optional step: Associating additional unencrypted data with the message
+encryptor->addAssociatedData(associatedData);
+
 encryptor->update(plaintext);
 auto ciphertext = encryptor->finish();
 
@@ -90,6 +100,10 @@ auto authTag = encryptor->getAuthTag();
 //
 
 auto decryptor = AESCipherBuilder{SymmetricCipherMode::GCM, SymmetricCipherKeySize::S_256, secretKey}.setIV(iv).buildAuthenticatedDecryptor();
+
+// Optional step: If associated data was set during encryption, add it here for the integrity check
+decryptor->addAssociatedData(associatedData);
+
 decryptor->update(ciphertext);
 
 // Set authentication tag **before** calling `finish()`.
--- a/src/mococrw/symmetric_crypto.h
+++ b/src/mococrw/symmetric_crypto.h
@@ -202,6 +202,18 @@ public:
      * @return the auth tag for this encryption operation
      */
     virtual std::vector<uint8_t> getAuthTag() const = 0;
+
+    /**
+     * Add associated data.
+     *
+     * This function adds the associated data in AEAD modes for authenticating while encrypting or verifying while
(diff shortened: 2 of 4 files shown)


doc/examples/example_symm.md
```diff
@@ -57,13 +57,20 @@ assert(std::equal(plaintext.begin(), plaintext.end(), decryptedtext.begin()));
 
 Authenticated ciphers not only encrypt data but also compute authentication tag over it. It allows to detect if data was modified after it was encrypted.
 
+It is also possible to associate additional unencrypted data with the message. The authentication tag will ensure the integrity of both encrypted and unencrypted data in that case.
```
neverpanic: Please remove the issue reference from the commit. Since this is open source code, these references are useless to everybody not in our network.

tests/unit/test_symmetric_crypto.cpp
```diff
@@ -532,3 +536,62 @@ TEST_F(SymmetricAuthenticatedCipherTest, throwsWhenWrongTagIsUsed) {
...
+    encryptor->setAssociatedData(_associatedData);
+    encryptor->update(_plaintext);
+    auto ciphertext = encryptor->finish();
+    auto tag = encryptor->getAuthTag();
+
+    auto decryptor = AESCipherBuilder{SymmetricCipherMode::GCM,
+                                      SymmetricCipherKeySize::S_256,
+                                      _secretKey}.setIV(encryptor->getIV()).buildAuthenticatedDecryptor();
+    decryptor->setAuthTag(tag);
+    _associatedData[2] ^= _associatedData[2];
```
neverpanic: XOR with itself is the same as writing `_associatedData[2] = 0x0`, isn't it? Did you mean to XOR with a static value?

neverpanic (changes requested):
Two minor points.

doc/examples/example_symm.md
miroshko: Done

tests/unit/test_symmetric_crypto.cpp
miroshko: My bad. But that means that the other place in the code I mindlessly copied from also needs to be fixed. Will do it as a separate commit.

src/mococrw/symmetric_crypto.h
```diff
@@ -193,6 +193,17 @@ class AuthenticatedEncryptionI
      */
     virtual void setAuthTag(const std::vector<uint8_t> &tag) = 0;
 
+    /**
+     * Set associated data.
```
garw: The remaining parts of the crypto interface are a streaming API. To the best of my understanding this here is as well (as it might be called multiple times and will just add more data. So I think, we should rename it to addAssociatedData.

tests/unit/test_symmetric_crypto.cpp
```diff
@@ -532,3 +536,62 @@ TEST_F(SymmetricAuthenticatedCipherTest, throwsWhenWrongTagIsUsed) {
...
+    ASSERT_THROW(decryptor->finish(), MoCOCrWException);
+
+    auto decryptor2 = AESCipherBuilder{SymmetricCipherMode::GCM,
+                                      SymmetricCipherKeySize::S_256,
+                                      _secretKey}.setIV(encryptor->getIV()).buildAuthenticatedDecryptor();
+    decryptor2->setAuthTag(tag);
+    decryptor2->update(ciphertext);
+
+    ASSERT_THROW(decryptor2->finish(), MoCOCrWException);
+}
```
garw, line 597: Due to the slightly weird mechanism of OpenSSL to use the UpdateCipher method that is also used for encryption, I would appreciate a test that checks if the resulting ciphertext (without the auth tag) matches if encrypted with and without additional data. This way we will immediately see what is going on if something is subtly broken around this UpdateCipher semantics. Just encrypt the data twice, once with and once without additional data and check if it matches. Please note that you need to explicitly set the IV for this also on the Encryptor.

src/mococrw/symmetric_crypto.h
```diff
@@ -193,6 +193,17 @@ class AuthenticatedEncryptionI
...
+     *
+     * This function sets the associated data in AEAD modes for authenticating while encrypting or verifying while
+     * decrypting. setAssociatedData() must be called before finalizing decryption using SymmetricCipherI::finish()
+     * and before putting encrypted data to the message using SymmetricCipherI::update()
+     *
+     * @param associatedData chunk of data to associate/verify.
+     */
+    virtual void setAssociatedData(const std::vector<uint8_t> &associatedData) = 0;
```
garw: If we want ABI stability, we probably need to add new functions to the end?

doc/examples/example_symm.md
garw:
```suggestion
It is also possible to associate additional unencrypted data with the message. The authentication tag will ensure the integrity of both encrypted and unencrypted data in that case. This is, e.g., used to protect the integrity of header fields that contain metadata that needs to be visible on transport.
```

garw (changes requested):
a few very small comments. Looks otherwise good to me.

src/mococrw/symmetric_crypto.h
miroshko: True. I renamed it.

tests/unit/test_symmetric_crypto.cpp
miroshko: Added a test

src/mococrw/symmetric_crypto.h
miroshko: I don't know what the prerequisites for ABI stability with our configuration are so I'm just trusting you and moving it to the end of the class declaration.

doc/examples/example_symm.md
miroshko: Done

neverpanic (approved):

wusto (approved):

garw (approved):

PR #70, opened 2020-05-12
ecies: Add method to calculate encoded key length

miroshko:


src/mococrw/ecies.h
```diff
@@ -142,6 +142,16 @@ class ECIESEncryptionCtx {
     std::unique_ptr<Impl> _impl;
 };
 
+/**
+ * @brief Get encoded key length in bytes
+ *
+ * Depending on the key length and the elliptic curve point conversion form the key will take different amount
+ * of bytes when encoded. This function can be used to conveniently deduct the encoded length while deserializing.
+ * @param privKey Private key
```
neverpanic: I guess we should document that this private key is really just used to figure out what key size is used in this instance of ECIES. Can you add a comment explaining that?

While you're at it, the rest of the code in this file does not add a space between function name and parameter list, so I'd prefer if we used the same style in this change.

src/mococrw/ecies.h
```diff
@@ -142,6 +142,17 @@ class ECIESEncryptionCtx {
     std::unique_ptr<Impl> _impl;
 };
 
+/**
+ * @brief Get encoded key length in bytes
+ *
+ * Depending on the key length and the elliptic curve point conversion form the key will take different amount
```
wusto:
```suggestion
 * Depending on the public key length and the elliptic curve point conversion form the public key will take different amount
```

src/mococrw/ecies.h
```diff
@@ -142,6 +142,17 @@ class ECIESEncryptionCtx {
     std::unique_ptr<Impl> _impl;
 };
 
+/**
+ * @brief Get encoded key length in bytes
```
wusto:
```suggestion
 * @brief Get encoded public key length in bytes
```

src/mococrw/ecies.h
miroshko: Done

src/mococrw/ecies.h
```diff
@@ -142,6 +142,17 @@ class ECIESEncryptionCtx {
...
+ *
+ * Depending on the key length and the elliptic curve point conversion form the key will take different amount
+ * of bytes when encoded. This function can be used to conveniently deduct the encoded length while deserializing.
+ * @param privKey Private key. This is only used to get the key size of the ECIES instance
```
wusto:
```suggestion
 * @param privKey Private key. This is only used to get the public key size of the ECIES instance
```

tests/unit/test_ecies.cpp
```diff
@@ -114,6 +114,41 @@ TEST_F(ECIESTests, testWithSalts)
...
+
+TEST_F(ECIESTests, testEncodedKeyLengthNotFactor8)
+{
+    AsymmetricKeypair secp521Key = AsymmetricKeypair::generate(ECCSpec(openssl::ellipticCurveNid::SECP_521r1));
+    AsymmetricPublicKey secp521PublicKey = secp521Key;
+    auto encCtx = encBuilder.buildEncryptionCtx(secp521PublicKey);
+    auto ephKey = encCtx->getEphemeralKey();
+    testGetEcPublicKeyOctetLength(ephKey, secp521Key);
+}
+
```
wusto: Can you please create one "static" test using the data in preparetestDataForECIESTests?

wusto (changes requested):
some minor stuff

neverpanic (approved):
LG2M for my comments.

src/ecies.cpp
```diff
@@ -299,6 +299,12 @@ class ECIESEncryptionCtx::Impl : public ECIESImpl
 
 };
 
+size_t getEcPublicKeyOctetLength(const AsymmetricPrivateKey &privKey, openssl::EllipticCurvePointConversionForm form) {
+    size_t len = (privKey.getKeySize() + 7) / CHAR_BIT;
+    size_t enc_len = form == openssl::EllipticCurvePointConversionForm::compressed ? 1 + len : 1 + 2 * len;
```
garw: Magic computation could use a comment on how these lengths come into place

src/ecies.cpp
```diff
@@ -299,6 +299,12 @@ class ECIESEncryptionCtx::Impl : public ECIESImpl
 
 };
 
+size_t getEcPublicKeyOctetLength(const AsymmetricPrivateKey &privKey, openssl::EllipticCurvePointConversionForm form) {
+    size_t len = (privKey.getKeySize() + 7) / CHAR_BIT;
```
garw: in such complex computations, a more explicit variable name benefits the readability of the overall code
```suggestion
    size_t keySizeInBytes = (privKey.getKeySize() + 7) / 8; // getKeySize() returns bits, we round up to full bytes
```
Moreover, where does CHAR_BIT come from? The name is not sufficiently verbose to explain what is going on if I didn't already understand this. So maybe just put in 8 and write a comment?

src/ecies.cpp
```diff
@@ -299,6 +299,12 @@ class ECIESEncryptionCtx::Impl : public ECIESImpl
 
 };
 
+size_t getEcPublicKeyOctetLength(const AsymmetricPrivateKey &privKey, openssl::EllipticCurvePointConversionForm form) {
```
garw: This method should check if this is actually an EC key and throw otherwise in order to prevent accidental misuse.

src/mococrw/ecies.h
garw: Why is this specific to ECIES? While I think about it, I get the impression that all of these key transformation methods rather belong into key.h? The conversion methods are actually a member of AsymmetricPublicKey, so I would propose to move this code there as well. Will be part of the PrivateKey/Keypair implicitly as well as they inherit from public key.

garw (changes requested):
Some minor issues. Major concern from my end is that this rather belongs into the key from my perspective. It is not really specific to ECIES and the other transformation methods are located there as well.

src/mococrw/ecies.h
miroshko: I moved it to AsymmetricPublicKey

src/ecies.cpp
miroshko: Added comments

tests/unit/test_ecies.cpp
miroshko: The method is now moved to the key class and the test is now ensuring that the lengths match the result of the toECPoint method for all possible cases, so another test is probably superfluous at this point

wusto (approved):

neverpanic (approved):

src/mococrw/key.h
```diff
@@ -123,6 +123,16 @@ class AsymmetricPublicKey
      */
     std::vector<uint8_t> toECPoint(openssl::EllipticCurvePointConversionForm form);
 
+    /**
+     * @brief Get length of the point of the current AsymmetricPublicKey in octet representation
+     *
+     * Depending on the public key length and the elliptic curve point conversion form, the public key will take different
+     * amount of bytes when encoded. This function can be used to conveniently deduct the encoded length while deserializing.
+     * @param form Conversion form
+     * @return Encoded public key size in bytes
```
garw, line 132: @throws if no EC key?

garw (changes requested):
Looks 99,9% good to me. Just a little update to the public documentation please.

OpenSSL naming again is kinda weird. EC_GROUP_get_degree gives you something about the underlaying field and it isn't actually always a degree but rather the bit-size of a member representation but well, ok. Good that we do this here :)
And kudos for catching this…

src/mococrw/key.h
miroshko: Done

garw (approved):

PR #71, opened 2020-05-13
ECDSA: Add support for IEEE1363 signature format for signing

blenk92:
Up until now, the IEEE1363 signature format has only been supported
for verifying ECDSA signatures. This commit adds support for creating
ECDSA signatures in the IEEE1363 format.

### DIFF
--- a/src/asymmetric_crypto_ctx.cpp
+++ b/src/asymmetric_crypto_ctx.cpp
@@ -400,7 +400,11 @@ namespace mococrw {
      * #  ECDSA  #
      * ###########
      */
-
+namespace {
+    ECDSASignatureFormat SDAtoDSASignatureFormat (ECSDASignatureFormat sigFormat) {
+        return static_cast<ECDSASignatureFormat>(static_cast<int>(sigFormat));
+    }
+}
     /*
      * Common base class for all PIMPL classes of the ECDSA contexts
      *
@@ -418,7 +422,6 @@ namespace mococrw {
 
         Key key;
     };
-
 
     /*
      * PIMPL-Class for ECDSASignaturePrivateKeyCtx
@@ -472,11 +475,11 @@ namespace {
         auto r = _BN_bin2bn(signature.data(), keySizeBytes);
         auto s = _BN_bin2bn(signature.data() + keySizeBytes, keySizeBytes);
         if (r == nullptr || s == nullptr) {
-            throw MoCOCrWException("Cannot extract ECSDA signature components");
+            throw MoCOCrWException("Cannot extract ECDSA signature components");
         }
         auto ecdsa = createManagedOpenSSLObject<SSL_ECDSA_SIG_Ptr>();
         _ECDSA_SIG_set0(ecdsa.get(), std::move(r), std::move(s));
-        return _i2d_ECSDA_SIG(ecdsa.get());
+        return _i2d_ECDSA_SIG(ecdsa.get());
     }
 }
 
@@ -485,18 +488,18 @@ namespace {
      */
     class ECDSASignaturePublicKeyCtx::Impl : public ECDSAImpl<AsymmetricPublicKey> {
     public:
-        Impl(const AsymmetricPublicKey& key, openssl::DigestTypes hashFunction, ECSDASignatureFormat format)
+        Impl(const AsymmetricPublicKey& key, openssl::DigestTypes hashFunction, ECDSASignatureFormat format)
             : ECDSAImpl{key, hashFunction},
               _sigFormat{format}
             {}
 
         void verifyDigest(const std::vector<uint8_t> &signature,
                           const std::vector<uint8_t> &messageDigest) {
-            if (_sigFormat == ECSDASignatureFormat::IEEE1363) {
+            if (_sigFormat == ECDSASignatureFormat::IEEE1363) {
                 size_t keySizeBytes = (key.getKeySize() + 7) / 8;
                 auto asn1Signature = _IEEE1363EcSignatureToAsn1ECSignature(signature, keySizeBytes);
                 _verifyAsn1(asn1Signature, messageDigest);
-            } else if (_sigFormat == ECSDASignatureFormat::ASN1_SEQUENCE_OF_INTS) {
+            } else if (_sigFormat == ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS) {
                 _verifyAsn1(signature, messageDigest);
             } else {
                 throw MoCOCrWException("ECDSA Signature type not recognized.");
@@ -533,17 +536,22 @@ namespace {
             }
         }
 
-        ECSDASignatureFormat _sigFormat;
+        ECDSASignatureFormat _sigFormat;
     };
 
     ECDSASignaturePublicKeyCtx::ECDSASignaturePublicKeyCtx(const AsymmetricPublicKey& key,
                                                            openssl::DigestTypes hashFunction,
                                                            ECSDASignatureFormat format)
+        : _impl(std::make_unique<ECDSASignaturePublicKeyCtx::Impl>(key, hashFunction, SDAtoDSASignatureFormat(format))) {}
+
+    ECDSASignaturePublicKeyCtx::ECDSASignaturePublicKeyCtx(const AsymmetricPublicKey& key,
+                                                           openssl::DigestTypes hashFunction,
+                                                           ECDSASignatureFormat format)
         : _impl(std::make_unique<ECDSASignaturePublicKeyCtx::Impl>(key, hashFunction, format)) {}
 
     ECDSASignaturePublicKeyCtx::ECDSASignaturePublicKeyCtx(const AsymmetricPublicKey& key,
                                                            openssl::DigestTypes hashFunction)
-        : ECDSASignaturePublicKeyCtx(key, hashFunction, ECSDASignatureFormat::ASN1_SEQUENCE_OF_INTS) {}
+        : ECDSASignaturePublicKeyCtx(key, hashFunction, ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS) {}
 
     ECDSASignaturePublicKeyCtx::ECDSASignaturePublicKeyCtx(const X509Certificate& cert,
                                                            openssl::DigestTypes hashFunction)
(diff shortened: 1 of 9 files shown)


neverpanic (approved):

src/asymmetric_crypto_ctx.cpp
```diff
@@ -409,27 +409,82 @@ namespace mococrw {
...
+    std::vector<uint8_t> _Asn1ECSignatureToIEEE1363EcSignature(const std::vector<uint8_t> &signature,
+                                                               size_t keySizeBytes) {
+        auto ecdsa = _d2i_ECDSA_SIG(signature);
+
+        auto r = _ECDSA_SIG_get0_r(ecdsa.get());
+        auto s = _ECDSA_SIG_get0_s(ecdsa.get());
+
+        std::vector<uint8_t> ieee(keySizeBytes*2);
+
+        BN_bn2binpad(r, ieee.data(), keySizeBytes);
```
garw: These are raw openssl functions? Wrapping an error handling seems to be missing

garw (changes requested):
BN_bn2binpad seems to have been used directly from OpenSSL

src/mococrw/openssl_wrap.h
```diff
@@ -1412,5 +1427,17 @@ std::vector<uint8_t> _i2d_ECSDA_SIG(const ECDSA_SIG*);
...
+ * Write an OpenSSL bignum in big-endian plain bytes representation of an
+ * unsigned integer.
+ *
+ * @param bignum Number to write to
+ * @param to Array of chars to write the number to
+ * @param tolen Length of to
+ * @return Number of bytes written into to
+ * @throw OpenSSLException if the supplied buffer is too small
+ */
+int _BN_bn2binpad(const BIGNUM *bignum, unsigned char *to, int tolen);
```
neverpanic: Argh, that stuff is so easy to miss :(

Usually I'd say we should convert this to a safer API returning a std::vector of the given size instead of providing a buffer to write to, i.e. this would be

```cpp
std::vector<uint8_t> _BN_bn2binpad(const BIGNUM* bignum, int tolen) {
    std::vector<uint8_t> out(tolen);
    int written = OpensslCallIsPositive::callChecked(lib::OpenSSLLib::SSL_BN_bn2binpad, bignum, out.data(), out.size());
    out.resize(written);
    return out;
}
```

Of course, that would modify the implementation of your conversion to be

```cpp
return _BN_bn2binpad(r, keysize) + _BN_bn2binpad(s, keysize);
```

neverpanic (commented):
Did we really have a typo in ECDSA and missed it?

src/mococrw/openssl_wrap.h
blenk92: modified the `_BN_bn2binpad` function. However i skipped the resize of the vector because especially in the current usecase it is important that out has exactly the length of the key size (because that is the IEEE signature format). In addition, it seems that vectors can not be concatenated using the + operator...

src/mococrw/openssl_wrap.h
```diff
@@ -1396,11 +1396,26 @@ std::vector<uint8_t> _EVP_derive_key(const EVP_PKEY *peerkey, const EVP_PKEY *ke
...
+
+/**
+ * Get the s signature component as bignum
+ */
+const BIGNUM* _ECDSA_SIG_get0_s(const ECDSA_SIG* sig);
+
 /**
  * Get the serialized ECSDA signature in ASN.1 format from ECSDA_SIG object
  */
 std::vector<uint8_t> _i2d_ECSDA_SIG(const ECDSA_SIG*);
```
wusto: There is still a typo :)

wusto (changes requested):

src/mococrw/openssl_wrap.h
blenk92: thx, found some more typos in comments and exception messages :-D

src/asymmetric_crypto_ctx.cpp
```diff
@@ -400,7 +400,11 @@ namespace mococrw {
      * #  ECDSA  #
      * ###########
      */
-
+namespace {
+    ECDSASignatureFormat SDAtoDSASignatureFormat (ECSDASignatureFormat sigFormat) {
```
neverpanic, line 408:
```suggestion
    /**
     * This is required for backwards compatibility, since we had a typo in ECDSA for a while.
     */
    ECDSASignatureFormat SDAtoDSASignatureFormat(ECSDASignatureFormat sigFormat) {
```

neverpanic (commented):
A minor nitpick, otherwise lg2m

src/asymmetric_crypto_ctx.cpp
```diff
@@ -401,6 +401,27 @@ namespace mococrw {
...
+    std::vector<uint8_t> _Asn1ECSignatureToIEEE1363EcSignature(const std::vector<uint8_t> &signature,
+                                                               size_t keySizeBytes) {
+        auto ecdsa = _d2i_ECDSA_SIG(signature);
+
+        auto r = _ECDSA_SIG_get0_r(ecdsa.get());
+        auto s = _ECDSA_SIG_get0_s(ecdsa.get());
+
+        std::vector<uint8_t> ieee(keySizeBytes*2);
+
+        BN_bn2binpad(r, ieee.data(), keySizeBytes);
```
wusto: I guess Martin was already leaving a comment here.
From openssl doc: 
BN_bn2binpad() returns the number of bytes written or -1 if the supplied buffer is too small.

Can you please either check the return value or add a comment why this is not necessary here?
Surely you are free to add this function to mococrw using add_function.py :)

src/asymmetric_crypto_ctx.cpp
```diff
@@ -472,11 +517,11 @@ namespace {
         auto r = _BN_bin2bn(signature.data(), keySizeBytes);
```
wusto, line 520: Can you please have a quick look, if moving this function to the impl class would resolve https://asc.bmwgroup.net/mgujira/browse/MGUSEC-4419?
If so I would favour to do so here. :)
But please clarify with Clemens first, if he wants the ticket merging :D

wusto (changes requested):

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -408,6 +408,24 @@ class RSASignaturePublicKeyCtx : public DigestVerificationCtx, public MessageVer
...
+ * @brief Serialization formats for ECDSA signatures
+ */
+enum class ECSDASignatureFormat {
+    ASN1_SEQUENCE_OF_INTS, /**< Encoding of (r,s) as ASN.1 sequence of integers as specified in ANSI X9.62 */
+    IEEE1363, /**< Encoding of (r,s) as raw big endian unsigned integers zero-padded to the key length
+               *   as specified in IEEE 1363 */
+};
+
+/**
+ * @brief Serialization formats for ECDSA signatures (correctly spelled)
```
garw: remove the correctly spelled and rather put (with typo) to the other one

src/mococrw/asymmetric_crypto_ctx.h
```diff
@@ -495,6 +505,17 @@ class ECDSASignaturePublicKeyCtx : public DigestVerificationCtx, public MessageV
     ECDSASignaturePublicKeyCtx(const AsymmetricPublicKey& key, openssl::DigestTypes hashFunction,
```
garw, line 517: add [[depricated]] ?

garw (changes requested):

src/mococrw/asymmetric_crypto_ctx.h
blenk92: https://en.cppreference.com/w/cpp/language/attributes/deprecated
am I missing something here? the method should be deprecated already?

src/mococrw/asymmetric_crypto_ctx.h
blenk92: done

src/asymmetric_crypto_ctx.cpp
blenk92: yeah, rebase with tears :-D

src/asymmetric_crypto_ctx.cpp
blenk92: tbh, the ticket also says:
'Moreover, pimpl classes (or rather data-structs) use multiple inheritance in the presense of data members, which I would generally consider to be discouraged.'
This is a bigger change which i would like to see in a different commit. I've just fixed the typos here...

blenk92 (commented):
I rebased with tears, sorry for the confusion

src/mococrw/asymmetric_crypto_ctx.h
garw: I'm pretty sure that this wasn't there when I last had a look at this but I guess Githubs review interface will never let us know…

src/mococrw/openssl_wrap.h
```diff
@@ -1412,5 +1427,17 @@ std::vector<uint8_t> _i2d_ECSDA_SIG(const ECDSA_SIG*);
  */
 SSL_BIGNUM_Ptr _BN_bin2bn(const uint8_t* data, size_t dataLen);
 
+/**
+ * Write an OpenSSL bignum in big-endian plain bytes representation of an
+ * unsigned integer.
+ *
+ * @param bignum Number to write to
```
garw: Rather to serialize? it is not written to the bignum param. In fact, it is const.

src/mococrw/openssl_wrap.h
```diff
@@ -1412,5 +1427,17 @@ std::vector<uint8_t> _i2d_ECSDA_SIG(const ECDSA_SIG*);
...
+ * @param to Array of chars to write the number to
```
garw: to? (see also below)

src/mococrw/openssl_wrap.h
```diff
@@ -1412,5 +1427,17 @@ std::vector<uint8_t> _i2d_ECSDA_SIG(const ECDSA_SIG*);
...
+ * @param tolen Length of to
+ * @return Number of bytes written into to
+ * @throw OpenSSLException if the supplied buffer is too small
```
garw: Buffer doesn't exist anymore?

garw (changes requested):
a few documentation issues. Looks otherwise good to me now

garw (approved):

wusto (approved):

PR #72, opened 2020-05-26
Asymetric crypto pimpl clean up and clang-tidy corrections

wusto:
The PIMPL pattern was not correctly applied to the asymmetry crypto
contexts. E.g. the pimpl classes were just used to hold the variables
but the actual implementation was outside of the pimpl class.

clang-tidy complains about a virtual function call during destruction in
the OpenSSLLibMock.
This patch adds an instruction for clang-tidy to ignore it.

### DIFF
--- a/src/asymmetric_crypto_ctx.cpp
+++ b/src/asymmetric_crypto_ctx.cpp
@@ -100,16 +100,6 @@ namespace mococrw {
 
 
     /*
-     * Base class for signature contexts that support specifying a hash function
-     */
-    class SignatureCtxImpl {
-    public:
-        SignatureCtxImpl(openssl::DigestTypes hashFunction) : hashFunction(hashFunction) {}
-
-        openssl::DigestTypes hashFunction;
-    };
-
-    /*
      * ####################
      * #  RSA Encryption  #
      * ####################
@@ -127,15 +117,15 @@ namespace mococrw {
         /*
          * Constructor that checks the type of the given key
          */
-        RSAImpl(const Key& key, std::shared_ptr<PaddingBase> padding) : key(key), padding(padding) {
+        RSAImpl(const Key& key, std::shared_ptr<PaddingBase> padding) : _key(key), _padding(padding) {
             // Not really nice but necessary since we can't get rid of the generic keypair
             if (key.getType() != AsymmetricKey::KeyTypes::RSA) {
                 throw MoCOCrWException("Expected RSA Key for RSA operation");
             }
         }
-
-        Key key;
-        std::shared_ptr<PaddingBase> padding;
+    protected:
+        Key _key;
+        std::shared_ptr<PaddingBase> _padding;
     };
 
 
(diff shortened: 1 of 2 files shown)


src/asymmetric_crypto_ctx.cpp
```diff
@@ -443,6 +461,7 @@ namespace {
         Key key;
     protected:
         ECDSASignatureFormat _sigFormat;
+        openssl::DigestTypes hashFunction;
```
garw: This library prefixes class variables with _

src/asymmetric_crypto_ctx.cpp
```diff
@@ -133,7 +123,7 @@ namespace mococrw {
                 throw MoCOCrWException("Expected RSA Key for RSA operation");
             }
         }
-
+    protected:
```
garw, line 126: This library prefixes class members with _. When you do a cleanup anyways,  you can just adapt it here as well.

src/asymmetric_crypto_ctx.cpp
```diff
@@ -284,17 +282,43 @@ namespace mococrw {
...
      */
     template <class Key>
-    class RSASignatureImpl : public RSAImpl<Key, RSASignaturePadding>, public SignatureCtxImpl {
+    class RSASignatureImpl : public RSAImpl<Key, RSASignaturePadding> {
     public:
         RSASignatureImpl(const Key& key, openssl::DigestTypes hashFunction, std::shared_ptr<RSASignaturePadding> padding)
-            : RSAImpl<Key, RSASignaturePadding>(key, padding), SignatureCtxImpl(hashFunction) {}
+            : RSAImpl<Key, RSASignaturePadding>(key, padding), hashFunction(hashFunction) {}
+    protected:
+        openssl::DigestTypes hashFunction;
```
garw: This library prefixes class members with _

garw (changes requested):
some (small) nitpicking

src/asymmetric_crypto_ctx.cpp
wusto: Done

src/asymmetric_crypto_ctx.cpp
wusto: Done

src/asymmetric_crypto_ctx.cpp
wusto: Done

src/asymmetric_crypto_ctx.cpp
```diff
@@ -127,15 +117,15 @@ namespace mococrw {
         /*
          * Constructor that checks the type of the given key
          */
-        RSAImpl(const Key& key, std::shared_ptr<PaddingBase> padding) : key(key), padding(padding) {
+        RSAImpl(const Key& key, std::shared_ptr<PaddingBase> padding) : _key(key), _padding(padding) {
```
neverpanic, line 120: Please remove the issue references from the commit messages – they don't make sense, since this is open source code and nobody has access to our internal ticket system.

neverpanic (changes requested):
Commit msg nitpick, lg2m otherwise.

neverpanic (approved):

garw (approved):

PR #73, opened 2020-05-26
Releasing & Versioning

blenk92:
Add release and versioning documentation. Also created commit for initial release. My propsoal would we to start at version 3.0.0 to keep the so-version and release version in sync (We've already increased the so-version to 2.0 in the past)
Release will be created as soon as this PR is merged.


README.md
```diff
@@ -20,23 +20,28 @@ OpenSSL with the following goals in mind:
...
  * X509 certificate processing
  * X509 certificate CA functionality
  * CRLs
- * RSA key management
+ * RSA Encryption and Signatures
+ * ECDSA Signatures
+ * EdDSA Signatures
+ * ECIES Encryption
+ * AES Encryption
+ * SHA 1/2/3 Hashing
```
neverpanic, line 40: Add KDF, message authentication codes?

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+### Public API
+We consider the public API of this project to be whatever is included in the
+`LIBRARY_PUBLIC_HEADERS` defined in `src/CMakeLists.txt`. Currently, this includes
+all the header files in `src/mococrw/`. Changes to the files will be versioned according
+to semantic versioning scheme.
+
+## PRs & Releasing
+
+### PR Checklist
+(probably to be extended)
```
neverpanic: Remove this comment. If we want to extend this, we will, if we don't want to, we'll have this comment sitting there that would no longer be correct.

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+
+What should be done with every PR (if no new release is created):
+* CHANGELOG is updated with the details of the change (Unreleased section)
+
+### Releasing Checklist
+(probably to be extended)
```
neverpanic: Same here.

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+
+Check if the following steps have been performed **before** creating a new release. This includes
+all kinds of releases (Major, Minor and Patch releases)
+* README is updated if necessary
+* CHANGELOG is updated (add the new release)
+* Library .so version is increased (Major releases only)
```
neverpanic: We should only increase the SOVERSION iff the ABI changes. There is no need to increase the SOVERSION if we do not break the ABI.

Also note that the library SOVERSION is not related to the MoCOCrW release version number. There is no need to have a MoCOCrW 2.0 install a `libmococorw.so.2` file.

Consequently, I'm not sure using 3.0 as initial version number is necessary.

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+
+If all the steps above have been done you are good to go and create a new github release.
```
neverpanic: Maybe we should recommend pushing a signed annotated tag, and then promote that to a GitHub release using the GitHub UI? WDYT?

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+This will also create a tag with the version number (make sure to increase the version number
+according to [semver.org](https://semver.org/))
+
+### Release Frequency
+We are developing this library based on our needs. That is, there won't be regular releases.
```
neverpanic:
```suggestion
We are developing this library based on our needs. That is, there is no regular release schedule.
```

"won't" is colloquial.

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+Instead we are going to create releases based on added functionality.
+
+### Release Strategy:
+In general, we won't create a new release for every PR. The following specifies what
```
neverpanic:
```suggestion
In general, we will not create a new release for every PR. The following specifies what
```

README.md
```diff
@@ -74,4 +79,8 @@ In order to generate the doxygen documentation please follow the next steps:
 on the project build directory run:
 
 cmake -DBUILD_DOCUMENTATION=ON ..
-make doc
\ No newline at end of file
+make doc
+
+## Versioning and Releases
+From now on, this library uses Semantic Versioning as described at [semver.org](https://semver.org).
```
garw: From now on sounds weird in a seemingly timeless README. Just drop it.

RELEASING.md
garw: ? What is this supposed to mean?

RELEASING.md
garw: Same here…

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+all the header files in `src/mococrw/`. Changes to the files will be versioned according
+to semantic versioning scheme.
+
+## PRs & Releasing
+
+### PR Checklist
+(probably to be extended)
+
+What should be done with every PR (if no new release is created):
+* CHANGELOG is updated with the details of the change (Unreleased section)
```
garw: Rather with a summary of the change than the details? Nobody cares about (implementation) details.

garw: Additionally: CI testing against project specific CI systems.

CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
```
garw, line 3: Release note for Version 3.0 that the library now uses semantic versioning and has up to date documentation?

README.md
```diff
@@ -20,23 +20,28 @@ OpenSSL with the following goals in mind:
...
 The library provides end-user interfaces for:
  * X509 certificate processing
  * X509 certificate CA functionality
  * CRLs
- * RSA key management
+ * RSA Encryption and Signatures
+ * ECDSA Signatures
+ * EdDSA Signatures
+ * ECIES Encryption
+ * AES Encryption
```
garw: (including GCM to support authenticated encryption with additional data)

README.md
```diff
@@ -20,23 +20,28 @@ OpenSSL with the following goals in mind:
...
 
 The library provides end-user interfaces for:
  * X509 certificate processing
  * X509 certificate CA functionality
  * CRLs
- * RSA key management
+ * RSA Encryption and Signatures
+ * ECDSA Signatures
+ * EdDSA Signatures
+ * ECIES Encryption
```
garw: According to this IEEE standard?

RELEASING.md
garw: hint to where: `src/CMakeLists.txt`?

garw (changes requested):

RELEASING.md
blenk92: Hmm yeah let's try to decouple that. But I guess ABI will be broken on any major release anyway. But we could skip having a major release for every ABI breaking change.

blenk92: done

RELEASING.md
blenk92: done

RELEASING.md
blenk92: done

RELEASING.md
blenk92: done

blenk92: done

RELEASING.md
blenk92: done

RELEASING.md
blenk92: done

blenk92: done

README.md
blenk92: done

README.md
blenk92: done

README.md
blenk92: done

README.md
blenk92: done

CHANGELOG.md
blenk92: done

RELEASING.md
blenk92: sound good, done

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+
+# Unreleased
+
+N/A
+
+# Release 1.0.0
+* Initial Release
+* Library now uses semantic versioning
+* Updated documentationt
+* Increased so.version to 3.0
```
neverpanic:
```suggestion
* Increased soversion to 3.0
```
The dot looks weird to me, suggest removing it.

README.md
```diff
@@ -20,23 +20,30 @@ OpenSSL with the following goals in mind:
...
 
 The library provides end-user interfaces for:
  * X509 certificate processing
  * X509 certificate CA functionality
  * CRLs
- * RSA key management
+ * RSA Encryption and Signatures
+ * ECDSA Signatures
+ * EdDSA Signatures
+ * ECIES Encryption (according to IEEE IEEE 1363a-2004)
```
neverpanic:
```suggestion
 * ECIES Encryption (according to IEEE 1363a-2004)
```

README.md
neverpanic: Doesn't seem to be done, still says "from now on".

neverpanic (commented):
A couple of minor things

README.md
blenk92: done

CHANGELOG.md
blenk92: done

README.md
blenk92: ah, sorry i've misread the comment. Should be done now

neverpanic (approved):

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+# Unreleased
+
+N/A
+
+# Release 1.0.0
```
garw: Well, if the SO version is now 3.0. I would argue that this is now also Release 3.0. This is the whole point of this versioning scheme…

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+* Initial Release
+* Library now uses semantic versioning
+* Updated documentationt
```
garw:
```suggestion
* Updated documentation
```

garw (dismissed):
SOVersion and lib version should coincide to avoid confusion.

CHANGELOG.md
blenk92: @neverpanic requesed to change it to 1.0.0  when i had release 3.0.0. I agree that this could be confusing, on the other hand this would mean that we have to major release on ABI breaking changes (But then i would also put this in releasing requirements). Taking ABI into account was also requested https://github.com/semver/semver/issues/162 at semver years ago, but has not beed added (not even optionally) @garw @neverpanic What to do?

neverpanic: @garw As I understand semver, it considers API only, but API and ABI can sometimes have breaking changes independent from each other.

For example, if we change the value of a constant in a header, or re-order an enum, that would break ABI, but the API would stay the same. Since the API didn't change, according to semver, the version number should not be increased -- however, the SOVERSION would have to.

For this reason, I would have decoupled SOVERSION from the library version number according to semver. Opinions?

garw: Well, I don't think that we have to be very religious with the semantic versioning idea. We should just feel free to also increase the version if the ABI breaks even if this is not strictly what the semantic versioning says but it is what (imho) makes sense and also matches the intention of semantic versioning.

So maybe we just put a comment in that we will not strictly follow the rules but will increase the version number also in case of ABI incompatible changes. I see no benefit in sticking to the rules just for the purpose of sticking to the rules.

As an advantage, we could have SO-version = library version - which from my perspective is something that simplifies a lot of things and reduces confusion.

neverpanic (approved):

PR #74, opened 2020-06-17
Add missing parameter in AESCipherBuilder docs

neverpanic:
AESCipherBuilder takes three arguments, but only two were specified in some of the documentation comments. Add the key size parameter.

I figured since this doesn't affect API and only fixes a small oversight, it's not really worth writing a changelog entry.


blenk92 (approved):

abmantis (approved):

neverpanic:
The CI did run and passed, but failed to report its status to GitHub. I'll close and re-open to retrigger it – unfortunately there doesn't seem to be a button to do so in Travis.

Issue #75, opened 2020-06-22
Cannot compile with OpenSSL 1.1.0l


dima91:
Hello!
I'm trying to compile this library in a docker image with OpenSSL at version 1.1.0l, but when I start to compile the process fails immediately:
```
root@4cba3a1fe286:~# cmake -Bbuild -SMoCOCrW -DBUILD_TESTING=Off                       
CMake Warning (dev) in CMakeLists.txt:
  No project() command is present.  The top-level CMakeLists.txt file must
  contain a literal, direct call to the project() command.  Add a line of
  code such as

    project(ProjectName)

  near the top of the file, but after cmake_minimum_required().

  CMake is pretending there is a "project(Project)" command on the first
  line.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is GNU 6.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ - works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so (found version "1.1.0l")  
-- Using OpenSSL 1.1.0l
-- Found Boost: /usr/include (found version "1.62.0")  
-- Configuring done
-- Generating done
-- Build files have been written to: /root/build
root@4cba3a1fe286:~# cmake --build build
Scanning dependencies of target mococrw
[  4%] Building CXX object src/CMakeFiles/mococrw.dir/asn1time.cpp.o
In file included from /root/MoCOCrW/src/mococrw/asn1time.h:21:0,
                 from /root/MoCOCrW/src/asn1time.cpp:19:
/root/MoCOCrW/src/mococrw/openssl_wrap.h:1349:13: error: 'NID_ED448' was not declared in this scope
     Ed448 = NID_ED448,
             ^~~~~~~~~
/root/MoCOCrW/src/mococrw/openssl_wrap.h:1350:15: error: 'NID_ED25519' was not declared in this scope
     Ed25519 = NID_ED25519
[...]

PR #76, opened 2020-06-22
symmetric: Avoid non-standard explicit_bzero(3)

neverpanic:
`explicit_bzero(3)` is a glibc extension and may not be available, for example on systems using musl. We have a good alternative in `util::vectorCleanse()`, which uses `OPENSSL_cleanse(3)`, so use that instead.

See: #75


garw (approved):
lgtm

PR #80, opened 2020-07-21
Compilable examples

wusto:
These patches add compilable examples to the MoCOCrW libary.
The examples provide an easy start for further development with the MoCOCrW library.


src/mococrw/openssl_wrap.h
```diff
@@ -29,8 +29,27 @@
 
 namespace mococrw
 {
+
+/**
+ * Enum for the digest types that can be used in signatures
+ * or hash computations.
+ */
+enum class DigestTypes {
```
neverpanic, line 37: Should we have a changelog entry for this, since it changes public API (although preserving compatibility)?

Also, this probably changes ABI, since the type is now called mococrw::DigestTypes and no longer mococrw::openssl::DigestTypes, right?

src/mococrw/private/IOUtils.h
```diff
@@ -51,6 +51,13 @@ auto openSSLObjectFromFile(const std::string &filename)
     return Func({buffer.data(), buffer.size()});
 }
 
+template<class Res, Res(Func)(const std::string&, const std::string&)>
```
neverpanic: Rather than requiring to specify the return type separately in Res, couldn't we use `std::result_of_t<>` from `type_traits` to automatically compute it? I'm thinking

```cpp
template<class F>
std::result_of_t<F&&(const std::string&, const std::string&)> openSSLObjectFromFile(F&& f, const std::string& filename, const std::string& password)
{
    auto buffer = bytesFromFile<char>(filename);
    return f({buffer.data(), buffer.size()}, password);
}
```

and then change the invocation to

```cpp
  return openSSLObjectFromFile(&mococrw::AsymmetricPrivateKey::readPrivateKeyFromPEM, filename, password);
```

Unfortunately `std::result_of_t<>` is undefined in C++11 when the given function cannot be invoked with the two `const std::string&` arguments, and only C++14 changes this to SFINAE, which is what we want in this case:

> As formulated in C++11, the behavior of std::result_of is undefined when INVOKE(std::declval<F>(), std::declval<ArgTypes>()...) is ill-formed (e.g. when F is not a callable type at all). C++14 changes that to a SFINAE (when F is not callable, std::result_of<F(ArgTypes...)> simply doesn't have the type member). 

Is MoCOCrW C++14 already? If not we'd probably have to protect this with a `std::enable_if_t<std::is_invocable_v<F(const std::string&, const std::string&)>> = 0`, except that `std::is_invocable` is C++17 only, so we'd have to either copy an implementation (https://stackoverflow.com/a/51188325) or use one from boost.

src/mococrw/private/IOUtils.h
```diff
@@ -75,3 +82,21 @@ mococrw::AsymmetricPublicKey loadPubkeyFromFile(const std::string &filename)
...
+{
+    return openSSLObjectFromFile<mococrw::AsymmetricPrivateKey,
+        mococrw::AsymmetricPrivateKey::readPrivateKeyFromPEM>(filename, password);
+}
+
+void writePemToFile(const std::string &pem, const std::string &filename)
+{
+    std::ofstream file{filename, std::ofstream::out};
+    if (!file.good()) {
+        std::string errorMsg{"Cannot open file for writing key."};
```
neverpanic:
```suggestion
        std::string errorMsg{"Cannot open file for writing PEM."};
```

I guess this may also be useful for other PEM-encoded structs, not only keys, right? Also, the formatting can be improved here, since ideally the output would be:

`Cannot open file "filename" for writing: std::sterror(errno)`.

src/CMakeLists.txt
```diff
@@ -60,14 +60,17 @@ set(LIBRARY_PUBLIC_HEADERS
...
 
+set(LIBRARY_PRIVATE_HEADER_DIRS
+    mococrw/private/
+)
+
 
 add_library(${LIBRARY_NAME} SHARED ${LIBRARY_SOURCES} ${LIBRARY_PUBLIC_HEADERS})
 add_library(${PROJECT_NAME}::${LIBRARY_NAME} ALIAS ${LIBRARY_NAME})
-target_include_directories(${LIBRARY_NAME} PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/cwrap")
+target_include_directories(${LIBRARY_NAME} PRIVATE ${LIBRARY_PRIVATE_HEADER_DIRS})
```
neverpanic: Is this even needed if you're including the headers with their relative path `mococrw/private` anyway? I don't think it is, in which case this should be removed.

examples/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+project(mococrw-examples)
+
+if(BUILD_EXAMPLES)
+    message(STATUS "Compiling examples")
+    set(EXAMPLE_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+    FIND_PACKAGE(Boost COMPONENTS program_options REQUIRED)
```
neverpanic:
```suggestion
    find_package(Boost COMPONENTS program_options REQUIRED)
```

Lower case, please, hardly anybody still writes CMake commands in upper case.

examples/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+    set(COMMON_EXAMPLE_HEADERS
```
neverpanic, line 7: Is this variable used at all? If not, I'd just remove this and `$EXAMPLE_SRC_DIR` completely to clean up the file a bit.

examples/common.h
```diff
@@ -0,0 +1,211 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+
+#include <unordered_map>
```
neverpanic: `:'<,'>sort` this block please.

examples/common.h
```diff
@@ -0,0 +1,211 @@
...
+#include <fstream>
+#include <iostream>
+#include <string>
+
+#include <boost/program_options.hpp>
```
neverpanic: Same here.

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+if(BUILD_EXAMPLES)
+    if(BUILD_TESTING)
```
neverpanic:
```suggestion
if(BUILD_EXAMPLES AND BUILD_TESTING)
```

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
...
+	set(EXAMPLE_BINARY_DIR ${CMAKE_BINARY_DIR}/examples)
+	add_test(
+	    NAME "ExampleTesting"
+	    COMMAND sh -c "MOCOCRW_EXAMPLE_BINARY_PATH=${EXAMPLE_BINARY_DIR} MOCOCRW_SRC_DIR=${EXAMPLE_DIR} /usr/bin/python3 -m unittest"
```
neverpanic: I don't think this will work when cross compiling with `CMAKE_CROSSCOMPILING_EMULATOR` set unless you modify the python script to start the target binary with the emulator manually.

tests/examples/common.py
```diff
@@ -0,0 +1,63 @@
+import os
+import tempfile
+import subprocess
+
+MOCOCRW_EXAMPLE_BINARY_PATH = "."
+try:
+    MOCOCRW_EXAMPLE_BINARY_PATH = os.environ["MOCOCRW_EXAMPLE_BINARY_PATH"]
+except KeyError:
+    print("Please set the MOCOCRW_EXAMPLE_BINARY_PATH environment variable.")
+    exit(1)
```
neverpanic: Why have a default value if you're going to `exit(1)` anyway on error? Also, I think this should be

```python
print(..., file=sys.stderr)
```

and

```python
sys.exit(1)
```

tests/examples/common.py
```diff
@@ -0,0 +1,63 @@
...
+]
+
+
+def add_invalid_parameters_and_execute(execute_string):
+    def execute(exec_string):
+        result = subprocess.run(exec_string.split(), capture_output=True)
+        assert result.returncode == 1
+        assert b"Unknown option selected: unrecognised option '--aw\xc3\xb6eljf'\n" \
+               b"Failure parsing command line.\n" == result.stderr
+    execute(execute_string + " --awöeljf asdölvkjyxc")
```
neverpanic: I would prefer if we didn't use strings but lists of arguments here so we can avoid another shell indirection (which may be required for this to work in a cross compilation environment anyway):

```suggestion
def add_invalid_parameters_and_execute(parameters):
    parameters.extend(["--awöeljf", "asdölvkjyxc"])
    result = subprocess.run(parameters, capture_output=True)
    assert result.returncode == 1
    assert ...
```

tests/examples/common.py
```diff
@@ -0,0 +1,63 @@
...
+    # Create working directory
+    new_working_dir = tempfile.mkdtemp()
+
+    current_working_dir = os.getcwd()
+    # Change to working directory
+    os.chdir(new_working_dir)
+    return current_working_dir, new_working_dir
+
+
+def combine_all_options(all_options, option_list):
```
neverpanic, line 44: Please write a documentation comment to explain what this function does, because I can't figure it out from reading the source.

tests/examples/common.py
```diff
@@ -0,0 +1,63 @@
...
+    first_options_list = option_list[0]
+    add_options = [add_opt + opt for add_opt in first_options_list for opt in all_options]
+    all_options += add_options
+    if len(option_list) == 1:
+        return all_options
+    return combine_all_options(all_options, option_list[1:])
+
+
+def build_option_list(options_dict):
+    if len(options_dict) == 0:
```
neverpanic: Usually, that's written as
```suggestion
    if not options_dict:
```
in Python.

tests/examples/common.py
```diff
@@ -0,0 +1,63 @@
...
+        return [""]
```
neverpanic, line 91: Are you sure you want to return a 1-element list with an empty string, rather than just an empty list?
Also, can you please document what this function does?

examples/common.h
```diff
@@ -0,0 +1,211 @@
...
+            ("chaining", "Print results in a form that they can be used as command line arguments for "
+                         "decryption/verification. Can't be combined with verbose")
+            ;
+    return desc;
+}
+
+bool isEncryptionOptionsValid(const po::variables_map &vm)
+{
+    // antivalence check (only one must be defined)
+    if (!(vm.count("decrypt") xor vm.count("encrypt"))) {
```
neverpanic:
```suggestion
    if (!(vm.count("decrypt") ^ vm.count("encrypt"))) {
```

examples/common.h
```diff
@@ -0,0 +1,211 @@
...
+    signVerify.add_options()
+            ("sign", "Sign the data")
+            ("verify", "Verify the data")
+            ;
+    return signVerify;
+}
+
+bool isGetAsymmetricSignAndVerfyOptionsValid(const po::variables_map &vm)
+{
+    if (!(vm.count("sign") xor vm.count("verify"))) {
```
neverpanic:
```suggestion
    if (!(vm.count("sign") ^ vm.count("verify"))) {
```

examples/common.h
```diff
@@ -0,0 +1,211 @@
...
+}
+
+bool parseCommandLineOptions(const po::options_description &cmdlineOptions, po::variables_map &vm,
+                             int argc, char *argv[])
+{
+    try {
+        po::store(po::parse_command_line(argc, argv, cmdlineOptions), vm);
+    } catch (boost::wrapexcept<boost::program_options::invalid_command_line_syntax> &e) {
+        std::cerr << "Failure parsing the input. Please check your input!" << std::endl;
+        std::cerr << e.what();
```
neverpanic:
```suggestion
        std::cerr << e.what() << std::endl;
```

examples/common.h
```diff
@@ -0,0 +1,211 @@
...
+        if (auto v = boost::any_cast<uint32_t>(&value))
+            std::cout << *v;
+        else if (auto v = boost::any_cast<std::string>(&value))
+            std::cout << *v;
+        else if (auto v = boost::any_cast<bool>(&value)) {
+            if (*v == true)
+                std::cout << "True";
+            else if (*v == false)
+                std::cout << "False";
+            else
```
neverpanic: What third state of a boolean variable would you expect here?

examples/common.h
```diff
@@ -0,0 +1,211 @@
...
+                std::cout << "error on boolean";
+        } else
```
neverpanic: Please add brackets everywhere. The branches without brackets are a bug waiting to happen.

neverpanic (changes requested):
I haven't yet read all of it, but there's some room for improvement.

src/mococrw/openssl_wrap.h
wusto: Discussed with @garw that we will do an ABI change and bump the version.

tests/examples/common.py
wusto: After building the option combinations a for-loop is iterating over the elements in the list. An empty list would mean that no test is executed. An empty element expands the "{}".format(element) to an valid expression.

tests/examples/CMakeLists.txt
wusto: What's your suggestion here?
Shall I check if CMAKE_CROSSCOMPILING_EMULATOR is enabled and abort the tests?
I guess that execute the examples using "qemu user mode" might be quite cumbersome as it is not installed on every system.

tests/examples/common.py
neverpanic: Don't explain it to me, write a comment that explains it to your future self.

tests/examples/CMakeLists.txt
neverpanic: Or, detect that, pass the value of CMAKE_CROSSCOMPILING_EMULATOR to the python script and support invoking things through that.

We already execute all other tests like that when cross-compiling, so if any tests should pass when cross-compiling, this setup is already required. If we could skip the Python script, CMake would do that for us, too.

examples/aes_example.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if (authenticatedEncryptor) {
+        std::cout << "Authentication Tag: " << utility::toHex(authenticatedEncryptor->getAuthTag()) << std::endl;
+    }
+}
+
+void aesDecrypt(const AesData &aesData)
+{
+    auto &ciphertext = aesData.data;
+    auto decryptorBuilder = AESCipherBuilder{aesData.operationMode, aesData.secretKey.size() == 32 ?
+                SymmetricCipherKeySize::S_256 : SymmetricCipherKeySize::S_128, aesData.secretKey}.setIV(aesData.iv);
```
blenk92, line 112: can we move the tenary operator into the same line (or even use outside of the builder) to make things more readable here?

examples/aes_example.cpp
```diff
@@ -0,0 +1,131 @@
...
+        authenticatedEncryptor->addAssociatedData(*aesData.authData);
+    }
+
+    return authenticatedEncryptor;
+}
+
+void aesEncrypt(const AesData &aesData)
+{
+    auto &plaintext = aesData.data;
+    auto encryptorBuilder = AESCipherBuilder{aesData.operationMode, aesData.secretKey.size() == 32 ?
```
blenk92, line 41: see below

examples/aes_example.cpp
```diff
@@ -0,0 +1,131 @@
...
+        std::cerr << "Failure encrypting the data." << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+
+    auto iv = encryptor->getIV();
+
+    std::vector<uint8_t> tag;
+
+    // The rest of the function outpus the data
```
blenk92, line 80:
```suggestion
    // The rest of the function outputs the data
```
not sure if this is what you meant

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+}
+
+void parseEncryption(struct AesData &aesData)
+{
+    parseData(aesData);
+}
+
+void parseDecryption(struct AesData &aesData)
```
blenk92, line 185: hmm i would suggest to move the parsing logic into `AesData`. Looks like this is strongly coupled to the class anyway. Or am I missing something here?

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+        ("key", po::value<std::string>(), "The key which shall be used for encryption/decryption in hex form (with or "
+                                          "without 0x prepended). Two sizes are supported: 128 and 256 bits (16 and 32 "
+                                          "bytes).")
+        ("operation-mode", po::value<std::string>(), "The operation mode. Supported: GCM, CTR, CBC. GCM is the only "
+                                                     "mode supporting Authenticated Encryption (AE)")
+        ("padding", po::value<std::string>()->default_value("PKCS"), "The padding mode. Optional. "
+                                                                     "Default: PKCS Supported: PKCS, NO")
+        ("iv", po::value<std::string>(), "The IV value used for en-/decryption in hex form (with or without 0x "
+                                         "prepended). If empty a random value is used for encryption. Default: empty. "
+                                         "This value is mandatory for decrpytion.")
```
blenk92, line 40:
```suggestion
                                         "This value is mandatory for decryption.")
```

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+    genericAesOptions.add_options()
+        ("key", po::value<std::string>(), "The key which shall be used for encryption/decryption in hex form (with or "
+                                          "without 0x prepended). Two sizes are supported: 128 and 256 bits (16 and 32 "
+                                          "bytes).")
+        ("operation-mode", po::value<std::string>(), "The operation mode. Supported: GCM, CTR, CBC. GCM is the only "
+                                                     "mode supporting Authenticated Encryption (AE)")
+        ("padding", po::value<std::string>()->default_value("PKCS"), "The padding mode. Optional. "
+                                                                     "Default: PKCS Supported: PKCS, NO")
+        ("iv", po::value<std::string>(), "The IV value used for en-/decryption in hex form (with or without 0x "
+                                         "prepended). If empty a random value is used for encryption. Default: empty. "
```
blenk92, line 39:
```suggestion
                                         "prepended). If empty, a random value is used for encryption. Default: empty. "
```
I would add a comma here, this makes it easier to get what you meant on the first "parse attempt"

blenk92, line 39: please move the `Default:empty` after `This value is mandatory for decryption.` Maybe also Default Value would be clearer. Or maybe even change to: "if no IV is given, a random IV will be chosen when encrypting. The IV must be given for decryption".

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+                                         "This value is mandatory for decrpytion.")
+    ;
+    po::options_description aesAuthenticatedOptions("Options for authenticated encryption (AE)");
+    aesAuthenticatedOptions.add_options()
+        ("auth-data", po::value<std::string>(), "Authentication data in hex form (with or without 0x "
+                                                "prepended). Optional. Can only be used in conjunction with GCM mode.")
```
blenk92, line 45: Same here. Imho infos such as `Optional` should not be in the middle of the descrption

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+    auto vm = *aesData.vm.get();
+
+    if (vm.count("chaining")) {
+        aesData.chaining = true;
+    }
+
+    std::vector<uint8_t> secretKey;
+    try {
+        aesData.secretKey = utility::fromHex(vm["key"].as<std::string>());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 144:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    std::cerr << "Please select a supported Padding scheme. " << paddingString << " is invalid." << std::endl;
+    exit(EXIT_FAILURE);
+}
+
+std::vector<uint8_t> getIv(const po::variables_map &vm)
+{
+    try {
+        return utility::fromHex(vm["iv"].as<std::string>());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 126:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+        aesData.secretKey = utility::fromHex(vm["key"].as<std::string>());
+    }  catch (MoCOCrWException &e) {
+        std::cerr << "Failure getting key hex string." << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+    std::vector<uint8_t> data;
+    try {
+        aesData.data = utility::fromHex(vm["data"].as<std::string>());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 152:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    if (vm.count("auth-tag-length")) {
+        aesData.authTagLength = vm["auth-tag-length"].as<size_t>();
+    }
+
+    if (vm.count("auth-data")) {
+        try {
+            aesData.authData = utility::fromHex(vm["auth-data"].as<std::string>());
+        }  catch (MoCOCrWException &e) {
```
blenk92, line 172:
```suggestion
        } catch (MoCOCrWException &e) {
```

examples/ca_example.cpp
```diff
@@ -0,0 +1,59 @@
...
+    return std::make_shared<X509Certificate>(std::move(rootCert));
+}
+
+std::shared_ptr<X509Certificate> signCsr(const struct CaData &caData)
+{
+    std::shared_ptr<X509Certificate> cert;
+    auto ca = getCa(caData);
+    try {
+        cert = std::make_shared<X509Certificate>(ca->signCSR(*caData.csr.get()));
+    }  catch (openssl::OpenSSLException &e) {
```
blenk92, line 48:
```suggestion
    } catch (openssl::OpenSSLException &e) {
```

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+        exit(EXIT_FAILURE);
+    }
+
+    // Subtract -1 to avoid race conditions while testing or instant use
+    Asn1Time notBefore(Asn1Time::now() - std::chrono::seconds(1));
+    auto notBeforeString = propertyTree.get(treePrefix + ".notBeforeAsn1", "");
+    if (!notBeforeString.empty()) {
+        try {
+            notBefore = Asn1Time::fromString(notBeforeString);
+        }  catch (openssl::OpenSSLException &e) {
```
blenk92, line 88:
```suggestion
        } catch (openssl::OpenSSLException &e) {
```

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+    parseCaData(caData);
+
+    std::shared_ptr<X509Certificate> cert;
+
+    if (vm.count("create")) {
+        try {
+            // Create a root certificate
+            parseCreateCertificateData(caData);
+            cert = createRootCertificate(caData);
+        }  catch (MoCOCrWException &e) {
```
blenk92, line 297:
```suggestion
        } catch (MoCOCrWException &e) {
```

examples/ecies_example.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+void encrypt_ecies(const struct EciesData &eciesData)
+{
+    std::unique_ptr<ECIESEncryptionCtx> encCtx = buildEncryptionContext(eciesData);
+
+    try {
+         encCtx->update(eciesData.data);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 98:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/ecies_main.cpp
```diff
@@ -0,0 +1,264 @@
...
+}
+
+void parseEncryptData(struct EciesData &eciesData, const po::variables_map &vm)
+{
+    parseCommonData(eciesData, vm);
+    if (vm.count("public-key")) {
+        try {
+            eciesData.pubKey = std::make_shared<AsymmetricPublicKey>(loadPubkeyFromFile(
+                        vm["public-key"].as<std::string>()));
+        }  catch (openssl::OpenSSLException &e) {
```
blenk92, line 214:
```suggestion
        } catch (openssl::OpenSSLException &e) {
```

examples/ecies_main.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return;
+    }
+
+    /* we got a certificate */
+    try {
+        eciesData.cert = std::make_shared<X509Certificate>(loadCertFromFile(vm["cert"].as<std::string>()));
+        succeedOrExit(!isPubKeyAnEccKey(eciesData.cert->getPublicKey()), "No ECC key is provided. Please provide a "
+                                                                         "certificate containing an elliptic curve "
+                                                                         "public key.");
+    }  catch (openssl::OpenSSLException &e) {
```
blenk92, line 230:
```suggestion
    } catch (openssl::OpenSSLException &e) {
```

examples/ecies_main.cpp
```diff
@@ -0,0 +1,264 @@
...
+    eciesData.vm = vm;
+    try {
+        if ((*vm.get()).count("decrypt")) {
+            parseDecryptData(eciesData, *vm.get());
+            decrypt_ecies(eciesData);
+        } else {
+            parseEncryptData(eciesData, *vm.get());
+            encrypt_ecies(eciesData);
+        }
+    }  catch (boost::wrapexcept<boost::program_options::invalid_command_line_syntax> &e) {
```
blenk92, line 252:
```suggestion
    } catch (boost::wrapexcept<boost::program_options::invalid_command_line_syntax> &e) {
```

examples/kdf_main.cpp
```diff
@@ -0,0 +1,126 @@
...
+    succeedOrExit(!vm.count("output-length"), "Please specify a output length in bytes (--output-length");
+}
+
+void getDefaultParams(struct KdfData &kdfData)
+{
+    auto vm = *kdfData.vm.get();
+    try {
+        kdfData.password = utility::fromHex(vm["password"].as<std::string>());
+        kdfData.salt = utility::fromHex(vm["salt"].as<std::string>());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 81:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/mac_example.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+void verify(const struct MacData &macData)
+{
+    auto hmac = doHmac(macData);
+    auto tag = hmac->finish();
+
+    try {
+        hmac->verify(macData.authenticationTag);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 41:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/mac_main.cpp
```diff
@@ -0,0 +1,121 @@
...
+    succeedOrExit(!vm.count("key"), "Please specify a key (--key");
+}
+
+void parseMacData(struct MacData &macData)
+{
+    auto vm = *macData.vm.get();
+    try {
+        macData.message = utility::fromHex(vm["message"].as<std::string>());
+        macData.key = utility::fromHex(vm["key"].as<std::string>());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 80:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/mac_main.cpp
```diff
@@ -0,0 +1,121 @@
...
+    if (macData.digestType == DigestTypes::NONE) {
+        std::cerr << "Failure parsing digest Type: " << vm["hash-algo"].as<std::string>() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+
+    if (vm.count("verify")) {
+        macData.verify = true;
+        try {
+            macData.authenticationTag = utility::fromHex(vm["authentication-tag"].as<std::string>());
+        }  catch (MoCOCrWException &e) {
```
blenk92, line 95:
```suggestion
        } catch (MoCOCrWException &e) {
```

examples/rsa_example.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+    std::cout << utility::toHex(ciphertext) << std::endl;
+}
+
+void decrypt_rsa(const struct RsaData &rsaData) {
+    std::vector<uint8_t> plaintext;
+    try {
+        RSAEncryptionPrivateKeyCtx rsaPrivCtx(*rsaData.privKey.get(), rsaData.padding);
+        plaintext = rsaPrivCtx.decrypt(rsaData.data);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 70:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/rsa_main.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    try {
+        if (vm->count("decrypt")) {
+            parseDecryptData(rsaData, vm);
+            decrypt_rsa(rsaData);
+        } else {
+            parseEncryptData(rsaData, vm);
+            encrypt_rsa(rsaData);
+        }
+    }  catch (boost::wrapexcept<boost::program_options::invalid_command_line_syntax> &e) {
```
blenk92, line 144:
```suggestion
    } catch (boost::wrapexcept<boost::program_options::invalid_command_line_syntax> &e) {
```

examples/sig_example.cpp
```diff
@@ -0,0 +1,203 @@
...
+    std::shared_ptr<MessageSignatureCtx> ctx(nullptr);
+    std::shared_ptr<RSASignaturePadding> rsaPadding = sigData.rsaPadding;
+    if (!rsaPadding) {
+        rsaPadding = std::make_shared<PSSPadding>();
+    }
+
+    try {
+        ctx = std::make_shared<RSASignaturePrivateKeyCtx>(*sigData.privKey.get(), sigData.digestType,
+                                                          rsaPadding);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 35:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_example.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+std::shared_ptr<MessageSignatureCtx>getEccSignCtx(const struct SigData &sigData)
+{
+
+    std::shared_ptr<MessageSignatureCtx> ctx(nullptr);
+    try {
+        ctx = std::make_shared<ECDSASignaturePrivateKeyCtx>(*sigData.privKey.get(), sigData.digestType,
+                                                            sigData.sigFormat);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 50:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_example.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+    return ctx;
+}
+
+std::shared_ptr<MessageSignatureCtx>getEccEdSignCtx(const struct SigData &sigData)
+{
+    std::shared_ptr<MessageSignatureCtx> ctx(nullptr);
+    try {
+        ctx = std::make_shared<EdDSASignaturePrivateKeyCtx>(*sigData.privKey.get());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 63:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_example.cpp
```diff
@@ -0,0 +1,203 @@
...
+        break;
+    default:
+        std::cerr << "Unknown key type. Supported are RSA, ECC and ECC_ED";
+        exit(EXIT_FAILURE);
+    }
+
+    std::vector<uint8_t> signature;
+    try {
+        signature = signCtx->signMessage(sigData.message);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 117:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_example.cpp
```diff
@@ -0,0 +1,203 @@
...
+    std::shared_ptr<MessageVerificationCtx> ctx(nullptr);
+
+    std::shared_ptr<RSASignaturePadding> rsaPadding = sigData.rsaPadding;
+    if (!rsaPadding) {
+        rsaPadding = std::make_shared<PSSPadding>();
+    }
+
+    try {
+        ctx = std::make_shared<RSASignaturePublicKeyCtx>(*sigData.pubKey.get(), sigData.digestType, rsaPadding);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 139:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_example.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+    return ctx;
+}
+
+std::shared_ptr<MessageVerificationCtx> verifyEcc(const struct SigData &sigData)
+{
+    std::shared_ptr<MessageVerificationCtx> ctx(nullptr);
+    try {
+        ctx = std::make_shared<ECDSASignaturePublicKeyCtx>(*sigData.pubKey.get(), sigData.digestType, sigData.sigFormat);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 152:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_example.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+    return ctx;
+}
+
+std::shared_ptr<MessageVerificationCtx> verifyEccEd(const struct SigData &sigData)
+{
+    std::shared_ptr<MessageVerificationCtx> ctx(nullptr);
+    try {
+        ctx = std::make_shared<EdDSASignaturePublicKeyCtx>(*sigData.pubKey.get());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 165:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_main.cpp
```diff
@@ -0,0 +1,206 @@
...
+void parseCommonData(SigData &sigData)
+{
+    auto vm = *sigData.vm.get();
+    if (vm.count("chaining")) {
+        sigData.chaining = true;
+    }
+
+    try {
+        sigData.message = utility::fromHex(vm["message"].as<std::string>());
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 138:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/sig_main.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+void parseSignData(SigData &sigData)
+{
+    auto vm = *sigData.vm.get();
+    parseCommonData(sigData);
+    try {
+        sigData.privKey = std::make_shared<AsymmetricPrivateKey>(loadPrivkeyFromFile(vm["private-key"].as<std::string>(),
+                    vm["private-key-password"].as<std::string>()));
+    }  catch (openssl::OpenSSLException &e) {
```
blenk92, line 164:
```suggestion
    } catch (openssl::OpenSSLException &e) {
```

examples/sig_main.cpp
```diff
@@ -0,0 +1,206 @@
...
+    try {
+        if (vm.count("public-key")) {
+            sigData.pubKey = std::make_shared<AsymmetricPublicKey>(loadPubkeyFromFile(
+                                                                       vm["public-key"].as<std::string>()));
+        } else {
+            auto  cert = loadCertFromFile(vm["cert"].as<std::string>());
+            sigData.pubKey = std::make_shared<AsymmetricPublicKey>(cert.getPublicKey());
+        }
+        sigData.signature = utility::fromHex(vm["signature"].as<std::string>());
+    }  catch (openssl::OpenSSLException &e) {
```
blenk92, line 184:
```suggestion
    } catch (openssl::OpenSSLException &e) {
```

src/parsingUtils.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+#undef addCertDetailIfSet
+
+
+void readJsonConfigFile(pt::ptree &propertyTree, const po::variables_map &vm)
+{
+    std::string filePath = vm["config-file"].as<std::string>();
+    try {
+        pt::json_parser::read_json(filePath, propertyTree);
+    }  catch (boost::property_tree::json_parser::json_parser_error &e) {
```
blenk92, line 121:
```suggestion
    } catch (boost::property_tree::json_parser::json_parser_error &e) {
```

examples/aes_example.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+    else {
+        encryptor = encryptorBuilder.buildEncryptor();
+    }
+
+    std::vector<uint8_t> ciphertext;
+    try {
+        encryptor->update(plaintext);
+        ciphertext = encryptor->finish();
+    }  catch (openssl::OpenSSLException &e) {
```
blenk92, line 66:
```suggestion
    } catch (openssl::OpenSSLException &e) {
```

examples/aes_example.cpp
```diff
@@ -0,0 +1,131 @@
...
+        decryptor = decryptorBuilder.buildAuthenticatedDecryptor();
+        setAuthenticatedCipherData(aesData, decryptor)->setAuthTag(*aesData.authTag);
+    } else {
+        decryptor = decryptorBuilder.buildDecryptor();
+    }
+
+    decryptor->update(ciphertext);
+    try {
+        std::cout << utility::toHex(decryptor->finish()) << std::endl;
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 126:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/aes_main.cpp
```diff
@@ -0,0 +1,216 @@
...
+    succeedOrExit(!vm.count("key"), "You need to provide a key for en-/decryption");
+    succeedOrExit(!vm.count("operation-mode"), "You need to specify an operation mode");
+
+    auto opMode = boost::to_upper_copy<std::string>(vm["operation-mode"].as<std::string>());
+    succeedOrExit(opMode.compare("GCM") && vm.count("auth-data"), "Authentication data is specified but "
+                                                                             "GCM is not selected.");
+    try {
+        auto keySize = utility::fromHex(vm["key"].as<std::string>()).size();
+        succeedOrExit(keySize != 16 && keySize != 32, "The given key size is not supported. Key size: " << keySize);
+    }  catch (MoCOCrWException &e) {
```
blenk92, line 81:
```suggestion
    } catch (MoCOCrWException &e) {
```

examples/ca_example.cpp
```diff
@@ -0,0 +1,59 @@
...
+ * #L%
+ */
+#include "ca_example.h"
+#include <iostream>
+
+std::shared_ptr<CertificateAuthority> getCa(const struct CaData &caData)
+{
+    return std::make_shared<CertificateAuthority>(
+                *caData.signParams,
+                1,
```
blenk92, line 26: a comment that explains what this magic number is would be nice

examples/ca_example.cpp
```diff
@@ -0,0 +1,59 @@
...
+                *caData.rootCert.get(),
+                *caData.privKey.get());
+}
+
+std::shared_ptr<X509Certificate> createRootCertificate(const struct CaData &caData)
+{
+    auto rootCert = CertificateAuthority::createRootCertificate(
+            *caData.privKey.get(),
+            *caData.certDetails,
+            0,
```
blenk92, line 36: same here

examples/ca_example.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+std::shared_ptr<X509Certificate> signCsr(const struct CaData &caData)
+{
+    std::shared_ptr<X509Certificate> cert;
+    auto ca = getCa(caData);
+    try {
+        cert = std::make_shared<X509Certificate>(ca->signCSR(*caData.csr.get()));
+    }  catch (openssl::OpenSSLException &e) {
+        std::cerr << "Error reading CSR in PEM format. Please validate  your CSR." << std::endl;
```
blenk92, line 49:
```suggestion
        std::cerr << "Error reading CSR in PEM format. Please validate your CSR." << std::endl;
```

examples/ca_example.h
```diff
@@ -0,0 +1,42 @@
...
+    std::shared_ptr<X509Certificate> rootCert;
+    std::shared_ptr<CertificateSigningRequest> csr;
+    std::shared_ptr<const pt::ptree> config;
+    std::shared_ptr<const po::variables_map> vm;
+    CaData(std::shared_ptr<const pt::ptree> config, std::shared_ptr<const po::variables_map> vm)
+        : config(config), vm(vm)
+    {}
+};
+
+std::shared_ptr<CertificateAuthority> getCa(const struct CaData &caData);
```
blenk92, line 40: why are these not member functions of `CaData`?

src/mococrw/openssl_wrap.h
neverpanic: I would expect the ChangeLog to be edited in the same commit. Please move the Changelog entry into this commit and make it clear that this is an ABI change.

Also, do not use gerund forms ("moving") in commit messages, but always imperative present tense ("move"). Don't describe what you did or are doing, describe what the commit does.

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+std::shared_ptr<BasicConstraintsExtension> getBasicConstraintsExtension(pt::ptree basicConstraintsSubTree)
+{
+    bool isCA = basicConstraintsSubTree.get("isCA", false);
+    int pathlen = basicConstraintsSubTree.get("pathlength", 0);
+    return std::make_shared<BasicConstraintsExtension>(isCA, pathlen);
+}
+
+template <class Builder, class Type>
+Builder invokeIfTrue(Builder b, std::function<Builder ()> func, Type val)
```
blenk92, line 38: not sure if we need such a function template if it is only used at one place (or at least i could only find a single place). `addKeyUsageIfTrue()` would be much easier to read without this

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+{
+    if (!val) {
+        return b;
+    }
+    return func();
+}
+
+#define addKeyUsageIfTrue(func, str) \
```
blenk92, line 46: also not sure if a macro is a good way to go here. This usually makes it harder to read the code, especially if they are not just wrapping a function call. Can't we do this somehow with a single function?

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+    std::cout << "\tnotBeforeAsn1 [ASN1 UTC Time format string] (If not set the current time is used)" << std::endl;
+    std::cout << "\tdigestType [string] (Available digests: SHA256, SHA384, SHA512, SHA3-256, SHA3-384, SHA3-512) "
+              << std::endl;
+    std::cout << "-----------------------------------------" << std::endl;
+}
+
+void printConfigHelp()
+{
+    std::cout << "The configuration file contains information for signing certificates or creating CSRs" << std::endl;
+    std::cout << "These information are ordered in sections. Each section can contain information for signing "
```
blenk92, line 174:
```suggestion
    std::cout << "This information is ordered in sections. Each section can contain information for signing "
```

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+void parseCommandlineArgs(int argc, char *argv[], po::variables_map &vm)
+{
+    po::options_description caOpts("Options for the certificate authority");
+    caOpts.add_options()
+            ("sign", "Sign a certificate signing request (CSR)")
+            ("create", "Create a root certificate authority. This will return a self signed certificate")
+            ("config-file", po::value<std::string>(), "The config file holding the configuration for the CA and "
+                                                      "intermediate CAs")
+            ("config-section", po::value<std::string>(), "Config section that shall be used (e.g. ca, rootca, ...)")
+            ("config-help", "Prints a help for the configuration file.")
```
blenk92, line 197: `Prints a help` sounds strange to me, but maybe it's just me :-D

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+            ("private-key", po::value<std::string>(), ("Path to the private key."))
+            ("private-key-password", po::value<std::string>()->default_value(""), "Optional: password for the private "
+                                                                                  "key")
+            ("output-path", po::value<std::string>(), "Write the certificate to a file. Define the path here. "
```
blenk92, line 201: What about "Write the certificate to the specified file instead of printing it on stdout"?

examples/ca_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+    auto vm = *caData.vm.get();
+    auto propertyTree = *caData.config.get();
+    const std::string treePrefix(vm["config-section"].as<std::string>());
+
+    caData.signParams = getSigningParams(propertyTree, treePrefix);
+    caData.privKey = std::make_shared<AsymmetricPrivateKey>(loadPrivkeyFromFile(vm["private-key"].as<std::string>(),
+                vm["private-key-password"].as<std::string>()));
+}
+
+void parseCreateCertificateData(struct CaData &caData)
```
blenk92, line 253: same as above, i think such functions should be part of the class...

blenk92 (changes requested):
i didn't look at all yet but want to share some comments already

src/mococrw/private/IOUtils.h
```diff
@@ -44,17 +44,17 @@ std::vector<T> bytesFromFile(const std::string &filename)
     return buffer;
 }
 
-template<class Res, Res(Func)(const std::string&)>
-auto openSSLObjectFromFile(const std::string &filename)
+template<class F, typename... Args>
```
neverpanic, line 48: I like the change, but the commit message is "Make IOUtils.h a private header", and not "extend openSSLObjectFromFile() to arbitrary number of arguments" and "add loadPrivkeyFromFile and writePemToFile".

This should either be part of a separate commit, or the commit message should change.

examples/common.h
```diff
@@ -0,0 +1,205 @@
...
+{
+    po::options_description signVerify("Options for asymmetric signatures and verification");
+    signVerify.add_options()
+            ("sign", "Sign the data")
+            ("verify", "Verify the data")
+            ;
+    return signVerify;
+}
+
+bool isGetAsymmetricSignAndVerfyOptionsValid(const po::variables_map &vm)
```
neverpanic, line 94: Naming is inconsistent here: First, you're mixing snake_case with camelCase, second you use get_encryption_options_description and isEncryptionOptionsValid (which should arguably be areEncryptionOptionsValid), but for get_asymmetric_sign_and_verify_options_description, the verification function is named is**Get**AsymmetricSignAndVer**fy**[sic]OptionsValid.

src/mococrw/private/parsingUtils.h
```diff
@@ -0,0 +1,42 @@
...
+#include <mococrw/basic_constraints.h>
+
+#include <boost/property_tree/ptree.hpp>
+#include <boost/program_options.hpp>
+
+namespace pt = boost::property_tree;
+namespace po = boost::program_options;
+namespace mococrw {
+
+DigestTypes getDigestType(std::string digestString);
```
neverpanic, line 35: Please add documentation comments to these functions.

src/mococrw/private/parsingUtils.h
```diff
@@ -0,0 +1,42 @@
...
+#include <mococrw/openssl_wrap.h>
+#include <mococrw/key_usage.h>
+#include <mococrw/distinguished_name.h>
+#include <mococrw/sign_params.h>
+#include <mococrw/basic_constraints.h>
+
+#include <boost/property_tree/ptree.hpp>
+#include <boost/program_options.hpp>
+
+namespace pt = boost::property_tree;
```
neverpanic, line 31: It is very uncommon for headers to define namespace aliases, since – as a user – you cannot get rid of these aliases once they are defined. Please remove these and but the full signature into the methods.

src/mococrw/private/parsingUtils.h
```diff
@@ -0,0 +1,42 @@
...
+namespace po = boost::program_options;
+namespace mococrw {
+
+DigestTypes getDigestType(std::string digestString);
+openssl::ellipticCurveNid getEllipticCurveNid(const std::string &curveString, bool &success);
```
neverpanic, line 36: This API looks weird. Why don't we just return a `boost::optional<openssl::ellipticCurveNid>` rather than providing a `bool &success` output argument?

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+if(BUILD_EXAMPLES AND BUILD_TESTING)
+    set(EXAMPLE_BINARY_DIR ${CMAKE_BINARY_DIR}/examples)
+    add_test(
+	NAME "ExampleTesting"
+	COMMAND sh -c "MOCOCRW_EXAMPLE_BINARY_PATH=${EXAMPLE_BINARY_DIR} MOCOCRW_SRC_DIR=${EXAMPLE_DIR} /usr/bin/python3 -m unittest"
```
neverpanic, line 5: This is subject to word splitting if you pass it to `sh -c`, so you should either quote or pass things as variables:

```suggestion
	COMMAND sh -c "MOCOCRW_EXAMPLE_BINARY_PATH='${EXAMPLE_BINARY_DIR}' MOCOCRW_SRC_DIR='${EXAMPLE_DIR}' /usr/bin/python3 -m unittest"
```

or

```suggestion
	COMMAND sh -c "env MOCOCRW_EXAMPLE_BINARY_PATH=$1 MOCOCRW_SRC_DIR=$2 /usr/bin/python3 -m unittest" -- "${EXAMPLE_BINARY_DIR}" "${EXAMPLE_DIR}"
```

Also, why do you hardcode `/usr/bin/python3` instead of relying on `$PATH` to locate the correct python3?

tests/examples/common.py
```diff
@@ -0,0 +1,99 @@
...
+    return current_working_dir, new_working_dir
+
+
+def combine_all_options(all_options, option_list):
+    """
+    This function combines the strings in the first string list with the elements of all other
+    string lists in all_options and adds the result to all_options.
+    e.g.
+    all_options = ""
+    option_list = [["hello"],["world"]
```
neverpanic, line 50: Did you mean:
```suggestion
    option_list = [["hello"],["world"]]
```

Also, this function doesn't work for me as described:

```
In [1]: def combine_all_options(all_options, option_list):
   ...:     first_options_list = option_list[0]
   ...:     add_options = [add_opt + opt for add_opt in first_options_list for opt in all_options]
   ...:     all_options += add_options
   ...:     if len(option_list) == 1:
   ...:         return all_options
   ...:     return combine_all_options(all_options, option_list[1:])
   ...:

In [2]: combine_all_options("", [["hello"], ["world"]])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-aa0b89e2c631> in <module>()
----> 1 combine_all_options("", [["hello"], ["world"]])

<ipython-input-1-b76d49b2a128> in combine_all_options(all_options, option_list)
      2     first_options_list = option_list[0]
      3     add_options = [add_opt + opt for add_opt in first_options_list for opt in all_options]
----> 4     all_options += add_options
      5     if len(option_list) == 1:
      6         return all_options

TypeError: can only concatenate str (not "list") to str
```

I'm assuming you wanted `all_options` to be `[""]` initially, but even then, your function does not behave as you described it:

```
In [5]: combine_all_options([""], [["hello", "world"]])
Out[5]: ['', 'hello', 'world']
```

Can we not get rid of this entire function somehow? I'd really like to spend my time on reviewing MoCOCrW example code, not code that tests a command line interface for MoCOCrW crypto functionality.

tests/examples/common.py
```diff
@@ -0,0 +1,99 @@
...
+    "" is prepended even if options_dict is empty.
+    """
+    if not options_dict:
+        # Returning an empty element
+        return [""]
+    option_list = []
+    for key, values in options_dict.items():
+        key_options = []
+        for value in values:
+            key_options.append("--{} {} ".format(key, value))
```
neverpanic, line 96: I'm not a huge fan of creating strings of arguments that then have to be passed through a shell to be split into arguments again. Could we not create lists of arguments, i.e. the result should contain `["--blub", "first", "--ship", "sinking"]`?

I'm assuming you're passing all of this to the subprocess module at some point to execute a binary.

examples/ecies_example.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <mococrw/ecies.h>
+#include <mococrw/key.h>
+
+#include <boost/program_options.hpp>
+
+using namespace mococrw;
+namespace po = boost::program_options;
```
neverpanic, line 9: Please do not use namespaces abbreviations in header files.

examples/ecies_example.h
```diff
@@ -0,0 +1,33 @@
...
+    std::shared_ptr<AsymmetricPublicKey> pubKey;
+    std::shared_ptr<X509Certificate> cert;
+    std::shared_ptr<ECCSpec> eccSpec;
+    std::vector<uint8_t> macValue;
+    openssl::EllipticCurvePointConversionForm ecForm;
+    std::shared_ptr<KeyDerivationFunction> kdfFunc = nullptr;
+    std::function<std::unique_ptr<MessageAuthenticationCode>(const std::vector<uint8_t>&)> macFunc = nullptr;
+    size_t macKeySize;
+    std::vector<uint8_t> data;
+    std::shared_ptr<po::variables_map> vm;
```
neverpanic, line 23: I thought you wanted to make the example independent of the command line variable parsing? Shouldn't this not be necessary then?

examples/ecies_example.h
```diff
@@ -0,0 +1,33 @@
...
+
+using namespace mococrw;
+namespace po = boost::program_options;
+
+struct EciesData{
+    std::shared_ptr<AsymmetricPublicKey> ephKey;
+    std::shared_ptr<AsymmetricPrivateKey> privKey;
+    std::shared_ptr<AsymmetricPublicKey> pubKey;
+    std::shared_ptr<X509Certificate> cert;
+    std::shared_ptr<ECCSpec> eccSpec;
```
neverpanic, line 16: All these objects are lightweight wrappers around a `std::shared_ptr<>` as far as I know. As a consequence, there's little use of wrapping them in another shared pointer, you can just copy them by value instead. That will also simplify some of the code you use above, like `encBuilder.buildEncryptionCtx(*eciesData.pubKey.get());`

neverpanic (changes requested):

neverpanic:
See #82.

PR #85, opened 2020-08-10
tests: Avoid random failure in AES-GCM test

neverpanic:
Encrypting a one-byte plaintext with different IVs and keys using AES-GCM has a chance of (1/255)^2 to produce the same output byte, which occasionally causes this test to fail. This is not a problem for our block cipher tests, since they always produce outputs that have a length that's a multiple of the block size.

Since our test does encryption with 0-byte, 1-byte and 10-byte strings, filtering size() > 1 is enough to avoid this random failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 
 * Exception message in AESCipher (thrown if key has unexpected length)
 * Install target of documentation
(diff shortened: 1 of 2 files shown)


neverpanic:
Hm, CTR is also a stream cipher, so should also be affected by this. Let me check.

blenk92 (approved):

PR #87, opened 2020-12-04
Remove wrapper and mock for rsa_oaep_label getter

dnnr:
The way this function is called does not match the interface
offered by OpenSSL. It expects a pointer-pointer to unsigned char as
output argument, but we pass just a pointer. This probably went
unnoticed in the build due to SFINAE.

The wrappers for the function are not used anywhere in MoCOCrW, and are
also very unlikely to be needed by any of our users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 ## Changed
 
+* Remove wrapper `openssl::_EVP_PKEY_CTX_get_rsa_oaep_label`. This is
+  technically an ABI break, but since the wrappers are not considered part of
+  the public API, we do not bump the SOVERSION for this.
+
 ## Fixed
 
 ## Added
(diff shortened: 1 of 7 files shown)


dnnr:
I'm in the process of verifying internally that this wrapper has not been used anywhere internally.

neverpanic (approved):

dnnr:
Internal checks have all passed, feel free to merge.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 
 ## Changed
 
+* Remove wrapper `openssl::_EVP_PKEY_CTX_get_rsa_oaep_label`
```
neverpanic: As discussed, let's document here why we're not bumping the SOVERSION for this, even though this is technically an ABI break.

neverpanic (changes requested):

CHANGELOG.md
dnnr: Done.

neverpanic (approved):

PR #88, opened 2020-12-07
*PublicKeyCtx: Sig validation error message

garw:
In case that a signature is invalid, we just forward to error
message of the OpenSSLException generation by _EVP_PKEY_verify
to the MoCOCrWException. As an invalid signature is not really
a failure of the OpenSSL call, the OpenSSLException error message
is just error:00000000:lib(0):func(0):reason(0): 0.
This is very confusing and gives no indication that it actually
was the signature validation that failed.

This patch amends the error message and adds the part that signature
validation failed to the what() message of the MoCOCrWException
that is thrown.

Change-Id: Ibbe3e001fa554af9462d18a952d893d121c0e252

### DIFF
--- a/src/asymmetric_crypto_ctx.cpp
+++ b/src/asymmetric_crypto_ctx.cpp
@@ -372,7 +372,9 @@ namespace mococrw {
                                  messageDigest.size());
             }
             catch (const OpenSSLException &e) {
-                throw MoCOCrWException(e.what());
+                throw MoCOCrWException((boost::format{"Signature validation failed. "
+                                                      "OpenSSL info: %1%"}
+                                                    % e.what()).str());
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/asymmetric_crypto_ctx.cpp
```diff
@@ -372,7 +372,9 @@ namespace mococrw {
                                  messageDigest.size());
             }
             catch (const OpenSSLException &e) {
-                throw MoCOCrWException(e.what());
+                throw MoCOCrWException((boost::format{"Signature validation failed. "
+                                                     "OpenSSL info: %1%"}
```
neverpanic: Increase indentation by one space to line up below the string in the line before.

neverpanic (commented):
Nitpick for formatting, content-wise this lg2m.

neverpanic (approved):

blenk92 (approved):

mbehr1 (commented):
:-) thx!

PR #89, opened 2020-12-08
Add support for AES-CMAC according to RFC 4493

doak:
This extends the existing generic `MessageAuthenticationCode` interface
with a second implementation besides HMAC. Given that the existing
naming suggests it this way, both implementations reside in `mac.cpp`
for now, but could be separated in future work.

The public interfaces are only extended, i.e. this change retains
API/ABI compatibility.

### DIFF
--- a/src/mac.cpp
+++ b/src/mac.cpp
@@ -90,7 +90,7 @@ public:
     }
 
 private:
-    openssl::SSL_HMAC_CTX_SharedPtr _ctx = nullptr;
+    openssl::SSL_HMAC_CTX_Ptr _ctx = nullptr;
     bool _isFinished = false;
     std::vector<uint8_t> _result;
 };
@@ -105,10 +105,7 @@ HMAC::~HMAC() = default;
 
 HMAC::HMAC(HMAC&& other) = default;
 
-HMAC& HMAC::operator=(HMAC&& other) {
-    this->_impl = std::move(other._impl);
-    return *this;
-}
+HMAC& HMAC::operator=(HMAC&& other) = default;
 
 void HMAC::update(const std::vector<uint8_t> &message)
 {
--- a/src/mac.cpp
+++ b/src/mac.cpp
@@ -16,17 +16,22 @@
  * limitations under the License.
  * #L%
  */
+
 #include "mococrw/mac.h"
 #include "mococrw/error.h"
 #include "mococrw/openssl_wrap.h"
 
-#include <openssl/hmac.h>
+#include <boost/format.hpp>
 #include <openssl/crypto.h>
+#include <openssl/hmac.h>
 
 namespace mococrw
 {
 
 MessageAuthenticationCode::~MessageAuthenticationCode() = default;
+
+
+/* HMAC */
 
 class HMAC::Impl
 {
@@ -90,6 +95,7 @@ private:
     std::vector<uint8_t> _result;
 };
 
+
 HMAC::HMAC(mococrw::openssl::DigestTypes hashFunction, const std::vector<uint8_t> &key)
 {
     _impl = std::make_unique<HMAC::Impl>(hashFunction, key);
@@ -117,6 +123,101 @@ std::vector<uint8_t> HMAC::finish()
 void HMAC::verify(const std::vector<uint8_t> &hmacValue)
 {
     _impl->verify(hmacValue);
+}
+
+
+/* CMAC */
+
+class CMAC::Impl
+{
+public:
+    Impl(openssl::CmacCipherTypes cipherType, const std::vector<uint8_t> &key)
+    {
+        const EVP_CIPHER* cipher = openssl::_getCipherPtrFromCmacCipherType(cipherType);
+
+        size_t expectedKeySize = openssl::_EVP_CIPHER_key_length(cipher);
+        if (key.size() != expectedKeySize) {
+            auto cipherName = openssl::_EVP_CIPHER_name(cipher);
+            auto formatter = boost::format("Invalid key size for %s: Expected %d bytes but got key with %d bytes.");
+            formatter % cipherName % expectedKeySize % key.size(); 
+            throw MoCOCrWException(formatter.str());
+        }
+
+        _ctx = openssl::_CMAC_CTX_new();
+        openssl::_CMAC_Init(_ctx.get(), key, cipher, nullptr);
+    }
+
+    ~Impl() = default;
+
+    Impl(Impl&&) = default;
+
+    void update(const std::vector<uint8_t> &message)
+    {
+        if (_isFinished) {
+            throw MoCOCrWException("update() can't be called after finish()");
+        }
+        openssl::_CMAC_Update(_ctx.get(), message);
+    }
+
+    std::vector<uint8_t> finish()
+    {
+        if (_isFinished) {
+            throw MoCOCrWException("finish() can't be called twice.");
+        }
+
+        _result = openssl::_CMAC_Final(_ctx.get());
+
+        _isFinished = true;
+
+        return _result;
+    }
+
+    void verify(const std::vector<uint8_t> &cmacValue)
+    {
+        if (!_isFinished) {
+            finish();
+        }
+
+        if (cmacValue.size() != _result.size()) {
+            throw MoCOCrWException("CMAC verification failed. Length differs.");
+        }
+
+        if (CRYPTO_memcmp(cmacValue.data(), _result.data(), cmacValue.size())) {
+            throw MoCOCrWException("CMAC verification failed. Calculated value: " + utility::toHex(_result)
+                                   + ". Received value: " + utility::toHex(cmacValue));
+        }
+    }
+
+private:
+    openssl::SSL_CMAC_CTX_Ptr _ctx = nullptr;
+    bool _isFinished = false;
+    std::vector<uint8_t> _result;
+};
+
+
+CMAC::CMAC(mococrw::openssl::CmacCipherTypes cipherType, const std::vector<uint8_t> &key)
+    : _impl(std::make_unique<CMAC::Impl>(cipherType, key))
+{}
+
+CMAC::~CMAC() = default;
+
+CMAC::CMAC(CMAC&& other) = default;
+
+CMAC& CMAC::operator=(CMAC&& other) = default;
+
+void CMAC::update(const std::vector<uint8_t> &message)
+{
+    _impl->update(message);
+}
+
+std::vector<uint8_t> CMAC::finish()
+{
+    return _impl->finish();
+}
+
+void CMAC::verify(const std::vector<uint8_t> &cmacValue)
+{
+    _impl->verify(cmacValue);
 }
 
 }
--- a/src/mococrw/mac.h
+++ b/src/mococrw/mac.h
@@ -25,8 +25,8 @@ namespace mococrw
 {
 
 /**
- * @brief The MessageAuthenticationCode class is the abstract base class for the different implementation of
- * message authentication codes.
+ * @brief The MessageAuthenticationCode class is the abstract base class for the different
+ * implementation of message authentication codes
  */
 class MessageAuthenticationCode {
 public:
@@ -37,21 +37,23 @@ public:
     virtual ~MessageAuthenticationCode();
 
     /**
-     * Adds the message to the hash. This function may be invoked multiple times.
+     * @brief Adds the message to the MAC
+     *
+     * This function may be invoked multiple times.
      * For finishing the calculation of the MAC invoke finish().
      *
-     * @param message chunk of data used for MAC.
-     * @throws MoCOCrWException if this function is invoked after finish was called.
+     * @param message chunk of data used for MAC
+     * @throws MoCOCrWException if this function is invoked after finish was called
      */
     virtual void update(const std::vector<uint8_t>& message) = 0;
 
     /**
-     * @brief Finalize the MAC.
+     * @brief Finalize the MAC
      *
      * This function calculates the message authentication code.
      *
-     * @throws MoCOCrWException if this function is invoked twice.
-     * @return The calculated message authentication code.
+     * @throws MoCOCrWException if this function is invoked twice
+     * @return The calculated message authentication code
      */
     virtual std::vector<uint8_t> finish() = 0;
 
@@ -63,7 +65,8 @@ public:
      * The length of the calculated value and the given value has to be the same!
      *
      * The comparison happens in constant time.
-     * @throws MoCOCrWException if the verification fails because the values or their lengths differ.
+     *
+     * @throws MoCOCrWException if the verification fails because the values or their lengths differ
      * @param macValue The value which shall be compared to the calculated value
      */
     virtual void verify(const std::vector<uint8_t> &macValue) = 0;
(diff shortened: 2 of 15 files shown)


src/mococrw/mac.h
```diff
@@ -145,4 +145,80 @@ class HMAC : public MessageAuthenticationCode {
     std::unique_ptr<Impl> _impl;
 };
 
+class CMAC : public MessageAuthenticationCode {
+public:
+    /**
+     * @brief Constructor
+     * @param hashFunction The hash function which shall be used
```
```suggestion
     * @param cipherType The type of encryption function which shall be used
```

src/mococrw/mac.h
```diff
@@ -145,4 +145,80 @@ class HMAC : public MessageAuthenticationCode {
...
+     */
+    CMAC(mococrw::openssl::CmacCipherTypes cipherType, const std::vector<uint8_t> &key);
+
+    /**
+     * @brief destructor
+     */
+    ~CMAC();
+
+    /**
+     * Adds the chunk of data to the hash. H(i-key || message1 [ || message2])
```
```suggestion
     * Adds the chunk of data to the MAC.
```

I believe `H(i-key || message1 [ ||  message2 ])` is the formula for HMAC, is it not?

Also, hashing isn't involved in AES-CMAC, so let's not use the word "hash".

src/mococrw/mac.h
```diff
@@ -145,4 +145,80 @@ class HMAC : public MessageAuthenticationCode {
...
+     * @throws MoCOCrWException if this function is invoked after finish was called.
+     */
+    void update(const std::vector<uint8_t>& message) override;
+
+
+    /**
+     * @brief Finalize the CMAC.
+     *
+     * This function calculates the message authentication code.
+     * H(o-key || H(i-key || message [ || message]))
```
neverpanic: This also applies here, this is how HMAC is computed.

src/mococrw/openssl_wrap.h
```diff
@@ -45,10 +45,20 @@ enum class DigestTypes {
     NONE = std::numeric_limits<int>::max()
 };
 
+/**
+ * Enum for the cipher types that can be used for CMAC computations.
+ */
+enum class CmacCipherTypes {
+    AES_CBC_128,
+    AES_CBC_256,
+    NONE = std::numeric_limits<int>::max()
```
neverpanic: I know that should have happened already, but can you please `#include <limits>` for `std::numeric_limits`?

src/openssl_wrap.cpp
```diff
@@ -1386,6 +1393,49 @@ SSL_HMAC_CTX_Ptr _HMAC_CTX_new()
...
+}
+
+std::vector<uint8_t> _CMAC_Final(CMAC_CTX *ctx)
+{
+    std::vector<uint8_t> cmac(EVP_MAX_BLOCK_LENGTH);
+    size_t length = 0;
+    OpensslCallIsOne::callChecked(
+            lib::OpenSSLLib::SSL_CMAC_Final, ctx,
+            cmac.data(), &length);
+    assert(length <= cmac.size());
```
neverpanic, line 1432: Can we statically guarantee that this will never happen? Otherwise it may be better to call `CMAC_Final` twice, once with a `nullptr` as buffer to obtain the required length, and then with a buffer of the required size.

tests/unit/test_cmac.cpp
```diff
@@ -0,0 +1,244 @@
+/*
```
neverpanic, line 1: Should we also test that invoking `update()` multiple times works as expected?

neverpanic (changes requested):
A couple of comments.

src/openssl_wrap.cpp
doak: There is even an API available (`EVP_CIPHER_CTX_block_size()`) to get the block size of a specific cipher.
But yes, MAC size is specified as the block size and `EVP_MAX_BLOCK_LENGTH` represents the biggest block size of any cipher supported by OpenSSL. Looking into the OpenSSL source of `CMAC_Final()` also proves that this should work.

Do you prefer to use `EVP_CIPHER_CTX_block_size()`?

tests/unit/test_cmac.cpp
doak: I would not expect any issues here, but sure, why not.
I will also note that on the list of potentiall improvements for the HMAC implementation.

src/mococrw/mac.h
doak: Yes, sorry. I did miss the comments of this header :(

src/mococrw/openssl_wrap.h
doak: Yep.
On the other hand nothing uses this at all. Do you mind just removing `NONE` and, if your linter on CI is good enough, perhaps also remove the `default` case in the mapper `_getCipherPtrFromCmacCipherType()`?
(I have just used that construct because it was done that way for HMAC.)

src/mococrw/mac.h
doak: Nice! I have just saw that the CI failed because of the undocumented parameter :)

doak:
Regarding doxygen comments: I just saw that a there is a mixture of Doxygen comments with and without trailing full stop (copied over from HMAC). I will remove the trailing full stops as well.

src/mococrw/openssl_wrap.h
neverpanic: The problem with doing that is that it would be inconsistent to not have it for CmacCipherTypes but keep it for HMAC, but removing it from HMAC is an ABI and API change. Let's just keep it.

src/openssl_wrap.cpp
neverpanic: No, this is fine as-is then.

src/mococrw/openssl_wrap.h
doak: Done.

neverpanic (approved):
lg2m now, but I'd like to wait for a second review.

src/mococrw/mac.h
doak: Done.

tests/unit/test_cmac.cpp
doak: Done.

doak:
I took the opportunity to harmonise Doxygen comments throughout `mococrw/mac.h` (including existing ones => dedicated commit) as well as to split up the test of the control flow into several test cases (only for CMAC).

neverpanic (approved):

neverpanic:
Unfortunately some of our internal systems use older copies of gtest, which seems to cause this issue:
```
/usr/bin/c++   -I/src/src -isystem /usr/src/gmock/include -isystem /usr/src/gmock -isystem /usr/src/googletest/include -isystem /usr/src/googletest -fPIE   -Wall -Wextra -pedantic -fstack-protector-strong -Wfatal-errors -pthread -std=gnu++14 -MD -MT tests/unit/CMakeFiles/cmactests.dir/test_cmac.cpp.o -MF tests/unit/CMakeFiles/cmactests.dir/test_cmac.cpp.o.d -o tests/unit/CMakeFiles/cmactests.dir/test_cmac.cpp.o -c /src/tests/unit/test_cmac.cpp
/src/tests/unit/test_cmac.cpp:155:25: error: expected constructor, destructor, or type conversion before '(' token
 INSTANTIATE_TEST_SUITE_P(
                         ^
compilation terminated due to -Wfatal-errors.
```

Additionally, it seems this doesn't compile with clang:
```
/usr/lib/ccache/clang++   -I/src/src -isystem /usr/src/gmock/include -isystem /usr/src/gmock -isystem /usr/src/googletest/include -isystem /usr/src/googletest -fsanitize=address -fno-omit-frame-pointer -fsanitize-address-use-after-scope -g -O1 -fPIE   -Wall -Wextra -pedantic -fstack-protector-strong -Wfatal-errors -pthread -std=gnu++14 -MD -MT tests/unit/CMakeFiles/cmactests.dir/test_cmac.cpp.o -MF tests/unit/CMakeFiles/cmactests.dir/test_cmac.cpp.o.d -o tests/unit/CMakeFiles/cmactests.dir/test_cmac.cpp.o -c /src/tests/unit/test_cmac.cpp
/src/tests/unit/test_cmac.cpp:156:9: fatal error: unknown type name 'cmac'
        cmac,
        ^
```

This may be the same issue with gtest, since the location is similar, and the difference may just be in the error messages printed by the compilers.

doak:
@neverpanic, I am sorry. I did not take into account that you could use an older version of _gtest_. It is marked as deprecated for my local version.
Fixed that. Yes: the second issue is very likley caused by the first.

neverpanic (approved):
Thanks that solves the issue. We have a centrally maintained copy of gtest that's a bit dated.

src/mac.cpp
```diff
@@ -119,4 +124,98 @@ void HMAC::verify(const std::vector<uint8_t> &hmacValue)
...
+/* CMAC */
+
+class CMAC::Impl
+{
+public:
+    Impl(openssl::CmacCipherTypes cipherType, const std::vector<uint8_t> &key)
+    {
+        const EVP_CIPHER* cipher = openssl::_getCipherPtrFromCmacCipherType(cipherType);
+
+        if (key.empty()) {
```
garw: The AES Crypto implementation does a test on the key size and provides a MoCOCrW exception with an easy to understand error message  if the sizes don't add up. We should probably do the same here.

src/mac.cpp
```diff
@@ -119,4 +124,98 @@ void HMAC::verify(const std::vector<uint8_t> &hmacValue)
...
+        }
+
+        if (CRYPTO_memcmp(cmacValue.data(), _result.data(), cmacValue.size())) {
+            throw MoCOCrWException("CMAC verification failed. Calculated value: " + utility::toHex(_result)
+                                   + ". Received value: " + utility::toHex(cmacValue));
+        }
+    }
+
+private:
+    openssl::SSL_CMAC_CTX_SharedPtr _ctx = nullptr;
```
garw: Any particular reasons this is a shared_ptr instead of a unique_ptr?

src/mac.cpp
```diff
@@ -119,4 +124,98 @@ void HMAC::verify(const std::vector<uint8_t> &hmacValue)
...
+
+CMAC::CMAC(mococrw::openssl::CmacCipherTypes cipherType, const std::vector<uint8_t> &key)
+    : _impl(std::make_unique<CMAC::Impl>(cipherType, key))
+{}
+
+CMAC::~CMAC() = default;
+
+CMAC::CMAC(CMAC&& other) = default;
+
+CMAC& CMAC::operator=(CMAC&& other) {
```
garw: isn't this the default as well?

src/mococrw/mac.h
```diff
@@ -134,13 +139,88 @@ class HMAC : public MessageAuthenticationCode {
...
+    std::unique_ptr<Impl> _impl;
+};
+
+class CMAC : public MessageAuthenticationCode {
+public:
+    /**
+     * @brief Constructor
+     * @param cipherType the type of encryption function which shall be used
+     * @param key the key used for CMAC
+     */
```
garw, line 159: In the current implementation, this throws if cipherType is NONE. I wonder if we should really add this NONE type. This way, we could syntactically ensure that this call will never fail due to an invalid cipher.

The documentation should also reflect that it throws a MoCOCrWException if the key size is wrong.

src/mococrw/openssl_wrap.h
garw: Honestly, I kind of agree that not having NONE is better design. I fear we have this implict optional build-in at other places for digests but I don't think that its a good idea to continue this. We should rather reconsider the NONE on DigestTypes in the future…

tests/unit/test_cmac.cpp
```diff
@@ -0,0 +1,267 @@
...
+TEST_P(CalculateCmacValues, verifyFailsForIncorrectCmac)
+{
+    auto testdata = GetParam();
+
+    auto cmacCalculator = CMAC(testdata.cipherType, utility::fromHex(testdata.key));
+    cmacCalculator.update(utility::fromHex(testdata.message));
+    cmacCalculator.finish();
+
+    auto incorrectCmac = utility::fromHex(testdata.expectedCmac);
+    incorrectCmac[0]++;
```
garw: Strictly speaking this opens the door to undefined behavior. Maybe we want to be a good example and do incorrectCmac[0] ^= 0x1;

garw (changes requested):
Looks mostly good. A few knitpicks regarding some design issues.

src/mac.cpp
doak: No hard rationale. It is done the same within the HMAC implementation and, since raw pointers are used for the next level of abstraction (within `CMAC::Impl`), I thought a `shared_ptr` reflect this fact the best.

Btw., my first intention was to extract the raw pointer as late as possible just for the calls of the `SSL_*` methods (i.e. within call of `*::callChecked()` in `openssl_wrap.cpp`), but I decided against it because it is used differently all over the place in the existing code. (I am not sure yet how much we are allowed to sacrifice consistency in favour of cleaner code.)

tests/unit/test_cmac.cpp
doak: I appreciate that.
Just to sort it out correctly for me: This particular case is well defined (unsigned integral type) but to avoid potentially problems, such a pattern should be avoided if possible, right?

src/mac.cpp
doak: No hard rationale. It is done the same within the HMAC implementation and, since raw pointers are used for the next level of abstraction (within `CMAC::Impl`), I thought a `shared_ptr` reflect this fact the best.

Btw., my first intention was to extract the raw pointer as late as possible just for the calls of the `SSL_*` methods (i.e. within call of `*::callChecked()` in `openssl_wrap.cpp`), but I decided against it because it is used differently all over the place in the existing code. (I am not sure yet how much we are allowed to sacrifice consistency in favour of cleaner code.)

src/mococrw/mac.h
doak: I just had another look into the code including OpenSSL itself. In fact `openssl::DigestTypes::NONE` is used e.g. for `EdDSASignaturePublicKeyCtx::verifyMessage()` to allow OpenSLL to select a default digest by its own.
Regardless that `NONE` is quite a bad name for his behaviour, there seems to be a valid use case for this.

Such a behaviour does not exist for AES-CMAC (in current OpenSSL implementation at least).
I guess you still vote for removing `openssl::CmacCipherTypes::None`, right?

src/mac.cpp
garw: Then, I'd suggest to move to unique_ptr here. This is no public interface. We even change it to unique_ptr in the hmac without impacting anyone due to the impl pattern. Can you just change it in both places?

src/mococrw/mac.h
garw: Yes. Let's try to (slowly but constantly) move towards a better code-base.

tests/unit/test_cmac.cpp
garw: A true, I didn't check that its unsigned. So you're right. Maybe we want to change this anyways?

doak: Sure.

src/mococrw/openssl_wrap.h
doak: Removed `None` from `CmacCipherTypes`.

src/mac.cpp
doak: Done.

src/mac.cpp
doak: Done.

src/mac.cpp
doak: Done.

src/mococrw/mac.h
doak: Exception with message is thrown. Documentation and test case updated.

tests/unit/test_cmac.cpp
doak: Done.

garw (approved):

neverpanic (approved):

PR #90, opened 2020-12-22
Make find_library for ASan work on Ubuntu 20.04

dnnr:
This plain find_library(ASAN_LIBRARY asan) somehow fails to find libasan
on Ubuntu 20.04 (and likely other Debian-based systems). Mentioning
libasan.so.5 as a library file name candidate fixes this. libasan.so.6
is there for future-compatibility.

### DIFF
--- a/cmake/AddressSanitizerTarget.cmake
+++ b/cmake/AddressSanitizerTarget.cmake
@@ -95,7 +95,7 @@ mark_as_advanced(
 unset(_ASAN_FLAGS)
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL GNU)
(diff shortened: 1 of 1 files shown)


neverpanic (changes requested):
A bit weird that there doesn't seem to be a corresponding `libasan-dev` package that provides the unversioned symlink, but such, I guess, is life.

I've also noticed that there's `libasan.so.6` now; should we also include that for future-compatibility?

dnnr:
It's weird indeed. Good point about future-compat, so I've added libasan.so.6.

_Edit: Updated the commit message now, too._

neverpanic (approved):

PR #91, opened 2020-12-22
Do not rely on global openssl.cnf in CSR test

dnnr:
This invocation of the openssl tool tries to load the global
configuration file by default. Having the test be influenced by this
external artifact is not ideal. More importantly, it also fails if the
file does not exist. Removing this dependency will allow us to use the
much smaller "install_sw" target (no docs, no config) when adding
OpenSSL to the build container.

### DIFF
--- a/tests/unit/test_csr.cpp
+++ b/tests/unit/test_csr.cpp
@@ -265,7 +265,7 @@ TEST_P(CSRTest, testCsrSignatureDigest)
         of << csr.toPEM();
         of.close();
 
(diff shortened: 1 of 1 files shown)


garw (approved):
I'm not sure if this is the only place where we have this issue. I also wouldn't be too surprised if other places actually rely on an openssl.cnf being present but it is not needed here. Let's give this approach a try

dnnr:
For what it's worth, it's been the only place that actually behaved differently when the file was missing. So I'm cautiously optimistic that that's it.

neverpanic (approved):

PR #92, opened 2020-12-22
Replace Travis with new GitHub Actions CI workflow

dnnr:
The new pipeline uses an ad-hoc build of the Docker image providing the
building and testing environment. This removes the reliance on an
external Docker Hub account. A third-party layer caching component
helps to reduce the time spent on creating the image.

In addition to the previous pipeline, we now employ a matrix build to
check all combinations of clang/gcc and ASan enabled/disabled.


dnnr:
This PR is technically ready, but note that it depends on #90 and #91 to be merged before the GHA build will actually run green.

As another prerequisite, someone with the appropriate privileges will need to swap out Travis with GHA in the branch protection rules for `openssl1.1`. This might involve a chicken-and-egg problem, since GHA does not seem to want to run my workflow in this PR, maybe because I'm not a project owner?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: satackey/action-docker-layer-caching@v0.0.11
+
+      - name: "Build Docker image"
+        run: |
+          docker build -t buildenv - < docker-build-env/Dockerfile
+
+      - name: "Build MoCOCrW"
+        run: |
+          mkdir build
+          docker run --rm -u $UID -v $PWD:/src:rw -w /src/build buildenv bash -c \
```
neverpanic: My endless shellscript paranoia would prefer:

```suggestion
          docker run --rm -u "$UID" -v "$PWD:/src:rw" -w /src/build buildenv bash -c \
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+              -DCMAKE_CXX_COMPILER=${{ matrix.compiler }} \
+              -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+              -DBUILD_TESTING=True \
+              -GNinja \
+              /src \
+            && ninja'
+
+      - name: "Run tests"
+        run: |
+          docker run --rm -u $UID -v $PWD:/src:rw -w /src/build buildenv bash -c 'ctest -j $(nproc)'
```
neverpanic:
```suggestion
          docker run --rm -u "$UID" -v "$PWD:/src:rw" -w /src/build buildenv bash -c 'ctest -j $(nproc)'
```

docker-build-env/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+        ninja-build \
+        pkg-config \
+        wget \
+        && rm -rf /var/lib/apt/lists/*
+
+# Verify that the OpenSSL dev package has not been installed inadvertently
+RUN ! dpkg -s libssl-dev
+
+# Build and install specific OpenSSL version
+RUN wget -O openssl.tar.gz https://github.com/openssl/openssl/archive/OpenSSL_1_1_1-stable.tar.gz \
```
neverpanic: Should we really use an unversioned tarball here? I know it reduces our maintenance effort when security fixes are merged, but I guess our caching layer wouldn't rebuild the container anyway unless we touch the Dockerfile, so we wouldn't get the newer releases without manual intervention anyway?

Use of a specific release would also allow us to verify a checksum here. @garw, any opinions?

neverpanic (changes requested):

docker-build-env/Dockerfile
dnnr: I've based this on the [status quo](https://github.com/blenk92/MoCOCrW-build-env/blob/master/Dockerfile), assuming that it's been the preferred way.

I've since discussed it with Max, and it seems unclear whether there was any good reason for doing this in the first place. Is there actually a desire to pin a specific version here and not just use the one provided by Ubuntu 20.04? If so, I agree that pinning it to a repository version and not just to 1.1.1 would make sense. If not, we could just remove this altogether.

.github/workflows/ci.yml
dnnr: :+1: It's not paranoia if they're really after you.

docker-build-env/Dockerfile
neverpanic: https://github.com/blenk92/MoCOCrW-build-env/commit/6af34872fc6faa62005fd0093756546385a2c744 switched from the Ubuntu package to self-installing. I don't see a documented reason for this, but I'm assuming it's because Ubuntu didn't ship 1.1.1 (but 1.1.0) back then.

Maybe we should re-visit this now that focal has 1.1.1.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  push:
+    branches:
+      - openssl1.1
+  pull_request:
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    strategy:
```
neverpanic, line 12: Can we disable fast-failure so that we'll get all results even if one of the jobs fails?

neverpanic (changes requested):

.github/workflows/ci.yml
dnnr: I've kept it like this on purpose, just to be economical with CPU time. But if you prefer it that way, I'll change it. :+1:

docker-build-env/Dockerfile
dnnr: Just to be clear, beware that simply installing libssl-dev yields essentially the same behavior as the current state of the PR: The actual version that ends up being used depends on when the image is built (not recalled from cache). Whether this is a bad thing depends on how much we trust that security fixes will not bring any surprises for MoCOCrW.

In similar scenarios, I've resorted to using Debian with snapshot.debian.org in order to really pin down everything. I'm not aware of an equivalent service for Ubuntu, and more importantly, as you already said, it adds maintenance effort.

neverpanic (approved):
LG2M now, let's wait for @garw to weigh in on the OpenSSL installation topic.

docker-build-env/Dockerfile
garw: I think this was added because some features (I think some SHA3 support) was only available in a newer version than the one that came with ubuntu. In my opinion we can continue in two ways
1) We just take Ubuntu's OpenSSL version.
2) We use a fixed version and bump it every once in a while.

My gut feeling is that 1) is good enough for this purpose here and less maintainence effort. We use the internal-CI anyways additionally to check against the version that we internally use.

dnnr: Makes sense to me. I've replaced it all with `libssl-dev` in the dependencies list.

garw (approved):
lg2m

neverpanic:
Now mandatory for the `openssl1.1` branch.

PR #93, opened 2021-01-04
Replace Travis with new GitHub Actions CI workflow

neverpanic:
The new pipeline uses an ad-hoc build of the Docker image providing the building and testing environment. This removes the reliance on an external Docker Hub account. A third-party layer caching component helps to reduce the time spent on creating the image.

In addition to the previous pipeline, we now employ a matrix build to check all combinations of clang/gcc and ASan enabled/disabled.

See also #92. I'm mirroring this in-repo, hoping that that will make GitHub trigger the action for verification purposes.


neverpanic:
It seems we now no longer need this, since the CI now runs in the other PR as well.

PR #97, opened 2021-05-31
Fix CA problems with DistinguishedName order

### LINKED ISSUE #95: MoCOCrW reorders distinguished name fields when loading a CA cert from PEM
When a CA's `X509Certificate` is loaded from a PEM file and then used to sign a certificate, the Issuer name in the signed certificate ends up with attributes in a different order from the Subject name in the parent CA certificate.

For instance, my CA had:

```Subject: C = PT, ST = Testing, O = Critical TechWorks, OU = Chronos Team, CN = Test-Signing-CA```

while the signed certificate ended up with:

```Issuer: CN = Test-Signing-CA, C = PT, ST = Testing, OU = Chronos Team, O = Critical TechWorks```

This causes the MoCOCrW itself to fail to validate the certificate it just signed (as well as openssl command line), and the sanity check inside function `CertificateAuthority::_signCSR()` throws an `mococrw::MoCOCrWException` with the error:

```
what(): Certificate creation failed: the generated certificate is invalid: unable to get local issuer certificate
```

### PR DESCRIPTION
**Pull request by garw**
The CertificateAuthority functionality in MoCoCrW has the
problem that distinguished names are defined as ordered in
the standard but this is not reflected in the DistinguishedName
objects in the library.
This lead to the problem that the CertificateAuthority objects
can't issue certificates with a CA certificate that uses a different
attribute order in the subject than the libraries default order.

This patch fixes the issue by directly copying the CAs subject
and the subject in the CSRs directly into the issuer and subject
fields of the newly issued certificate.

This fixes #95.

On the way, improve one the const correctness of some of the OpenSSL
methods as OpenSSL has improved since we began developing this library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ## Fixed
 
+* CertificateAuthority now copies the subject of the CA directly into the
+  issuer field of the issued certificate. This resolves problems around
+  different orders of items in the underlying distinguished name. This fixes
+  issue #95. The DistinguishedName object is still not order-aware when loading
+  a DN from OpenSSL. This is to be fixed in a later step.
+
 ## Added
 
 # Release 4.1.0
--- a/tests/unit/test_ca.cpp
+++ b/tests/unit/test_ca.cpp
@@ -445,6 +445,33 @@ TEST_F(CATest, testGetNextSerialNumber)
     EXPECT_EQ(newCaCert.getSerialNumber(), nextSerial);
 }
 
+/* Signing should succeed even if the CA has a subject order that
+ * does not match MoCOCrW's default DN order.
+ * This is a regression test for issue #95
+ */
+TEST_F(CATest, testSigningWithCustomOrderCA)
+{
+    auto orderedSubject = DistinguishedName::CustomOrderBuilder{}
+            .localityName("TestTown")
+            .countryName("DE")
+            .organizationName("MyOrg")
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: tests/unit/test_ca.cpp [hunk 1]
```diff
@@ -445,6 +445,33 @@ TEST_F(CATest, testGetNextSerialNumber)
     EXPECT_EQ(newCaCert.getSerialNumber(), nextSerial);
 }
 
+/* Signing should succeed even if the CA has a subject order that
+ * does not match MOCoCrW's default DN order.
```
neverpanic: Capitalization:

```suggestion
 * does not match MoCOCrW's default DN order.
```

neverpanic (commented):
A small nitpick.

neverpanic (approved):
LG2M.

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ## Fixed
 
+* CertificateAuthority now copies the subject of the CA directly into the
+  issuer field of the issued certificate. This resolves problems around
+  different orders of items in the underlying distinguished name. The
+  DistinguishedName object is still not order-aware when loading a DN
+  from OpenSSL. This is to be fixed in a later step.
```
neverpanic:
```suggestion
  from OpenSSL. This is to be fixed in a later step. (#95)
```

I'd suggest we reference the ticket.

blenk92 (approved):

vkrikun:
Looks good.

PR #98, opened 2021-07-08
Allow to sign past/future certificates

### LINKED ISSUE #96: Can not sign a certificate out of notBefore and notAfter validity.
The "sanity check" in `CertificateAuthority::_signCSR()` does not allow me to sign a certificate with expired validity or a certificate that will be valid in the future, because validation at the moment must succeed at the end of the function. Anything that is not valid at the moment of the signature can not be signed.

miroshko:
within validity bounds of the root certificate

Resolves #96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ All notable changes to this project will be documented in this file.
 
 ## Fixed
 
+## Added
+
+# Release 4.1.1
+
+## Changed
+
+## Fixed
+
 * CertificateAuthority now copies the subject of the CA directly into the
   issuer field of the issued certificate. This resolves problems around
   different orders of items in the underlying distinguished name. This fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ All notable changes to this project will be documented in this file.
   different orders of items in the underlying distinguished name. This fixes
   issue #95. The DistinguishedName object is still not order-aware when loading
   a DN from OpenSSL. This is to be fixed in a later step.
+* X509Certificate::signCSR doesn't validate the certificate at the current
+  system time anymore but at certificate's notBefore and notAfter dates.
+  This fixes issue #96 by allowing to sign past and future certificates but
+  also ensures that the certificate's validity period does not exceed the
+  validity bounds of the issuing certificate.
 
 ## Added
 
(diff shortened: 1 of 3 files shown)


src/x509.cpp
```diff
@@ -211,7 +211,10 @@ void X509Certificate::verify(const std::vector<X509Certificate> &trustStore,
 {
     VerificationContext ctx;
     ctx.addTrustedCertificates(trustStore)
-       .addIntermediateCertificates(intermediateCAs);
+       .addIntermediateCertificates(intermediateCAs)
+       .setVerificationCheckTime(getNotBeforeAsn1());
+    verify(ctx);
+    ctx.setVerificationCheckTime(getNotAfterAsn1() - Asn1Time::Seconds(1));
```
miroshko: still need to clarify this -1 second. according to the name and to the brief glance at the openssl code this shouldn't be needed

src/ca.cpp
```diff
@@ -98,7 +98,13 @@ X509Certificate CertificateAuthority::_signCSR(const CertificateSigningRequest &
     //Sanity check: certificate must be verifiable now
     try {
         auto cert = X509Certificate{std::move(newCertificate)};
-        cert.verify({_rootCert}, {});
+        X509Certificate::VerificationContext ctx;
+        ctx.addTrustedCertificates({_rootCert})
+            .addIntermediateCertificates({})
+            .setVerificationCheckTime(cert.getNotBeforeAsn1());
+        cert.verify(ctx);
+        ctx.setVerificationCheckTime(cert.getNotAfterAsn1() - Asn1Time::Seconds(1));
```
garw, line 108: If we need this, please add a comment due to the asymmetrie with NotBefore. However, I'd also think that it is not necessary.

garw (changes requested):
Clearify the notAfter issue. 
Otherwise this looks good to me. Tests look also good. Wanted to see the CA / cert overlap problem but it is included :)

Please update the release notes to include this (so far unreleased change).

src/ca.cpp
miroshko: I added a comment, that's indeed how OpenSSL checks it

garw (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ All notable changes to this project will be documented in this file.
 
 ## Fixed
 
+## Added
+
+# Release 4.1.1
+
+## Changed
+
```
neverpanic, line 16: I'd drop headlines with no content from the release notes.

neverpanic (commented):
Nitpick, lg2m otherwise.

wusto (approved):

PR #99, opened 2022-03-31
Add examples of certificate validation

istepic:


examples/cert-validation.cpp
```diff
@@ -0,0 +1,112 @@
...
+       .addIntermediateCertificate(intermediateCA_1)
+       .addCertificateRevocationList(intermediateCRL)
+       .enforceSelfSignedRootCertificate();
+
+    try {
+        cert.verify(ctx_4);
+    } catch (const MoCOCrWException &e) {
+        std::cerr << "Failed to verify certificate.\n";
+        std::cerr << e.what() << '\n';
+    }
```
garw, line 109: You/we may want to illustrate the option to do certificate validation for a certain date and time.

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include <iostream>
+#include <chrono>
+#include <thread>
```
garw, line 21: why?

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include <iostream>
+#include <chrono>
```
garw, line 20: why?

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+    } catch (const MoCOCrWException &e) {
+        std::cerr << "Failed to verify end certificate.\n";
+        std::cerr << e.what() << '\n';
+        exit(EXIT_FAILURE);
+    }
+    std::cout << "Success!\n";
+
+    // Verify after end certificate expiry
+    std::cout << "Verifying after end certificate expired...\n";
+    std::this_thread::sleep_for(std::chrono::seconds(6)); // Intermediate CA signs certificate that last 5 seconds
```
garw, line 237: Well, let's try to avoid weird racy not best practice code in the examples. We have seen in particular that this short expiry times are problematic when running in emulators (e.g. ARM). Better make the validity times longer and illustrate time checks with the ability to check at a given point in time.

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+                rootCertDetails,
+                // serial number of the root certificate
+                0,
+                rootCaSelfSignParams);
+}
+
+/*
+* Define CA Class for root CA based on root CA Certificate and keypair with random CA settings
+*/
+CertificateAuthority defineRootCa(const X509Certificate &rootCaCert,
```
garw, line 77: Define seems to not quite represent what is going on here. Maybe rather `configure` or `build`?
Moreover, it would be nice to see an explanation here that MoCoCrW has the `CertificateAuthority` class to provide an interface for CAs to configure how certificates should be issued. + Explanation that they are configured how they issue certificates (as in, which extensions etc should be present in the certificates). They can then be used to sign CSRs. What information is taken from the CSR? What is configured in the CA object?

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+X509Certificate signCertificateWithRootCa(const X509Certificate &signingCert,
+                                          const AsymmetricKeypair &signingKey,
+                                          const CertificateSigningRequest &csr)
+{
+    auto ca = defineRootCa(signingCert, signingKey);
+    try {
+        return ca.signCSR(csr);
+    } catch (const openssl::OpenSSLException &e) {
+        /* low level OpenSSL failure */
+        std::cerr << "Error reading CSR in PEM format. Please validate your CSR." << std::endl;
```
garw, line 116: Why should we see a PEM error here? The CSR is an in-memory object here?

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+                                                const X509Certificate &rootCert)
+{
+    auto intermediateCaCertDetails = DistinguishedName::Builder{}
+            .organizationalUnitName("ezcode")
+            .organizationName("eazz")
+            .countryName("DE")
+            .commonName("Common secondary certificate")
+            .build();
+
+    auto csr = CertificateSigningRequest(intermediateCaCertDetails,
```
garw, line 142: We could use this to explain here that things like the CA status (being a CA or not) are not part of the CSR but are rather configured by the issuer.

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+                                         digestType);
+    auto intermediateCaCert = signCertificateWithRootCa(rootCert, rootCaKey, csr);
+
+    return intermediateCaCert;
+}
+
+/*
+* Define CA Class for intermediate CA based on CA Certificate, keypair and random settings
+*/
+CertificateAuthority defineIntermediateCa(const X509Certificate &intermediateCaCert,
```
garw, line 153: s.a. regarding naming

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+X509Certificate signCertificateWithIntermediateCa(const X509Certificate &signingCert,
+                                                  const AsymmetricKeypair &signingKey,
+                                                  const CertificateSigningRequest &csr)
+{
+    auto ca = defineIntermediateCa(signingCert, signingKey);
+    try {
+        return ca.signCSR(csr);
+    } catch (const openssl::OpenSSLException &e) {
+        /* low level OpenSSL failure */
+        std::cerr << "Error reading CSR in PEM format. Please validate your CSR." << std::endl;
```
garw, line 189: s.a.

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    // Make intermediate CA
+    auto intermediateCaEccKey = AsymmetricKeypair::generateECC();
+    auto intermediateCaCert = createIntermediateCaCertificate(intermediateCaEccKey, rootCaKeypair, rootCaCert);
+
+    // Sign a certificate with intermediate CA
+    auto endCertKeypair = AsymmetricKeypair::generateRSA();
+    auto endCertDetails = DistinguishedName::Builder{}
+            .organizationalUnitName("Car IT")
+            .organizationName("BMW")
```
garw, line 215: For new code we try to avoid BMW references to keep the code more generic.

examples/creating-cert-chain-with-validation.cpp
```diff
@@ -0,0 +1,247 @@
...
+            .build();
+    auto csr = CertificateSigningRequest(endCertDetails,
+                                         endCertKeypair,
+                                         digestType);
+    auto endCert = signCertificateWithIntermediateCa(intermediateCaCert, intermediateCaEccKey, csr);
+
+    // Verify end certificate with intermediate CA's certificate
+    std::cout << "Verifying end certificate...\n";
+    try {
+        endCert.verify({rootCaCert}, {intermediateCaCert});
```
garw, line 227: see below w.r.t. racy-ness

garw (changes requested):
Looks already like a good illustration of the capabilities of our code. Try to be, in general, a bit more verbose in the comments. The purpose of the examples is that a reader unfamiliar with the code can go through them, the functionality is demonstrated in the code while being explained/introduced in the comments.

PR #100, opened 2022-04-05
Add cert validation examples

istepic:


examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+       These artifacts are taken from `tests` folder in the repo
+       for demonstration purposes.
+    */
+
+    /*
+        Basic certificate validation using X509Certificate::verify() against
+        issuing CA
+    */
+    {
+        std::cout << "Validating certificate against an issuing CA\n";
```
garw: we prefer `<< std::endl` over \n as it flushes the buffer.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+            std::cerr << "Failed to validate certificate.\n";
+            std::cerr << "OpenSSL error: " << e.what() << '\n';
+            exit(EXIT_FAILURE);
+        }
+        std::cout << "Certificate successfully validated\n---\n";
+    }
+
+    /*
+        Basic certificate validation using X509Certificate::verify() against
+        issuing certificate chain
```
garw: Maybe be a bit more verbose here and explain the the verify method can take a vector of possible intermediate CAs as second argument. Any of those might be used to construct a valid chain. However, not all of them need to be used.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+        Basic certificate validation using X509Certificate::verify() against
+        issuing CA
+    */
+    {
+        std::cout << "Validating certificate against an issuing CA\n";
+        try {
+            cert.verify({intermediateCA}, {});
+        } catch (const MoCOCrWException &e) {
+            std::cerr << "Failed to validate certificate.\n";
+            std::cerr << "OpenSSL error: " << e.what() << '\n';
```
garw: can be simplified by removing the `;` in the previous line and `std:cerr` in this line.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+    /*
+        Negative case - Validation attempt of the certificate issued by other CA's.
+    */
+    {
+        X509Certificate::VerificationContext ctx;
+        ctx.addTrustedCertificate(rootCA)
+           .addIntermediateCertificate(intermediateCA);
+        std::cout << "Verifying certificate that is issued from other CA's\n";
+        try {
+            cert_from_other_chain.verify(ctx);
```
garw, line 130: We should probably exit with an error after this line as this shouldn't be reachable if everything works as expected.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+        enforceSelfSignedRootCertificate(). In this case intermediate CA
+        certificate is not self-signed so exception is expected.
+    */
+    {
+        X509Certificate::VerificationContext ctx;
+        ctx.addTrustedCertificate(intermediateCA)
+           .enforceSelfSignedRootCertificate();
+        std::cout << "Checking if intermediate certificate is self signed\n";
+        try {
+            rootCA.verify(ctx);
```
garw, line 170: s.a.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+            std::cerr << "Failed as expected.\n";
+            std::cerr << "OpenSSL error: " << e.what() << '\n';
+        }
+        std::cout << "---\n";
+    }
+
+    /*
+        Add CRL to VerificationContext for intermediate CA which has
+        ceritificate stored in variable `cert` revoked (serial number 0x1002).
+        CRL's are added with addCertificateRevocationList() method.
```
garw:
```suggestion
        CRLs are added with addCertificateRevocationList() method.
```

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+        root CA's CRL.
+    */
+    {
+        X509Certificate::VerificationContext ctx;
+        ctx.addTrustedCertificate(rootCA)
+           .addIntermediateCertificate(intermediateCA)
+           .addCertificateRevocationList(intermediateCRL);
+        std::cout << "Checking if certificate is revoked\n";
+        try {
+            cert.verify(ctx);
```
garw, line 193: s.a.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+    {
+        X509Certificate::VerificationContext ctx;
+        ctx.addTrustedCertificate(rootCA)
+           .addIntermediateCertificate(intermediateCA)
+           .addCertificateRevocationList(rootCRL)
+           .enforceCrlsForAllCAs()
+           .enforceSelfSignedRootCertificate();
+        std::cout << "Checking enforcement of CRL's\n";
+        try {
+            cert.verify(ctx);
```
garw, line 215: s.a.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+        } catch (const MoCOCrWException &e) {
+            std::cerr << "Failed as expected.\n";
+            std::cerr << "OpenSSL error: " << e.what() << '\n';
+        }
+        std::cout << "---\n";
+    }
+
+    /*
+        Adding custom time in VerificationContext
```
garw: Maybe be a bit more verbose here. Something like. 

> The VerificationContext also allows the caller to specify a specific time (and date) that should be used for certificate validation instead of the current system time.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,248 @@
...
+        ctx.addTrustedCertificate(rootCA)
+           .addIntermediateCertificate(intermediateCA)
+           .addCertificateRevocationList(rootCRL)
+           .addCertificateRevocationList(intermediateCRL)
+           .enforceCrlsForAllCAs()
+           .enforceSelfSignedRootCertificate()
+           .setVerificationCheckTime(Asn1Time::max());
+        std::cout << "Validating certificate in future when stuff is expired\n";
+        try {
+            cert.verify(ctx);
```
garw, line 240: s.a.

garw (changes requested):
Looks mostly good now. A few remaining knitpicks.

Moreover, maybe make the commit msg a bit more verbose. Take a look at the history to get an impression of the usual style in this project.

examples/cert-validation.cpp
istepic: Why do we care about flushing here?

garw: I don't. It's just common practice as you don't know who might be editing the code later. If someone adds code that may crash below at a later point in time, your buffer output will not be displayed. It is probably not really important here but it simplifies developing and reading of code if common patterns are repeated instead of always deciding individually.

garw (commented):
looks good. Can be merged as soon as our internal CI also confirms.

wusto (approved):

examples/cert-validation.cpp
```diff
@@ -0,0 +1,250 @@
...
+            std::cerr << "Failed to validate certificate. OpenSSL error: " << e.what() << std::endl;
+            exit(EXIT_FAILURE);
+        }
+        std::cout << "Certificate successfully validated\n---" << std::endl;
+    }
+
+    /*
+     *   Basic certificate validation using X509Certificate::verify() against
+     *   issuing certificate chain. X509Certificate::verify() method can accept
+     *   a vector of trusted root CA's or a vector of intermediate CA's where
```
garw:
```suggestion
     *   a vector of trusted root CAs or a vector of intermediate CAs where
```
small typo

examples/cert-validation.cpp
```diff
@@ -0,0 +1,250 @@
...
+            cert.verify(ctx);
+        } catch (const MoCOCrWException &e) {
+            std::cerr << "Failed to validate certificate. OpenSSL error: " << e.what() << std::endl;
+            exit(EXIT_FAILURE);
+        }
+        std::cout << "Certificate inside certificate chain successfully validated\n---" << std::endl;
+    }
+
+    /*
+     *   Negative case - Validation attempt of the certificate issued by other CA's.
```
garw: s.a.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,250 @@
...
+     */
+    {
+        X509Certificate::VerificationContext ctx;
+        ctx.addTrustedCertificate(rootCA)
+           .addIntermediateCertificate(intermediateCA);
+        std::cout << "Verifying certificate that is issued from other CA's" << std::endl;
```
garw: s.a.

examples/cert-validation.cpp
```diff
@@ -0,0 +1,250 @@
...
+        }
+        std::cout << "---" << std::endl;
+    }
+
+    /*
+     *   Add CRL to VerificationContext for intermediate CA which has
+     *   ceritificate stored in variable `cert` revoked (serial number 0x1002).
+     *   CRLs are added with addCertificateRevocationList() method.
+     *   Exception is thrown because validated certificate is a part of
+     *   root CA's CRL.
```
garw, line 184: Here it is correct :)

examples/cert-validation.cpp
```diff
@@ -0,0 +1,250 @@
...
+     *   Exception is expected.
+     */
+    {
+        X509Certificate::VerificationContext ctx;
+        ctx.addTrustedCertificate(rootCA)
+           .addIntermediateCertificate(intermediateCA)
+           .addCertificateRevocationList(rootCRL)
+           .enforceCrlsForAllCAs()
+           .enforceSelfSignedRootCertificate();
+        std::cout << "Checking enforcement of CRL's" << std::endl;
```
garw: here without '

garw (changes requested):
just a couple of remaining knitpicks. Overall, looks good now.

garw (approved):

PR #102, opened 2022-04-25
Small clean up of CMakeLists.txt

johnfxgalea-e:
Cleans-up CMakeLists.txt to include MoCOCrW's copyright notice and `project` command to avoid warning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,23 @@
+# #%L
+# %%
+# Copyright (C) 2022 BMW Car IT GmbH
+# %%
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# #L%
+
 cmake_minimum_required(VERSION 3.5)
+
+project("MoCOCrW" LANGUAGES CXX)
 
 include(CTest)                      # Enable test target in Makefile
 include(GNUInstallDirs)             # For standard Linux include directories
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -89,8 +108,11 @@ set(SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/src)
 #TODO set(DOC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/doc)
 set(TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR}/tests)
 
-# Enable all subprojects to find each other's include files
-include_directories(${SRC_DIR})
+include_directories(
+    ${SRC_DIR} # Enable all subprojects to find each other's include files
```
garw: What is the rationale behind this? In modern CMake, we work with remote targets. The include directories are then implictly added when the respective remote target is listed in the in target_link_libraries.
See, e.g., https://github.com/bmwcarit/MoCOCrW/blob/openssl1.1/src/CMakeLists.txt#L70

Anything that is broken for you in particular?

garw (changes requested):
Simple changes look ok. Can you please explain the rationale behind the include dir change. We want to avoid global include paths as much as possible to avoid accidental mixtures.

CMakeLists.txt
johnfxgalea-e: Unfortunately, this is more of a work-around to a known problem with finding OpenSSL libraries. Basically, building a recent version of OpenSSL (in my case > 3.00), at a custom path, results in installation to place the library inside a directory called "lib64". Despite setting OPENSSL_ROOT_DIR, `find_package` does not find the library:

```
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY)
```

As an initial work-around, I manually set OPENSSL variables as parameters so that cmake is able to find the library and headers (hence the rationale behind `include_directories`). However, thinking about this again, I don't think this is the right approach either. 

Turns out that a similar issue was encountered by CMake devs (https://gitlab.kitware.com/cmake/cmake/-/issues/22945) with FindOpenSSL. I built and installed the current version of CMake, namely 3.23.1, and cmake managed to find the library inside "lib64". On CMake version 3.16.3, it does not work for me. 

Therefore, at some point, we should bump up `cmake_minimum_required` so that MoCOCrW users can easily integrate custom OpenSSL builds. I think I should remove that change; users can simply rename lib64 to lib in the meantime, if preferred over changing the CMakeLists.txt

johnfxgalea-e:
PTAL

garw (approved):

PR #104, opened 2022-04-29
Fix test library target in CMakeLists.txt

johnfxgalea-e:
Adds CMakeLists.txt fix so that test no longer has missing OpenSSL target.


johnfxgalea-e:
@garw PTAL

garw (approved):

PR #105, opened 2022-04-29
Fix CA tests

istepic:
CA tests were failing randomly because CA's notBefore was set on
first use which then can cause problems in sanity check section of
CertificateAuthority::_signCSR function. If CA's notBefore was set
to +1 second compared to CA certificate's notBefore, it would
complain with "certificate is not valid yet" on the first
cert.verify(ctx). If CA's notBefore was set to -1 second compared
to issued cert, it would complaind with "certificate expired" on
the second cert.verify(ctx).

Solution was to set CA's notBefore and CA certificate's notBefore
to the same value in test SetUp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,27 @@ All notable changes to this project will be documented in this file.
 
 ## Fixed
 
+* CA Tests' SetUp was changed so that all the objects involved do not depend on time when
+  construction of object is made. This led to problems where object of CA class had notBefore
+  attribute set to greater value than CA's certificate's notBefore which should never happen.
+  This subtle bug in test SetUp has greater chance of appearing when running in slower
+  environments, e.g., qemu.
+
 ## Added
+
+* Exceptions with better error messages were added in sanity check section of
+  CertificateAuthority::_signCSR function. This provides better understanding of
+  scenarios which we dont allow:
+    - Issued certificate has greater notAfter attribute than CA's certificate (issued
+      certificate's validity period should not exceed issuing certificate's validity
+      period)
+    - Case with the CA's notBefore being larger than the issued certificate's notBefore. This
+      results in issued certificate that are valid *before* issuing certificate which
+      should never happen.
+  This is not a behavior change in the library in the sense that a certificate that was issued
+  before won't be issued anymore. Certificates with these properties were already rejected
+  by CertificateAuthority::signCSR but with a rather misleading and generic error message.
+  This change just improves the error reporting.
 
 # Release 4.1.1
 
--- a/src/ca.cpp
+++ b/src/ca.cpp
@@ -95,9 +95,20 @@ X509Certificate CertificateAuthority::_signCSR(const CertificateSigningRequest &
 
     _signCertificate(newCertificate.get(), _privateKey, signParams);
 
-    //Sanity check: certificate must be verifiable now
+    // Sanity check: certificate must be verifiable now
     try {
         auto cert = X509Certificate{std::move(newCertificate)};
+        // CA's notBefore can't be smaller than issued cert's notBefore i.e. CA can't
+        // issue a certificate older than CA itself
+        if (_rootCert.getNotBeforeAsn1() > cert.getNotBeforeAsn1()) {
+            throw MoCOCrWException("CSR: Issued certificate can not be valid before issuing "
+                                   "certificate (CA's cert)");
+        }
+        // CA's notAfter has to be bigger than Cert's notAfter.
+        if (_rootCert.getNotAfterAsn1() < cert.getNotAfterAsn1()) {
+            throw MoCOCrWException("CSR: Certificate's validity period should not exceed the "
+                                   "validity bounds of the issuing certificate (CA's cert).");
+        }
         X509Certificate::VerificationContext ctx;
         ctx.addTrustedCertificates({_rootCert})
             .addIntermediateCertificates({})
--- a/examples/ca-csr-example.cpp
+++ b/examples/ca-csr-example.cpp
@@ -32,6 +32,7 @@ using namespace mococrw;
 
 // Digest type to used by all examples
 static const DigestTypes digestType = DigestTypes::SHA512;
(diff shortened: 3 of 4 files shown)


src/ca.cpp
```diff
@@ -95,9 +95,24 @@ X509Certificate CertificateAuthority::_signCSR(const CertificateSigningRequest &
 
     _signCertificate(newCertificate.get(), _privateKey, signParams);
 
-    //Sanity check: certificate must be verifiable now
+    // Sanity check: certificate must be verifiable now
     try {
         auto cert = X509Certificate{std::move(newCertificate)};
+        // CA's notBefore can't be smaller than issued cert's notBefore i.e. CA can't
+        // issue a certificate older than CA itself
+        if (this->getRootCertificate().getNotBeforeAsn1() > cert.getNotBeforeAsn1()) {
```
garw: This project doesn't explictly use this-> to access members and we normally also don't use the public accessors if they just refer to an internal object.

src/ca.cpp
```diff
@@ -95,9 +95,24 @@ X509Certificate CertificateAuthority::_signCSR(const CertificateSigningRequest &
...
+            throw MoCOCrWException(
+                std::string("CSR: CA shouldn't be able to issue a certificate older than"
```
garw: Explicitly creating a std::string shouldn't be necessary.

src/ca.cpp
```diff
@@ -95,9 +95,24 @@ X509Certificate CertificateAuthority::_signCSR(const CertificateSigningRequest &
...
+                            " itself. Something's wrong.")
+                );
+        }
+        // CA's notAfter has to be bigger than Cert's notAfter.
+        if (this->getRootCertificate().getNotAfterAsn1() < cert.getNotAfterAsn1()) {
```
garw: s.a.

src/ca.cpp
```diff
@@ -95,9 +95,24 @@ X509Certificate CertificateAuthority::_signCSR(const CertificateSigningRequest &
...
+            throw MoCOCrWException(
+                std::string("CSR: Certificate's validity period should not exceed the "
```
garw: s.a.

garw (changes requested):
See the comments. Moreover, please extend the commit message to reflect that you add the sanity check to give better error messages. Also emphasize why this is no behavioral change of the librarys.
This is also something that should be reflected in the change-log as it will be publicly visible.

CHANGELOG.md
```diff
@@ -8,8 +8,25 @@ All notable changes to this project will be documented in this file.
...
+  attribute set to greater value than CA's certificate's notBefore which should never happen.
+  This subtle bug in test set up has greater chance of appearing when running in slower
+  environments, e.g., qemu.
+
 ## Added
 
+* Exceptions with better error messages were added in sanity check section of
+  CertificateAuthority::_signCSR function. This provides better understanding of
+  scenarios which we dont allow:
+    - Issued certificate has greater notAfter attribute that CA's certificate (issued
```
garw:
```suggestion
    - Issued certificate has greater notAfter attribute than CA's certificate (issued
```

CHANGELOG.md
```diff
@@ -8,8 +8,25 @@ All notable changes to this project will be documented in this file.
...
+      certificates' validity period should not exceed issuing certificates' validity
```
garw:
```suggestion
      certificate's validity period should not exceed issuing certificate's validity
```

CHANGELOG.md
```diff
@@ -8,8 +8,25 @@ All notable changes to this project will be documented in this file.
...
+      period)
+    - CA's notBefore attribute is earlier than CA's certificates' notBefore date. This
```
garw: CA's attribute vs CA's attribute? This needs an issued certificate somewhere?
I'd just say here: 

> Accordingly, for the case with the CA's notBefore being larger than the issued certificate's notBefore.

CHANGELOG.md
```diff
@@ -8,8 +8,25 @@ All notable changes to this project will be documented in this file.
...
+      results in issued certificate that are valid *before* issuing certificate which
+      should never happen.
+  This does not change any functionality of the library.
```
garw: I'd be a bit more verbose here:
This is no behavior change in the library in the sense that a certificate that was issued before won't be issued anymore. Certificates with these properties were already rejected by `CertificateAuthority::signCSR` but with a rather misleading and generic error message. This change just improves the error reporting.

src/ca.cpp
```diff
@@ -95,9 +95,20 @@ X509Certificate CertificateAuthority::_signCSR(const CertificateSigningRequest &
...
 
-    //Sanity check: certificate must be verifiable now
+    // Sanity check: certificate must be verifiable now
     try {
         auto cert = X509Certificate{std::move(newCertificate)};
+        // CA's notBefore can't be smaller than issued cert's notBefore i.e. CA can't
+        // issue a certificate older than CA itself
+        if (_rootCert.getNotBeforeAsn1() > cert.getNotBeforeAsn1()) {
+            throw MoCOCrWException("CSR: CA shouldn't be able to issue a certificate older than"
+                                   " itself. Something's wrong.");
```
garw: Please be a bit more formal in the error message. The message for the other case looks good. Just use the same here with changed roles.

CHANGELOG.md
```diff
@@ -8,8 +8,25 @@ All notable changes to this project will be documented in this file.
 
 ## Fixed
 
+* CA Tests' SetUp was changed so that all the objects involved do not depend on time when
+  construction of object is made. This led to problems where object of CA class had notBefore
+  attribute set to greater value than CA's certificate's notBefore which should never happen.
+  This subtle bug in test set up has greater chance of appearing when running in slower
```
garw: Reads strange. "bug in test set up" doesn't group very well in my mind.

garw (changes requested):
Just a few knitpicks mostly language.

garw (approved):
lg2m now

PR #106, opened 2022-05-12
Add rfc attempt

johnfxgalea-e:
Just a temporary PR for quick feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,10 @@ env:
 
   DOCKER_WITH_HSM_TAG: buildenv_with_hsm
   DOCKER_WITH_HSM_FILE_PATH: ./docker-build-env/Dockerfile-With-HSM
- 
+
+  ASAN_OPTIONS: "fast_unwind_on_malloc=false"
+  LSAN_OPTIONS: "suppressions=/src/lsan_suppression.txt"
+
   TOKEN_LABEL: token-label
   USER_PIN: 1234
   SO_PIN: 4321
@@ -88,7 +91,9 @@ jobs:
 
       - name: "Run tests"
         run: |
-          docker run --rm -e SOFTHSM2_CONF="${{ matrix.softhsm2_conf }}" \
+          # Initialise and delete token prior to and after the execution of ctests
+          # so that HSM tests/examples can work.
+          docker run --rm -e ASAN_OPTIONS="${{ env.ASAN_OPTIONS }}" -e LSAN_OPTIONS="${{ env.LSAN_OPTIONS }}" \
             -u "$UID"${{ matrix.softhsm2_grp }} \
             -v "$PWD:/src:rw" -w /src/build ${{ matrix.docker_tag }}  bash -c \
             'if [ "${{ matrix.hsm_flag }}" = "HSM_ENABLED=ON" ]; then \
--- a/cmake/AddressSanitizerTarget.cmake
+++ b/cmake/AddressSanitizerTarget.cmake
@@ -114,6 +114,7 @@ string(CONCAT _asan_options
     "detect_stack_use_after_return=true,"
     "${ASAN_OPTIONS}"
 )
+
 add_custom_target(asan
     COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR}
 
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -26,6 +26,30 @@ add_test(
     COMMAND hash-example
 )
 
+# Only build HSM example if HSM features have been enabled.
+#
+# TODO: Make the HSM flag be an imported variable so clients
+# can easily determine its status.
+if(HSM_ENABLED)
+    add_executable(hsm-store-example hsm-store-example.cpp)
+    target_link_libraries(hsm-store-example PUBLIC MoCOCrW::mococrw)
+    target_compile_definitions(hsm-store-example PUBLIC HSM_ENABLED)
+    add_test(
+	NAME hsmStoreExample
+        COMMAND hsm-store-example
+    )
+
+    add_executable(hsm-gen-example hsm-gen-example.cpp)
+    target_link_libraries(hsm-gen-example PUBLIC MoCOCrW::mococrw)
+    target_compile_definitions(hsm-gen-example PUBLIC HSM_ENABLED)
+    add_test(
+	NAME hsmGenExample
+        COMMAND hsm-gen-example
+    )
+else()
+    message(WARNING "Skipping HSM example, as HSM features are disabled.")
+endif()
+
 add_executable(kdf-example kdf-example.cpp)
 target_link_libraries(kdf-example PUBLIC MoCOCrW::mococrw)
 add_test(
--- a/src/hsm.cpp
+++ b/src/hsm.cpp
@@ -1,0 +1,51 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2022 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include "mococrw/hsm.h"
+#include <boost/format.hpp>
+
+namespace mococrw
+{
+using namespace openssl;
+
+HSM::HSM() {}
+
+HsmEngine::HsmEngine(const std::string &id, const std::string &modulePath, const std::string &pin)
+        : _id(id), _modulePath(modulePath), _pin(pin)
+{
+    // Fetch engine via ID.
+    _engine = _ENGINE_by_id(id.c_str());
+
+    _ENGINE_ctrl_cmd_string(_engine.get(), "MODULE_PATH", _modulePath);
+    _ENGINE_ctrl_cmd_string(_engine.get(), "PIN", _pin);
+    _ENGINE_init(_engine.get());
+}
+
+HsmEngine::~HsmEngine() { _ENGINE_finish(_engine.get()); }
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPublicKey(const std::string &keyID)
+{
+    return _ENGINE_load_public_key(_engine.get(), keyID.c_str());
+}
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
+{
+    return _ENGINE_load_private_key(_engine.get(), keyID.c_str());
+}
+
+}  // namespace mococrw
--- a/src/mococrw/hsm.h
+++ b/src/mococrw/hsm.h
@@ -1,0 +1,162 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2022 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+
+#include <boost/optional/optional.hpp>
+#include "openssl_wrap.h"
+#include "p11_wrap.h"
+
+namespace mococrw
+{
+/**
+ * Driver for Hardware Security Modules (HSMs).
+ */
+class HSM
+{
+public:
+    HSM();
+    virtual ~HSM() {}
+
+protected:
+    /**
+     * Returns the name of the HSM driver.
+     */
+    virtual const std::string getName() = 0;
+
+    /**
+     *  Loads public key from HSM.
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string &keyID) = 0;
+
+    /**
+     * Stores public key to HSM.
+     */
+    virtual void storePublicKey(EVP_PKEY *key,
+                                const std::string &label,
+                                const std::string &keyID) = 0;
+
+    /**
+     * Loads private key from HSM.
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPrivateKey(const std::string &keyID) = 0;
+
+    /**
+     * Stores private key to HSM.
+     */
+    virtual void storePrivateKey(EVP_PKEY *key,
+                                 const std::string &label,
+                                 const std::string &keyID) = 0;
+
+    /**
+     * Generates a key pair via HSM.
+     *
+     * \note Currently, RSA keys are only supported.
+     */
+    virtual void generateKey(unsigned int bits,
+                             const std::string &label,
+                             const std::string &id) = 0;
+
+    // Many of protected functions provided by the HSM class are seen
+    // as internal, not to be used by the User of MoCOCrW but specific
+    // friends:
+    friend class AsymmetricPublicKey;
+    friend class AsymmetricKeypair;
+    friend class RSASpec;
+};
+
+/**
+ * HSM driver implemented using OpenSSL's engine interface.
+ */
+class HsmEngine : public HSM
+{
+public:
+    HsmEngine(const std::string &id, const std::string &modulePath, const std::string &pin);
+    virtual ~HsmEngine();
+
+    ENGINE *internal();
+    const ENGINE *internal() const;
+
+protected:
+    /** Pointer to OpenSSL ENGINE. */
+    openssl::SSL_ENGINE_Ptr _engine;
+    /** Engine ID. */
+    const std::string &_id;
+    /** Path to Module. */
+    const std::string &_modulePath;
+    /** Pin to access PKCS11 Engine. */
+    const std::string &_pin;
+
+    virtual const std::string getName() override = 0;
+
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string &keyID) override;
+
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPrivateKey(const std::string &keyID) override;
+};
+
+/**
+ * An HSM Engine that leverages the PKCS11 interface.
+ */
+class Pkcs11Engine : public HsmEngine
+{
+public:
+    Pkcs11Engine(const std::string &module_path, const std::string &pin);
+    virtual ~Pkcs11Engine();
+
+protected:
+    /** The LibP11 context */
+    p11::P11_PKCS11_CTX_Ptr pkcs11Ctx;
+    p11::P11_SlotInfo slotInfo;
+
+    virtual const std::string getName() override;
+
+    virtual void storePublicKey(EVP_PKEY *key,
+                                const std::string &label,
+                                const std::string &keyID) override;
+
+    virtual void storePrivateKey(EVP_PKEY *key,
+                                 const std::string &label,
+                                 const std::string &keyID) override;
+
+    virtual void generateKey(unsigned int bits,
+                             const std::string &label,
+                             const std::string &id) override;
+
+    /**
+     * A helper function to prepare for an operation, such as key storage or generation,
+     * on the HSM. It returns the slot, which the operation will be performed on.
+     *
+     * Internally, this function finds and authenticates with the first slot of an appropriate
+     * token. A session will be open with both read and write permissions.
+     */
+    virtual p11::P11_PKCS11_SLOT_PTR startOperation();
+
+    /**
+     * A helper function to finish after an operation, such as key storage or generation,
+     * on the HSM. It takes the pointer of the slot returned earlier by startOperation().
+     *
+     * Internally, the function will perform de-authentication with the slot.
+     */
+    virtual void finishOperation(p11::P11_PKCS11_SLOT_PTR token);
+
+private:
+    /** Engine ID for PKCS11 HSM. */
+    static const std::string ENGINE_ID;
+};
+
+}  // namespace mococrw
--- a/src/mococrw/key.h
+++ b/src/mococrw/key.h
@@ -1,7 +1,7 @@
 /*
  * #%L
  * %%
- * Copyright (C) 2018 BMW Car IT GmbH
+ * Copyright (C) 2022 BMW Car IT GmbH
  * %%
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,9 @@
  * #L%
  */
 #pragma once
+#ifdef HSM_ENABLED
+#include "mococrw/hsm.h"
+#endif
 #include "openssl_wrap.h"
 
 namespace mococrw
@@ -82,6 +85,19 @@ public:
      * @throws This method may throw an OpenSSLException if OpenSSL indicates an error
      */
     static AsymmetricPublicKey readPublicKeyFromPEM(const std::string& pem);
+
+#ifdef HSM_ENABLED
+    /**
+     * Loads a public key from an HSM, creating an @ref AsymmetricPublicKey
+     * object as a result.
+     */
+    static AsymmetricPublicKey readPublicKeyFromHSM(HSM& hsm, const std::string& keyID);
+
+    /**
+     * Stores a public key to an HSM.
+     */
+    void writePublicKeyToHSM(HSM& hsm, const std::string& keyLabel, const std::string& keyID);
+#endif
 
     /**
      * @brief Returns a public key object based on the provided EC point
@@ -223,6 +239,21 @@ public:
      */
     static AsymmetricKeypair generateRSA();
 
+#ifdef HSM_ENABLED
+    /**
+     * Generate a RSA asymmetric keypair with default Spec.
+     *
+     * Currently, a default-spec is an RSASpec with 2048
+     * bit modulus. (@see RSASpec)
+     *
+     * @throws This method may throw an OpenSSLException or
+     * a P11Excetion if OpenSSL or P11 indicates an error.
+     */
+    static AsymmetricKeypair generateRSAViaHSM(HSM& hsm,
+                                               const std::string& label,
+                                               const std::string& id);
+#endif
+
     /**
      * Generate an ECC asymmetric keypair with default Spec.
      *
@@ -254,6 +285,19 @@ public:
     static AsymmetricKeypair readPrivateKeyFromPEM(const std::string& pem,
                                                    const std::string& password);
 
+#ifdef HSM_ENABLED
+    /**
+     * Loads a private key from an HSM, creating an @ref AsymmetricPublicKey
+     * object as a result.
+     */
+    static AsymmetricKeypair readPrivateKeyFromHSM(HSM& hsm, const std::string& keyID);
+
+    /**
+     * Stores a private key to an HSM.
+     */
+    void writePrivateKeyToHSM(HSM& hsm, const std::string& keyLabel, const std::string& keyID);
+#endif
+
 private:
     AsymmetricKeypair(AsymmetricKey&& key) : AsymmetricPublicKey{std::move(key)} {}
 };
@@ -278,6 +322,16 @@ public:
      *
      */
     virtual AsymmetricKey generate() const = 0;
+
+#ifdef HSM_ENABLED
+    /**
+     * Generate an AsymmetricKey via HSM from this
+     * spec instance.
+     */
+    virtual AsymmetricKey generateViaHSM(HSM& hsm,
+                                         const std::string& label,
+                                         const std::string& id) const = 0;
+#endif
 };
 /**
  * RSA specification used to hold the necessary parameters to generate a RSA key pair.
@@ -292,6 +346,11 @@ public:
     explicit RSASpec(unsigned int numBits) : _numBits{numBits} {}
     RSASpec() : RSASpec{defaultSize} {}
     AsymmetricKey generate() const override;
+#ifdef HSM_ENABLED
+    AsymmetricKey generateViaHSM(HSM& hsm,
+                                 const std::string& label,
+                                 const std::string& id) const override;
+#endif
     inline unsigned int numberOfBits() const { return _numBits; }
 
 private:
@@ -313,6 +372,11 @@ public:
     ECCSpec() : ECCSpec{defaultCurveNid} {}
     inline openssl::ellipticCurveNid curve() const { return _curveNid; }
     AsymmetricKey generate() const override;
+#ifdef HSM_ENABLED
+    AsymmetricKey generateViaHSM(HSM& hsm,
+                                 const std::string& label,
+                                 const std::string& id) const override;
+#endif
 
 private:
     openssl::ellipticCurveNid _curveNid;
--- a/src/mococrw/p11_wrap.h
+++ b/src/mococrw/p11_wrap.h
@@ -1,0 +1,227 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2022 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+
+#include "p11_lib.h"
+#include "util.h"
+
+namespace mococrw
+{
+namespace p11
+{
+/**
+ * Template to wrap OpenSSLs "_free" functions
+ * into a functor so that a std::unique_ptr
+ * can use them.
+ */
+template <class R, class P, R(Func)(P *)>
+struct P11Deleter
+{
+    template <class T>
+    void operator()(T *ptr) const noexcept
+    {
+        if (ptr) {
+            Func(ptr);
+        }
+    }
+};
+
+/**
+ * An Exception for LibP11 errors.
+ *
+ * This exception is thrown by all methods when an OpenSSL error occurs.
+ */
+class P11Exception final : public std::exception
+{
+public:
+    template <class StringType>
+    explicit P11Exception(StringType &&message) : _message{std::forward<StringType>(message)}
+    {
+    }
+
+    /**
+     * Generate an exception with the defalt LibP11 error-string
+     * as message.
+     *
+     */
+    P11Exception() : _message{generateP11ErrorString()} {}
+
+    const char *what() const noexcept override { return _message.c_str(); }
+
+private:
+    static std::string generateP11ErrorString();
+    std::string _message;
+};
+
+/*
+ * Wrap all the pointer-types returned by LibP11.
+ */
+
+using P11_PKCS11_CTX_Ptr =
+        std::unique_ptr<PKCS11_CTX, P11Deleter<void, PKCS11_CTX, lib::LibP11::P11_PKCS11_CTX_free>>;
+
+/*
+ * The memory referred to by these pointer types live out of scope of their pointers.
+ * Therefore, simply wrap the C pointer, without use of smart pointer (this is a bit ugly).
+ */
+using P11_PKCS11_SLOT_PTR = PKCS11_SLOT *;
+using P11_PKCS11_TOKEN_PTR = PKCS11_TOKEN *;
+
+/* Below is the "wrapped" LibP11 library. By convention, all functions start with an
+ * underscore to visually distinguish them from the methods of the class P11Lib and
+ * from the native LibP11 methods.
+ */
+
+/**
+ * Returns a new PKCS11 context for HSM interaction.
+ */
+P11_PKCS11_CTX_Ptr _PKCS11_CTX_new(void);
+
+/**
+ * Loads the PKCS11 module to use.
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ * @param module The PKCS#11 module filename.
+ */
+void _PKCS11_CTX_load(PKCS11_CTX *ctx, const std::string &module);
+
+struct P11_SlotInfo
+{
+    PKCS11_SLOT *_slots;     // Array consisting of slot descriptors on the HSM.
+    unsigned int _numSlots;  // The size of the array, i.e., the number of slots.
+
+    P11_SlotInfo();
+    P11_SlotInfo(PKCS11_SLOT *slots, unsigned int numSlots);
+};
+
+/**
+ * Obtains information of all current slots on the HSM.
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ */
+P11_SlotInfo _PKCS11_enumerate_slots(PKCS11_CTX *ctx);
+
+/**
+ * Destroys slot information obtained by _PKCS11_enumerate_slots().
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ * @param slotInfo The slot-list information to destroy.
+ */
+void _PKCS11_release_all_slots(PKCS11_CTX *ctx, P11_SlotInfo &slotInfo);
+
+/**
+ * Finds a token on the HSM according to the passed slot information \p slotInfo.
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ * @param slotInfo Slot information obtained via _PKCS11_enumerate_slots().
+ */
+P11_PKCS11_SLOT_PTR _PKCS11_find_token(PKCS11_CTX *ctx, P11_SlotInfo &slotInfo);
+
+/**
+ * Returns the token descriptor associated with \p slot.
+ *
+ * @param slot The slot of the returned token.
+ */
+P11_PKCS11_TOKEN_PTR _PKCS11_getTokenFromSlot(PKCS11_SLOT *slot);
+
+enum SessionMode { ReadOnly = 0, ReadWrite = 1 };
+
+/**
+ * Opens a sessions with the passed \p slot.
+ *
+ * @param slot The slot to open session with.
+ * @param rw Whether to open the session in read/write mode. Set to != 0
+ * if read/write mode is required.
+ */
+void _PKCS11_open_session(PKCS11_SLOT *slot, SessionMode mode);
+
+/**
+ * Performs a user login to the HSM.
+ *
+ * @param slot The slot to consider for login.
+ * @param pin The pin to perform the login.
+ *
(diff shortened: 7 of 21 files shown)


src/mococrw/openssl_wrap.h
```diff
@@ -77,6 +77,18 @@ struct SSLDeleter
     }
 };
 
+template <class P, int(Func)(P*)>
+struct SSLIntDeleter
```
garw: What is the purpose of this copy of SSLDeleter?

src/mococrw/pkcs11_engine.h
```diff
@@ -0,0 +1,48 @@
...
+{
+public:
+    explicit Pkcs11Engine(const std::string &module_path, const std::string &pin);
+    ENGINE *internal();
+    const ENGINE *internal() const;
+
+    /**
+     * Load public key via Engine.
+     */
+    AsymmetricPublicKey loadPublicKey(const std::string& key_id);
```
garw: I'm not sure if should have this functionality here. The other loaders are in the copy of the key objects. I think I would prefer to have them there as well, or let's say at least an option to use it from there so that you can say something like
`AsymmetricPublicKey::readKeyFromEngine(engine&, key_id&)`

src/mococrw/pkcs11_engine.h
```diff
@@ -0,0 +1,48 @@
...
+#include "mococrw/asymmetric_crypto_ctx.h"
+
+namespace mococrw
+{
+
+/**
+ * A PKCS-11 Engine
+ *
+ */
+class Pkcs11Engine
```
garw: Maybe we want to be a bit more general here and have an engine class that is subclassed with a PKCS11Engine class?

src/mococrw/pkcs11_engine.h
```diff
@@ -0,0 +1,48 @@
...
+    explicit Pkcs11Engine(const std::string &module_path, const std::string &pin);
+    ENGINE *internal();
+    const ENGINE *internal() const;
+
+    /**
+     * Load public key via Engine.
+     */
+    AsymmetricPublicKey loadPublicKey(const std::string& key_id);
+
+protected:
```
garw: Any particular reason for this? Normally, we default to private…

src/pkcs11_engine.cpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2018 BMW Car IT GmbH
```
garw: Update date for new files ;)

src/pkcs11_engine.cpp
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include <boost/format.hpp>
+#include <boost/scoped_ptr.hpp>
```
garw: Is this header-only? 
If we want to introduce new dependencies, we should discuss. This lib should try to have as little dependencies as possible.

src/pkcs11_engine.cpp
```diff
@@ -0,0 +1,62 @@
...
+const std::string Pkcs11Engine::ENGINE_ID = "pkcs11";
+
+Pkcs11Engine::Pkcs11Engine(const std::string &module_path,
+		const std::string &pin) {
+	_engine = _ENGINE_by_id(ENGINE_ID.c_str());
+
+	// Set module path:
+	std::vector<char> module_path_copy(module_path.begin(), module_path.end());
+	module_path_copy.push_back('\0');
+	_ENGINE_ctrl_cmd(_engine.get(), "MODULE_PATH", 0,
```
garw: Any particular reasons to not use std::string::c_str? If yes, we should think about putting this into the wrap layer and let `_ENGINE_ctrl_cmd` accept `std::string`s

src/pkcs11_engine.cpp
```diff
@@ -0,0 +1,62 @@
...
+			(void*) &module_path_copy[0], nullptr, 1);
```
garw: Avoid c-style casts.

src/pkcs11_engine.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const ENGINE* Pkcs11Engine::internal() const {
+	return _engine.get();
+}
+
+AsymmetricPublicKey Pkcs11Engine::loadPublicKey(const std::string& key_id)
+{
+    SSL_EVP_PKEY_Ptr evp_key(_ENGINE_load_public_key(_engine.get(), key_id.c_str()));
+    return AsymmetricPublicKey{std::move(evp_key)};
```
garw: You could probably just return it. Maybe needs an explicit constructor call but no need for a temporary object. This then also enables RVO.

garw (commented):
Looks from the overall perspective quite good.
I left some minior fgeneral comments in some of the places.

Did you check how the interface will change for OpenSSL 3.0 crypto providers? Can we easily adapt this design also for providers?

src/pkcs11_engine.cpp
johnfxgalea-e: Done.

src/pkcs11_engine.cpp
johnfxgalea-e: This was a stale inclusion which I forgot to remove for the quick PR.

Done.

src/pkcs11_engine.cpp
johnfxgalea-e: Done.

src/pkcs11_engine.cpp
johnfxgalea-e: Done.

src/mococrw/pkcs11_engine.h
johnfxgalea-e: Done.

src/mococrw/pkcs11_engine.h
johnfxgalea-e: Done.

src/mococrw/openssl_wrap.h
johnfxgalea-e: It has a different prototype for Func, with int being returned. 

Do you know of a way to make the C++ code cleaner, by any chance?

src/pkcs11_engine.cpp
johnfxgalea-e: Done.

src/mococrw/pkcs11_engine.h
johnfxgalea-e: Done.

src/mococrw/openssl_wrap.h
garw: Hm, wouldn't this just work if we remove the function pointer primitive and just say `typename Func` here?

src/mococrw/hsm.h
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Returns whether the HSM has been initialised.
+     */
+    bool isInitialised();
+
+
+
+private:
+    bool is_init;
```
istepic: Could we do the initialization in the constructor of the derived classes and keep this base class pure virtual?

src/mococrw/hsm.h
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Returns the name of the HSM driver.
+     */
+    virtual const std::string getName() = 0;
+
+    /**
+     * Initialises the HSM.
+     */
+    virtual void initialise();
```
istepic: Why do we need this as opposed to using constructors of the derived classes?

src/mococrw/hsm.h
istepic: Is this necessary? Could we make this base class pure virtual?

src/mococrw/hsm.h
```diff
@@ -0,0 +1,110 @@
...
+    /** Engine ID. */
+    const std::string _id;
+    /** Path to Module. */
+    std::string _modulePath;
+};
+
+/**
+ * An HSM Engine that leverages the PKCS11 interface.
+ */
+class Pkcs11Engine : public HsmEngine
```
istepic, line 115: Most HSM's use PKCS#11 interface so this shouldn't be necessary.

src/mococrw/key.h
```diff
@@ -91,6 +92,12 @@ class AsymmetricPublicKey
      */
     static AsymmetricPublicKey readPublicKeyFromPEM(const std::string& pem);
 
+    /**
+     * Loads a public key from an HSM, creating an @ref AsymmetricPublicKey
+     * object as a result.
+     */
+    static AsymmetricPublicKey readPublicKeyFromHSM(HSM &hsm, const std::string& keyID);
```
istepic:
```suggestion
    static AsymmetricPublicKey readPublicKeyFromHSM(HSM &hsm, const std::string& keyID);
```
```suggestion
    static AsymmetricPublicKey readPublicKeyFromHSM(const HSM &hsm, const std::string& keyID);
```

src/hsm.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+Pkcs11Engine::Pkcs11Engine(std::string &module_path, std::string &pin)
+    : HsmEngine(ENGINE_ID, module_path) {
+  _pin = pin;
+}
+
+void Pkcs11Engine::initialise() {
+  _ENGINE_ctrl_cmd(_engine.get(), "PIN", 0, _pin, nullptr, 1);
+  HsmEngine::initialise();
```
istepic: Chain-calling the initalise functions is a bit confusing

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: The rationale is to have the constructor set necessary options/information, and then have initialise() be the actual function that internally invokes _ENGINE_init(). I think separation helps design.

src/mococrw/hsm.h
johnfxgalea-e: Since this boolean is applicable to all sub-classes, it makes sense to have it here. Why do you feel pure virtual is a must?

johnfxgalea-e: See above comment.

src/mococrw/hsm.h
johnfxgalea-e: Pkcs11Engine are OpenSSL engines. We could have Pkcs11Providers in the future, hence the proposal of this class hierarchy.

src/mococrw/key.h
johnfxgalea-e: I'm not sure if this is the best as this function internally calls `ENGINE_load_public_key(ENGINE *e, const char *key_id, UI_METHOD *ui_method, void *callback_data)`.

istepic: Why should that matter? I don't see `hsm` object being changed anywhere in the call graph to warrant non-consting it.

johnfxgalea-e: Because an HSM is an abstraction of `Engine *e` and the `ENGINE_load_public_key` prototype does not guarantee that the engine referred to by `e` is not modified by said function.

istepic: In `hsm` object you are holding a pointer to the `ENGINE` object and that pointer shouldn't change. Or should it?

johnfxgalea-e: Sure the pointer won't change, but recall const is not transitive to the engine data. I don't want users to mistakenly think that Engine data is also const when passing a const HSM object. I guess I can add your recommended change and insert a note as documentation to inform users.

istepic: I'm not sure why users of this API would care what happens with underlying `ENGINE` object that is referenced by `hsm` object. Will user at any point want to know what's going on in the underlying `ENGINE` object? This is what you're trying to abstract. AFAIC we can resolve this conversation since there are probably more important things that have to be discussed first and there are other instances where const is not used correctly.

johnfxgalea-e: If this review comment is a nit-pick, then I would prefer to leave it as original.... Resolving.

.github/workflows/ci.yml
```diff
@@ -61,4 +65,12 @@ jobs:
 
       - name: "Run tests"
         run: |
-          docker run --rm -e SOFTHSM2_CONF='/usr/share/softhsm/softhsm2.conf' -u "$UID":softhsm -v "$PWD:/src:rw" -w /src/build ${{ env.DOCKER_TAG }} bash -c 'ctest -j $(nproc)'
+          # Initialise and delete token prior to and after the execution of ctests
+          # so that HSM tests/examples can work.
+          # Unfortunately, we need to use cmake --build to target and run tests with our ASAN setup,
```
garw: Can you be a bit more verbose here. I don't understand this necessity and the reasons. It also wasn't present before or was this removed in a previous commit introducing softhsm support?

.github/workflows/ci.yml
```diff
@@ -61,4 +65,12 @@ jobs:
...
+          # as defined in AddressSanitizerTarget.cmake.
+          docker run --rm -e SOFTHSM2_CONF='/usr/share/softhsm/softhsm2.conf' -u "$UID":softhsm \
+            -v "$PWD:/src:rw" -w /src/build ${{ env.DOCKER_TAG }} bash -c \
+            'softhsm2-util --init-token --free --label ${{ env.TOKEN_LABEL }} --pin ${{ env.USER_PIN }} --so-pin ${{ env.SO_PIN  }} \
+            && if [ "${{ matrix.build_type }}" = "ASAN" ]; then cmake --build . --target "asan"; else ctest; fi \
```
garw: Did you remove parallelization of tests intentionally? If yes, please put a comment here so that it isn't later-on added by someone unsuspectingly.

cmake/AddressSanitizerTarget.cmake
```diff
@@ -111,17 +111,25 @@ endif()
 # Add asan target.
 string(CONCAT _asan_options
     "detect_odr_violation=0,max_free_fill_size=65535,"
-    "detect_stack_use_after_return=true,"
+    "fast_unwind_on_malloc=false,detect_stack_use_after_return=true,"
     "${ASAN_OPTIONS}"
 )
+
+set(LSAN_SUPPRESSION lsan_suppression.txt)
```
garw: Not sure if this is the right approach. We use/share this cmake File across projects and may want to update it here again at some point. Can't you just add the supression list from the outside in the project configuration? To my understanding, this is what the LSAN_OPTIONS variable is about.

cmake/AddressSanitizerTarget.cmake
```diff
@@ -111,17 +111,25 @@ endif()
 # Add asan target.
 string(CONCAT _asan_options
     "detect_odr_violation=0,max_free_fill_size=65535,"
-    "detect_stack_use_after_return=true,"
+    "fast_unwind_on_malloc=false,detect_stack_use_after_return=true,"
```
garw: Can we briefly check if this was updated internally as well? Otherwise - reason?

examples/CMakeLists.txt
```diff
@@ -26,6 +26,27 @@ add_test(
     COMMAND hash-example
 )
 
+# Only build HSM example if HSM features have been enabled.
+if(HSM_ENABLED)
+    add_executable(hsm-store-example hsm-store-example.cpp)
+    target_link_libraries(hsm-store-example PUBLIC MoCOCrW::mococrw)
+    target_compile_definitions(hsm-store-example PUBLIC HSM_ENABLED)
```
garw, line 36: This probably needs some rework. The status if the lib was compiled with out without HSM support should come from the imported target. But we can merge it like this for the time being. It will just be required for the real thing as otherwise the customers need to take care of this all the time - which isn't particularly good to handle.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Re-Enumerate slots/tokens of the HSM. This is often done to update slot
+     * list after adding/removing tokens/slots.
+     */
+    virtual void reenumerateSlots() = 0;
+
+    /**
+     *  Loads public key from HSM.
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string keyID) = 0;
```
garw: What kind of API is this supposed to be? internal one? Real customer APIs should not return OpenSSL objects but rather the wrappers that the library offers. Same for the ones below. 
If it is supposed to be an internal API, please move into some kind of `internal` namespace.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+
+namespace mococrw
+{
+/**
+ * Driver for Hardware Security Modules (HSMs).
+ */
+class HSM
+{
+public:
+    explicit HSM();
```
garw: Explicit doesn't make sense for a constructor with empty parameter list.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+    virtual void storePrivateKey(EVP_PKEY *key,
+                                 const std::string label,
+                                 const std::string keyID) = 0;
+
+    /**
+     * Generates a key pair via HSM.
+     *
+     * \note Currently, RSA keys are only supported.
+     */
+    virtual void generateKey(unsigned int bits, std::string label, std::string id) = 0;
```
garw: I don't like this API because it silently assumes that keys are always RSA. Moreover, we already have a mechanism to specify information about private keys with the KeySpec API as defined in key.h. This should be reused here.

I also believe that it would be better if it actually returns the produced key.

src/mococrw/hsm.h
garw: We usually pass const refs of strings in C++

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+    virtual ~HsmEngine();
+
+    ENGINE *internal();
+    const ENGINE *internal() const;
+
+    virtual const std::string getName() = 0;
+
+    virtual void reenumerateSlots();
+
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string keyID);
```
garw: We prefer `override` in derived functions as it ensures that there is a virtual base.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+class HsmEngine : public HSM
+{
+public:
+    explicit HsmEngine(const std::string &id, std::string &modulePath, std::string &pin);
+    virtual ~HsmEngine();
+
+    ENGINE *internal();
+    const ENGINE *internal() const;
+
+    virtual const std::string getName() = 0;
```
garw: This doesn't do anything?

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+    virtual void generateKey(unsigned int bits, std::string label, std::string id) = 0;
+};
+
+/**
+ * HSM driver implemented using OpenSSL's engine interface.
+ */
+class HsmEngine : public HSM
+{
+public:
+    explicit HsmEngine(const std::string &id, std::string &modulePath, std::string &pin);
```
garw: Why should this be explicit? This prevents calling it with c string constants or requires noisy explicit conversion.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+    virtual const std::string getName();
+
+    virtual void storePublicKey(EVP_PKEY *key, const std::string label, const std::string keyID);
+
+    virtual void storePrivateKey(EVP_PKEY *key, const std::string label, const std::string keyID);
+
+    virtual void generateKey(unsigned int bits, std::string label, std::string id);
+
+protected:
+    /** The LibP11 context */
```
garw: Why are these protected? What is the rationale of them not being private?

src/mococrw/openssl_wrap.h
```diff
@@ -63,7 +63,7 @@ using CmacCipherTypes = mococrw::CmacCipherTypes;
  * into a functor so that a std::unique_ptr
  * can use them.
  */
-template <class P, void(Func)(P*)>
+template <class R, class P, R(Func)(P*)>
```
garw: I fear this is an ABI breaking change. Though I like the avoidance of code duplication, I'm not sure it is worth it. We would need to remark this in the changelog and pull the major version of this lib and make sure everyone is recompiling. Should work for most scenarios but there may be exceptions…

src/mococrw/openssl_wrap.h
```diff
@@ -1495,5 +1507,59 @@ SSL_BIGNUM_Ptr _BN_bin2bn(const uint8_t* data, size_t dataLen);
  */
 std::vector<uint8_t> _BN_bn2binpad(const BIGNUM* bignum, int tolen);
 
+/* Engine related */
+
+/**
+ * Loads an engine based on the passed ID \p engine_id.
+ */
+SSL_ENGINE_Ptr _ENGINE_by_id(std::string engine_id);
```
garw: const ref?

src/mococrw/openssl_wrap.h
```diff
@@ -1495,5 +1507,59 @@ SSL_BIGNUM_Ptr _BN_bin2bn(const uint8_t* data, size_t dataLen);
...
+/**
+ * Initialises engine \p e.
+ */
+void _ENGINE_init(ENGINE* e);
+
+/**
+ * Issues a command to the engine \p e. Takes an opaque void * value as input data.
+ */
+void _ENGINE_ctrl_cmd(ENGINE* e,
+                      std::string cmd_name,
```
garw: const ref?

src/mococrw/openssl_wrap.h
```diff
@@ -1495,5 +1507,59 @@ SSL_BIGNUM_Ptr _BN_bin2bn(const uint8_t* data, size_t dataLen);
...
+                      long i,
+                      void* opaque_value,
+                      void (*f)(void),
+                      int cmd_optional);
+/**
+ * Issues a command to the engine \p e. Takes a string as input data.
+ */
+void _ENGINE_ctrl_cmd(ENGINE* e,
+                      std::string cmd_name,
```
garw: const ref?

src/mococrw/openssl_wrap.h
```diff
@@ -1495,5 +1507,59 @@ SSL_BIGNUM_Ptr _BN_bin2bn(const uint8_t* data, size_t dataLen);
...
+                      int cmd_optional);
+
+/**
+ * Load private key via Engine \p e. Mainly used with HSMs which are modelled as
+ * OpenSSL engines.
+ *
+ * @param e The engine for loading the private key.
+ * @param key_id The ID of the key.
+ */
+SSL_EVP_PKEY_Ptr _ENGINE_load_private_key(ENGINE* e, std::string key_id);
```
garw: const ref?

src/mococrw/openssl_wrap.h
```diff
@@ -1495,5 +1507,59 @@ SSL_BIGNUM_Ptr _BN_bin2bn(const uint8_t* data, size_t dataLen);
...
+
+/**
+ * Load private key via Engine \p e. Mainly used with HSMs which are modelled as
+ * OpenSSL engines.
+ *
+ * @param e The engine for loading the private key.
+ * @param key_id The ID of the key.
+ */
+SSL_EVP_PKEY_Ptr _ENGINE_load_public_key(ENGINE* e, std::string key_id);
```
garw: const ref?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+ */
+
+using P11_PKCS11_CTX_Ptr =
+        std::unique_ptr<PKCS11_CTX, P11Deleter<void, PKCS11_CTX, lib::LibP11::P11_PKCS11_CTX_free>>;
+
+/*
+ * The memory referred to by these pointer types live out of scope of their pointers.
+ * Therefore, simply wrap the C pointer, without use of smart pointer (this is a bit ugly).
+ */
+using P11_PKCS11_SLOT_PTR = PKCS11_SLOT *;
```
garw, line 83: Why don't we just keep using the pointer notation? What is the purpose of this alias? They're used inconsistently in the rest of this file anyways.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+ */
+void _PKCS11_CTX_load(PKCS11_CTX *ctx, std::string &module);
+
+class P11_SlotInfo
+{
+public:
+    PKCS11_SLOT *_slots;     // Array consisting of slot descriptors on the HSM.
+    unsigned int _numSlots;  // The size of the array, i.e., the number of slots.
+
+    explicit P11_SlotInfo();
```
garw: see previous on usage of `explicit` on empty param list.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Loads the PKCS11 module to use.
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ * @param module The PKCS#11 module filename.
+ */
+void _PKCS11_CTX_load(PKCS11_CTX *ctx, std::string &module);
+
+class P11_SlotInfo
```
garw: If everything is public, we coventionally use `struct` for data to demonstrate that this is rather plan data than actual functionality.

What is the purpose of this anyways? The raw pointer array looks potentially dangerous. Doesn't this class require a destructor?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+P11_PKCS11_TOKEN_PTR _PKCS11_getTokenFromSlot(PKCS11_SLOT *slot);
+
+/**
+ * Opens a sessions with the passed \p slot.
+ *
+ * @param slot The slot to open session with.
+ * @param rw Whether to open the session in read/write mode. Set to != 0
+ * if read/write mode is required.
+ */
+void _PKCS11_open_session(PKCS11_SLOT *slot, int rw);
```
garw: part of this layer should also be to eliminate insane datatype such as this `int`. The clean choise: enum with two elements. this creates visually appealing function calls such as `_PKCS11_open_session(slot, SessionMode::ReadOnly)`

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Performs a user login to the HSM.
+ *
+ * @param slot The slot to consider for login.
+ * @param pin The pin to perform the login.
+ *
+ * \note This function performs a user login and not an SO login.
+ */
+void _PKCS11_login(PKCS11_SLOT *slot, std::string pin);
```
garw: const ref?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+ * Stores a private key inside the HSM.
+ *
+ * @param token The token used to store the private key.
+ * @param pk The key to store.
+ * @param label The label of the key to store.
+ * @param id The ID of the key to store.
+ */
+void _PKCS11_store_private_key(PKCS11_TOKEN *token,
+                               EVP_PKEY *pk,
+                               std::string label,
```
garw: const ref?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Stores a public key inside the HSM.
+ *
+ * @param token The token used to store the public key.
+ * @param pk The key to store.
+ * @param label The label of the key to store.
+ * @param id The ID of the key to store.
+ */
+void _PKCS11_store_public_key(PKCS11_TOKEN *token, EVP_PKEY *pk, std::string label, std::string id);
```
garw: const ref?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+ * Generates a key inside the HSM.
+ *
+ * @param token The token used to generate the key.
+ * @param bits The size of the modulus in bits.
+ * @param label The label of the generated key.
+ * @param id The ID of the generated key.
+ *
+ * \note Due to limited support offered by LibP11, this function only generates RSA keys.
+ */
+void _PKCS11_generate_key(PKCS11_TOKEN *token,
```
garw, line 216: This looks like its only RSA keys. Can we make this clear in the name?

We need btw ECC keys for our use-cases. Having RSA is rather a nice-to-have and maybe good for testing and exploration but overall insufficient.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,223 @@
...
+                          unsigned int bits,
+                          std::string label,
```
garw: const ref?

src/openssl_wrap.cpp
```diff
@@ -1557,5 +1557,61 @@ std::vector<uint8_t> _EVP_derive_key(const EVP_PKEY *peerkey, const EVP_PKEY *ke
...
+                      std::string cmd_name,
+                      long i,
+                      void *opaque_value,
+                      void (*f)(void),
+                      int cmd_optional)
+{
+    OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_ENGINE_ctrl_cmd,
+                                  e,
+                                  /* yes, it's ugly but OpenSSL expect a pointer to unsigned char
+                                   * that is pointing to non-const */
```
garw: Did you check the code if it really doesn't change anything here?

src/openssl_wrap.cpp
```diff
@@ -1557,5 +1557,61 @@ std::vector<uint8_t> _EVP_derive_key(const EVP_PKEY *peerkey, const EVP_PKEY *ke
...
+                                  i,
+                                  opaque_value,
+                                  f,
+                                  cmd_optional);
+}
+
+void _ENGINE_ctrl_cmd(
+        ENGINE *e, std::string cmd_name, long i, std::string val, void (*f)(void), int cmd_optional)
+{
+    _ENGINE_ctrl_cmd(
```
garw: How sure are we that this is safe?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+namespace mococrw
+{
+namespace p11
+{
+std::string P11Exception::generateP11ErrorString()
+{
+    // Note, pkcs11_CTX_new() internally loads LibP11 error strings,
+    // i.e., it invokes ERR_load_PKCS11_strings(). We therefore
+    // don't trigger the loading ourselves.
+    auto error = openssl::lib::OpenSSLLib::SSL_ERR_get_error();
```
garw, line 46: Do we really get p11 errors via OpenSSL? Why is this here then a duplication of the existing OpenSSL error class?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+        ::cppc::CallCheckContext<::cppc::IsNotNegativeReturnCheckPolicy, P11ExceptionErrorPolicy>;
+
+P11_SlotInfo::P11_SlotInfo() : _slots(nullptr), _numSlots(0) {}
+
+P11_SlotInfo::P11_SlotInfo(PKCS11_SLOT *slots, unsigned int numSlots)
+        : _slots(slots), _numSlots(numSlots)
+{
+}
+
+static std::string parseID(std::string idHexString)
```
garw: const ref?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+    unsigned int nslotsp;
+    P11IsNonNegative::callChecked(lib::LibP11::P11_PKCS11_enumerate_slots, ctx, &slotsp, &nslotsp);
+
+    // Based on the information obtained by slot enumaration, we instantiate
+    // a corresponding P11_SlotInfo object as a return value.
+    P11_SlotInfo slotInfo(slotsp, nslotsp);
+    return slotInfo;
+}
+
+void _PKCS11_release_all_slots(PKCS11_CTX *ctx, P11_SlotInfo &slotInfo)
```
garw, line 134: Why does this function exist? Shouldn't this be the destructor of P11_SlotInfo?
In general, I'm not that happy about the fact that the API allows contexts to be deleted before the respective slotinfo. This seems unsafe. Maybe we should discuss how to avoid problems here.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+{
+    if (ctx == nullptr) {
```
garw, line 136: normally, calling `free` on a nullptr is considered safe and just does nothing.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+    P11IsNonNegative::callChecked(
+            lib::LibP11::P11_PKCS11_is_logged_in, slot, 0 /* Check normal user. */, &result);
+
+    // Result is 1 if logged in.
+    return result == 1;
+}
+
+void _PKCS11_store_private_key(PKCS11_TOKEN *token, EVP_PKEY *pk, std::string label, std::string id)
+{
+    std::string parsedStrID = parseID(id);
```
garw, line 232: use `auto` keyword in scenarios like this.

.github/workflows/ci.yml
```diff
@@ -9,6 +9,9 @@ on:
 env:
   DOCKER_TAG: buildenv
   DOCKER_FILE_PATH: ./docker-build-env/Dockerfile
+  TOKEN_LABEL: token-label
```
garw, line 19: In order to do proper testing, we would need to build the non-hsm version in a container that doesn't have the p11 libraries installed. Otherwise, you won't know if you code/build scripts silently assumes their existence in some cases.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+    unsigned char *parsedID =
+            reinterpret_cast<unsigned char *>(const_cast<char *>(parsedStrID.c_str()));
```
garw, line 234: Every const case please with comment and statement that you checked that this is just due to bad const-correctness in upstream API.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+                                  token,
+                                  pk,
+                                  const_cast<char *>(label.c_str()),
+                                  parsedID,
+                                  parsedStrID.size());
+}
+
+void _PKCS11_store_public_key(PKCS11_TOKEN *token, EVP_PKEY *pk, std::string label, std::string id)
+{
+    std::string parsedStrID = parseID(id);
```
garw: s.a.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,274 @@
...
+                                  parsedStrID.size());
+}
+
+void _PKCS11_generate_key(PKCS11_TOKEN *token, unsigned int bits, std::string label, std::string id)
+{
+    std::string parsedStrID = parseID(id);
+    unsigned char *parsedID =
+            reinterpret_cast<unsigned char *>(const_cast<char *>(parsedStrID.c_str()));
+
+    // Note, PKCS11_generate_key() function only generates RSA keys.
```
garw, line 270: Honestly, we don't need to create RSA keys, so I wonder how useful adding a deprecated function is really is for us.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Returns the name of the HSM driver.
+     */
+    virtual const std::string getName() = 0;
+
+    /**
+     * Re-Enumerate slots/tokens of the HSM. This is often done to update slot
+     * list after adding/removing tokens/slots.
+     */
+    virtual void reenumerateSlots() = 0;
```
garw: Why is this part of the public API? Looks very internal to me.

tests/unit/test_p11wrapper.cpp
```diff
@@ -0,0 +1,400 @@
+/*
```
garw, line 1: Only gave the tests a quick glance…

src/pkcs11_hsm.cpp
```diff
@@ -0,0 +1,112 @@
...
+    _PKCS11_login(slot, _pin);
+
+    return slot;
+}
+
+void Pkcs11Engine::finishOperation(P11_PKCS11_SLOT_PTR slot)
+{
+    // Perform log out.
+    _PKCS11_logout(slot);
+}
```
garw: This all looks like it should be a RAII object.

src/pkcs11_hsm.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+    P11_PKCS11_SLOT_PTR slot = startOperation();
+
+    // Get token to peform operation on.
+    P11_PKCS11_TOKEN_PTR token = _PKCS11_getTokenFromSlot(slot);
+
+    // Perform key storage.
+    _PKCS11_store_public_key(token, key, label, keyID);
+
+    finishOperation(slot);
```
garw, line 74: This pattern here intensifies my feeling from above.

tests/unit/test_p11wrapper.cpp
```diff
@@ -0,0 +1,400 @@
...
+    void SetUp() override;
+    void TearDown() override;
+
+protected:
+    LibP11Mock &_mock() const { return LibP11MockManager::getMockInterface(); }
+};
+
+namespace testutils
+{
+PKCS11_CTX *somePKCS11CtxPtr()
```
garw, line 59: purpose?

garw (changes requested):
See detailed comments. 
We also probably need some tests for the functionality that is added by the HSM classes?

Moreover, please be more verbose with your commit message: What does this change do and for which reasons.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

examples/CMakeLists.txt
johnfxgalea-e: Yeah, makes sense...So the approach that I think is best is to create a FindMoCOCrW CMake script to expose such a flag to users.

src/mococrw/hsm.h
johnfxgalea-e: Rationale:  these fields can still be accessible from (and very likely will be required by) sub-classes that may be added in the future.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done

src/mococrw/p11_wrap.h
johnfxgalea-e: Done

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Sure, but it does not really make sense to have ctx NULL here. I added an exception raise because such a case would be strange.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

tests/unit/test_p11wrapper.cpp
johnfxgalea-e: That's fair. Will be breaking this PR down to smaller multiple one anyways...

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done!

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: This is a pure function which renders its class (i.e., HSM) as abstract and is overridden by subclasses of HSM.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: I concur! Done.

cmake/AddressSanitizerTarget.cmake
johnfxgalea-e: Done

cmake/AddressSanitizerTarget.cmake
johnfxgalea-e: Done.

.github/workflows/ci.yml
johnfxgalea-e: No, it was incorrect before according to my current reading. In order to trigger ctests with appropriate asan options, you need to trigger tests at build time. See `add_custom_target(asan .. `  AddressSanitizerTarget.cmake.

.github/workflows/ci.yml
johnfxgalea-e: Done.

.github/workflows/ci.yml
johnfxgalea-e: I circumvented this by setting up ASAN/LSAN options via Docker's -env param.

Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: ECC is out-of-scope for this PR (it is already big enough). It will be addressed in the next one.

src/p11_wrap.cpp
johnfxgalea-e: From above: 

> ECC is out-of-scope for this PR (it is already big enough). It will be addressed in the next one.

tests/unit/test_p11wrapper.cpp
johnfxgalea-e: Since we are mocking, this code returns a dummy pointer to a PKCS11_CTX value.

.github/workflows/ci.yml
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Indeed! LibP11 registers error strings via OpenSSL's ERR_load_strings() function. See https://github.com/OpenSC/libp11/blob/d5e07497e915c339f4c9e7534750eb4d8309b99b/src/p11_err.c#L68

examples/CMakeLists.txt
johnfxgalea-e: For now, I added a TODO comment. Punting this for a future PR.

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: It was more of a means to avoid using directly LibP11 definitions at upper levels of the MoCOCrW's code-base.

What do you mean by:

> They're used inconsistently in the rest of this file anyways.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: That function is internal. The user leverages the KeySpec API.

johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: I ended up removing this function completely for simplicity. Done.

PR #107, opened 2022-05-12
Apply clang-format to MoCOCrW

johnfxgalea-e:
Applies Clang-format to the MoCOCrW code base.

Adds a CI check to ensure future code changes conform to the required format.


.github/workflows/ci.yml
```diff
@@ -25,6 +25,15 @@ jobs:
         run: |
           docker build -t buildenv - < docker-build-env/Dockerfile
 
+      - name: "Check code format"
+        run: |
+          clang-format --version
```
wusto: are you sure that you want to run this outside of the container?
Is clang-format available there?

docker-build-env/Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM ubuntu:focal
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-recommends install \
         ca-certificates \
         clang \
+        clang-format \
```
wusto: Why do you add clang-format to the container, if clang-format isn't ran in the container?

wusto (changes requested):
hmm. Seems like github executes a first workflow after it is integrated. :-/
See https://github.com/bmwcarit/MoCOCrW/actions/runs/2314404295/workflow

I guess in the current state your workflow is not executable.

.github/workflows/ci.yml
```diff
@@ -25,6 +25,11 @@ jobs:
         run: |
           docker build -t buildenv - < docker-build-env/Dockerfile
 
+      - name: "Check code format"
+        run: |
+          docker run --rm -u "$UID" -v "$PWD:/src:rw" -w /src buildenv bash -c \
```
wusto: Just checked the output of the jobs.

Seems like this is ran for every combination of compiler and build_type.
I guess running it once should be enough.

wusto (changes requested):

.github/workflows/ci.yml
johnfxgalea-e: Done.

scripts/check_format.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-clang-format --version
-find . -iname *.h -o -iname *.cpp | xargs clang-format -i
+clang-format-10 --version
+find . -iname *.h -o -iname *.cpp | xargs clang-format-10 -i
```
wusto, line 4: there is one .hpp file.
Do we want to check this file, too?

.github/workflows/ci.yml
```diff
@@ -7,6 +7,24 @@ on:
   pull_request:
 
 jobs:
+  check-format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout"
```
wusto, line 17: lot's of code duplication.

As I am not an expert on github CI: Can you please check if there is a better option?
E.g. defining a parent job.

wusto (changes requested):

scripts/check_format.sh
johnfxgalea-e: No. That is because the clang-format version differed from previous use (it is fine now).

.github/workflows/ci.yml
johnfxgalea-e: I can do something. :eyes:

johnfxgalea-e: Done.

wusto (approved):

wusto:
I ran clang-format-10 on the openssl1.1 branch.
The result does not differ from the changes of John.

There is only one tiny difference
```
diff --git a/tests/unit/test_distinguished_name.cpp b/tests/unit/test_distinguished_name.cpp
index 8d7c322..d433e7c 100644
--- a/tests/unit/test_distinguished_name.cpp
+++ b/tests/unit/test_distinguished_name.cpp
@@ -51,9 +51,7 @@ void DistinguishedNameTest::SetUp()
     _x509Name = _X509_NAME_new();
 }
-void DistinguishedNameTest::TearDown()
-{ /* intentionally empty */
-}
+void DistinguishedNameTest::TearDown() { /* intentionally empty */ }
 TEST_F(DistinguishedNameTest, createDistinguishedName)
 { 
```
So still approved :)

PR #108, opened 2022-05-30
PKCS#11 testing infrastrucuture

istepic:
Add packages to the docker container that are necessary to test
HSM functionality (provided by SoftHSM2) from MoCOCrW
(MoCOCrW -> OpenSSL -> libp11 engine -> SoftHSM2)

Create keys inside SoftHSM2 with docker-build-env/entrypoint.sh
script. This will probably need expansion as test base grows.

### DIFF
--- a/docker-build-env/Dockerfile
+++ b/docker-build-env/Dockerfile
@@ -10,9 +10,20 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-rec
         git \
         googletest \
         libboost-all-dev \
+        # libp11 engine
+        libengine-pkcs11-openssl \
+        # headers for p11 engine
+        # TODO: this may not be necessary if we don't use libp11 directly (only though openssl)
+        libp11-dev \
+        # softhsm2 module
+        libsofthsm2 \
         libssl-dev \
         make \
         ninja-build \
+        # for pkcs11-tool which we use to create keys in token
+        opensc \
+        # p11-kit-modules allows loading of libp11 engine without having to edit openssl.cnf
+        p11-kit-modules \
         pkg-config \
         wget \
         && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


docker-build-env/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# Initialize softhsm and create a key
```
johnfxgalea-e: This seems like test logic to me. I don't think the user would expect a key to be created when the Docker image is created.

docker-build-env/Dockerfile
```diff
@@ -13,4 +13,21 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-rec
         ninja-build \
         pkg-config \
         wget \
+        # headers for p11 engine
+        # TODO: this may not be necessary if we don't use libp11 directly (only though openssl)
+        libp11-dev \
```
johnfxgalea-e, line 17: Nit: Dependencies are listed in alphabetical order.

docker-build-env/Dockerfile
```diff
@@ -13,4 +13,21 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-rec
...
+        # softhsm2 module
+        libsofthsm2 \
+        # sofhsm2 tooling for softhsm initialization
+        softhsm2 \
```
johnfxgalea-e: Doesn't this come with libsofthsm2?

docker-build-env/Dockerfile
istepic: Done.

docker-build-env/entrypoint.sh
istepic: Done.

docker-build-env/Dockerfile
istepic: Done.

johnfxgalea-e (approved):
LGTM

johnfxgalea-e:
@istepic Docker build has been separated into its own action file as a result of a recent commit. Fix conflicts by modifying `./.github/actions/build-docker`, etc..

johnfxgalea-e (approved):

miroshko (approved):

PR #109, opened 2022-06-10
Adds FindLibP11.cmake

johnfxgalea-e:
Adds FindLibP11.cmake so that find_package() may be invoked for the LibP11 library.

### DIFF
--- a/cmake/FindLibP11.cmake
+++ b/cmake/FindLibP11.cmake
@@ -1,0 +1,63 @@
+# - FindLibP11
+# Find the LibP11 Library.
+#
+# Configuration:
+#  LIBP11_ROOT_DIR     - Root directory containing the installation of LibP11.
+#                        Used as a hint to find the library.
+#
+# This module defines the following variables:
+#  LIBP11_FOUND        - If LibP11 is available
+#  LIBP11_INCLUDE_DIR  - The include directory of LibP11.
+#  LIBP11_LIBRARY      - The LibP11 Library.
+#
+# This module defines the following `IMPORTED` targets:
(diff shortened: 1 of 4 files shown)


garw (approved):
Looks roughly okay but this needs the be an optional dependency that can be disabled (or rather should be explicitly enabled). MoCoCrW needs to be compilable if there is no libp11 present on the system.

garw (changes requested):
err… wrong button :/

johnfxgalea-e:
> Looks roughly okay but this needs the be an optional dependency that can be disabled (or rather should be explicitly enabled). MoCoCrW needs to be compilable if there is no libp11 present on the system.

Done.

garw (approved):
lg2m

cmake/FindLibP11.cmake
```diff
@@ -0,0 +1,63 @@
...
+# Configuration:
+#  LIBP11_ROOT_DIR     - Root directory containing the installation of LibP11.
+#                        Used as a hint to find the library.
+#
+# This module defines the following variables:
+#  LIBP11_FOUND        - If LibP11 is available
+#  LIBP11_INCLUDE_DIR  - The include directory of LibP11.
+#  LIBP11_LIBRARY      - The LibP11 Library.
+#
+# This module defines the following`IMPORTED` targets:
```
blenk92: looks like there should be a white space between ``following`IMPORTED` ``

johnfxgalea-e: Done.

blenk92 (approved):

PR #111, opened 2022-07-06
Create two Docker images to test MoCOCrW with and without HSM support separately.

johnfxgalea-e:
In order to test properly MoCOCrW's builds with and without HSM support, we introduce two docker images and update our CI. In particular, we want to avoid cases where a build, which is intended to be without HSM support, incorrectly relies on an HSM dependency to successfully compile. 

Note, Dockerfile-With-HSM is based on Dockerfile, specifying solely PKCS11 dependencies.

### DIFF
--- a/docker-build-env/Dockerfile
+++ b/docker-build-env/Dockerfile
@@ -10,20 +10,11 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-rec
         git \
         googletest \
         libboost-all-dev \
-        # libp11 engine
-        libengine-pkcs11-openssl \
-        # headers for p11 engine
-        # TODO: this may not be necessary if we don't use libp11 directly (only though openssl)
-        libp11-dev \
-        # softhsm2 module
-        libsofthsm2 \
         libssl-dev \
+        # For llvm-symbolizer (used by ASAN).
+        llvm \
         make \
         ninja-build \
-        # for pkcs11-tool which we use to create keys in token
-        opensc \
-        # p11-kit-modules allows loading of libp11 engine without having to edit openssl.cnf
-        p11-kit-modules \
         pkg-config \
         wget \
         && rm -rf /var/lib/apt/lists/*
--- a/docker-build-env/Dockerfile
+++ b/docker-build-env/Dockerfile
@@ -11,8 +11,6 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-rec
         googletest \
         libboost-all-dev \
         libssl-dev \
-        # For llvm-symbolizer (used by ASAN).
-        llvm \
         make \
         ninja-build \
         pkg-config \
--- a/docker-build-env/Dockerfile-With-HSM
+++ b/docker-build-env/Dockerfile-With-HSM
@@ -1,0 +1,15 @@
+FROM buildenv
+
+# Install MoCOCrW dependencies (except OpenSSL)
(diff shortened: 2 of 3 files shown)


docker-build-env/Dockerfile-With-HSM
```diff
@@ -0,0 +1,14 @@
...
+        # libp11 engine
+        libengine-pkcs11-openssl \
+        # headers for p11 engine
+        libp11-dev \
+        # for pkcs11-tool which we use to create keys in token
+        opensc \
+        # p11-kit-modules allows loading of libp11 engine without having to edit openssl.cnf
+        p11-kit-modules \
+        # softhsm2: includes both softhsm2-util and libsofthsm2
+        softhsm2
```
wusto: Please remove the package manager's files at the end of installation to reduce the container size.
```suggestion
        softhsm2 \
        && rm -rf /var/lib/apt/lists/*
```

wusto (changes requested):
nitpick. Otherwise looks good to me.

docker-build-env/Dockerfile-With-HSM
johnfxgalea-e: Done.

wusto (approved):

docker-build-env/Dockerfile
```diff
@@ -10,20 +10,11 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-rec
...
-        # libp11 engine
-        libengine-pkcs11-openssl \
-        # headers for p11 engine
-        # TODO: this may not be necessary if we don't use libp11 directly (only though openssl)
-        libp11-dev \
-        # softhsm2 module
-        libsofthsm2 \
         libssl-dev \
+        # For llvm-symbolizer (used by ASAN).
+        llvm \
```
istepic: Any specific reason why we need this now?

johnfxgalea-e: No, I can push this to another PR. Fixing.

johnfxgalea-e: Done.

istepic:
> In order to test properly MoCOCrW's builds with and without HSM support, we introduce two docker images and update our CI. In particular, we want to avoid cases where a build, which is intended to be without HSM support, incorrectly relies on an HSM dependency to successfully compile.
> 
> Note, Dockerfile-With-HSM is based on Dockerfile, specifying solely PKCS11 dependencies.

Can't this be somehow resolved with CMake?

johnfxgalea-e:
> > In order to test properly MoCOCrW's builds with and without HSM support, we introduce two docker images and update our CI. In particular, we want to avoid cases where a build, which is intended to be without HSM support, incorrectly relies on an HSM dependency to successfully compile.
> > Note, Dockerfile-With-HSM is based on Dockerfile, specifying solely PKCS11 dependencies.
> 
> Can't this be somehow resolved with CMake?

Indeed, one would configure CMake to compile MoCOCrW with or without HSM support. Specifically, this is achieved via the `-DHSM_ENABLED` option. 

However, we want to have a high-level of confidence that the build system does not incorrectly rely on an HSM dependency when the above option is set to `OFF`. Don't forget, prior to this PR, LibP11 and other HSM dependencies are always installed by the CI, regardless whether the CI builds MoCOCrW with/without HSM support. Having two docker images seems to be easiest approach to check this, imo.

istepic (approved):

garw (approved):
lg2m

PR #113, opened 2022-07-06
Add P11 API

johnfxgalea-e:
In effort to support the PKCS11 HSM interface, this PR adds P11 API function wrappers so that they may be used by MoCOCrW.

The p11_lib.h and p11_wrap.h are intermediary layers; they wrap LibP11's C API and particularly `p11_wrap.h`, is intended to be used internally by MoCOCrW. Crucially, their functions are not intended to be called directly by the user.

Tests are also included with this PR.

### DIFF
--- a/src/mococrw/p11_lib.h
+++ b/src/mococrw/p11_lib.h
@@ -1,0 +1,85 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2022 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+
+extern "C" {
+#include <libp11.h>
+}
+
+namespace mococrw
+{
+namespace p11
+{
+namespace lib
+{
+/**
+ * Thin wrapper around the "naked" P11 library (namely libp11).
+ *
+ * Prefix all method names with 'P11_' to distinguish them from their
+ * LibP11 counterparts.
+ */
+class LibP11
+{
+public:
+    /* Initialisation */
+    static PKCS11_CTX *P11_PKCS11_CTX_new(void) noexcept;
+    static int P11_PKCS11_CTX_load(PKCS11_CTX *ctx, const char *ident) noexcept;
+
+    /* Slot and Token Management */
+    static int P11_PKCS11_enumerate_slots(PKCS11_CTX *ctx,
+                                          PKCS11_SLOT **slotsp,
+                                          unsigned int *nslotsp) noexcept;
+    static void P11_PKCS11_release_all_slots(PKCS11_CTX *ctx,
+                                             PKCS11_SLOT *slots,
+                                             unsigned int nslots) noexcept;
+    static PKCS11_SLOT *P11_PKCS11_find_token(PKCS11_CTX *ctx,
+                                              PKCS11_SLOT *slots,
+                                              unsigned int nslots) noexcept;
+
+    /* Session and Login */
+    static int P11_PKCS11_open_session(PKCS11_SLOT *slot, int rw) noexcept;
+    static int P11_PKCS11_login(PKCS11_SLOT *slot, int so, const char *pin) noexcept;
+    static int P11_PKCS11_logout(PKCS11_SLOT *slot) noexcept;
+    static int P11_PKCS11_is_logged_in(PKCS11_SLOT *slot, int so, int *res) noexcept;
+
+    /* Key Management */
+    static int P11_PKCS11_store_private_key(PKCS11_TOKEN *token,
+                                            EVP_PKEY *pk,
+                                            char *label,
+                                            unsigned char *id,
+                                            size_t id_len) noexcept;
+    static int P11_PKCS11_store_public_key(PKCS11_TOKEN *token,
+                                           EVP_PKEY *pk,
+                                           char *label,
+                                           unsigned char *id,
+                                           size_t id_len) noexcept;
+    static int P11_PKCS11_generate_key(PKCS11_TOKEN *token,
+                                       int algorithm,
+                                       unsigned int bits,
+                                       char *label,
+                                       unsigned char *id,
+                                       size_t id_len) noexcept;
+
+    /* Clean-up Functions */
+    static void P11_PKCS11_CTX_unload(PKCS11_CTX *ctx) noexcept;
+    static void P11_PKCS11_CTX_free(PKCS11_CTX *ctx) noexcept;
+};
+}  // namespace lib
+}  // namespace p11
+}  // namespace mococrw
--- a/src/mococrw/p11_wrap.h
+++ b/src/mococrw/p11_wrap.h
@@ -18,6 +18,7 @@
  */
 #pragma once
 
+#include <boost/core/noncopyable.hpp>
 #include "openssl_wrap.h"
 #include "p11_lib.h"
 #include "util.h"
@@ -67,77 +68,75 @@ public:
      *
      */
     P11Exception() : OpenSSLException() {}
-
-    /**
-     * Error Strings.
-     *
-     * These are declared here to reduce string duplication as well as facilitate
-     * tests that check exceptions.
-     */
-    static const std::string nullCtxExceptionString;
-    static const std::string nullSlotListExceptionString;
-    static const std::string nullTokenExceptionString;
-    static const std::string mismatchLoginExceptionString;
 };
 
 /*
- * Wrap all the pointer-types returned by LibP11.
+ * Wrap all the pointer-types returned by LibP11. The following
+ * shared pointers have custom deleters, which are specified upon their
+ * construction.
  */
 
-using P11_PKCS11_CTX_Ptr = std::shared_ptr<PKCS11_CTX>;
-using P11_PKCS11_SLOT_LIST_PTR = std::shared_ptr<PKCS11_SLOT>;
+using P11_PKCS11_CTX_SharedPtr = std::shared_ptr<PKCS11_CTX>;
+using P11_PKCS11_SLOT_LIST_SharedPtr = std::shared_ptr<PKCS11_SLOT>;
 
 /* The memory referred to by these pointer types live out-of-scope of their pointers.
  * Therefore, we leverage the aliasing constructor of shared pointers when the following
  * smart pointer types are created.
  */
-using P11_PKCS11_SLOT_PTR = std::shared_ptr<PKCS11_SLOT>;
-using P11_PKCS11_TOKEN_PTR = std::shared_ptr<PKCS11_TOKEN>;
+using P11_PKCS11_SLOT_SharedPtr = std::shared_ptr<PKCS11_SLOT>;
+using P11_PKCS11_TOKEN_SharedPtr = std::shared_ptr<PKCS11_TOKEN>;
 
 /**
  * Details information related to a slot list, obtained via enumeration of HSM.
  */
-class P11_SlotInfo
+class P11SlotInfo : private boost::noncopyable
 {
 public:
-    P11_SlotInfo(P11_PKCS11_CTX_Ptr ctx);
+    P11SlotInfo(P11_PKCS11_CTX_SharedPtr ctx);
 
-    struct SlotListDeleter
+    class SlotListDeleter
     {
-        SlotListDeleter(P11_PKCS11_CTX_Ptr ctx, int numSlots) : _ctx(ctx), _numSlots(numSlots) {}
+    public:
+        SlotListDeleter(P11_PKCS11_CTX_SharedPtr ctx, int numSlots) : _ctx(ctx), _numSlots(numSlots)
+        {
+        }
         void operator()(PKCS11_SLOT *);
 
-        P11_PKCS11_CTX_Ptr _ctx;
+    private:
+        P11_PKCS11_CTX_SharedPtr _ctx;
         int _numSlots;
     };
 
     /**
      * Finds the slot of a token on the HSM.
      */
-    P11_PKCS11_SLOT_PTR findSlot();
+    P11_PKCS11_SLOT_SharedPtr findSlot();
 
 private:
-    P11_PKCS11_CTX_Ptr _ctx;             // The context associated with the slot information.
-    P11_PKCS11_SLOT_LIST_PTR _slotList;  // Array consisting of slot descriptors on the HSM.
-    unsigned int _numSlots;              // The size of the array, i.e., the number of slots.
+    P11_PKCS11_CTX_SharedPtr _ctx;             // The context associated with the slot information.
+    P11_PKCS11_SLOT_LIST_SharedPtr _slotList;  // Array consisting of slot descriptors on the HSM.
+    unsigned int _numSlots;                    // The size of the array, i.e., the number of slots.
 };
+
+// Shorthand for shared pointer of type P11SlotInfo
+using P11SlotInfo_SharedPtr = std::shared_ptr<P11SlotInfo>;
 
 /**
  * Session mode denotes whether to open a session in read/write mode.
  */
 enum SessionMode { ReadOnly = 0, ReadWrite = 1 };
 
-class P11_Session
+class P11Session : private boost::noncopyable
 {
 public:
-    P11_Session(P11_PKCS11_SLOT_PTR slot, const std::string &pin, SessionMode mode);
-    ~P11_Session();  // After session is made out-of-scope, trigger logout.
+    P11Session(P11_PKCS11_SLOT_SharedPtr slot, const std::string &pin, SessionMode mode);
+    ~P11Session();  // After session is made out-of-scope, trigger logout.
 
-    /* Returns the token descriptor associated with the passed slot. */
-    P11_PKCS11_TOKEN_PTR getTokenFromSlot();
+    /* Returns the token descriptor of the slot associated with the session. */
+    P11_PKCS11_TOKEN_SharedPtr getTokenFromSessionSlot();
 
 private:
-    P11_PKCS11_SLOT_PTR _slot;
+    P11_PKCS11_SLOT_SharedPtr _slot;
 
     /* Opens a sessions with the passed slot. */
     void openSession(SessionMode mode);
@@ -148,10 +147,10 @@ private:
     /* Performs a user logout from the HSM. This function only performs a user logout, and not an SO
      * logout. */
     void logout();
-
-    /* Returns true if a user is currently logged in. */
-    bool isLoggedIn();
 };
+
+// Shorthand for shared pointer of type P11SlotInfo
+using P11Session_SharedPtr = std::shared_ptr<P11Session>;
 
 /* Below is the "wrapped" LibP11 library. By convention, all functions start with an
  * underscore to visually distinguish them from the methods of the class P11Lib and
@@ -164,7 +163,7 @@ private:
  * @param module The PKCS#11 module filename.
  * @return A new PKCS11 context.
  */
-P11_PKCS11_CTX_Ptr _PKCS11_CTX_create(const std::string &module);
+P11_PKCS11_CTX_SharedPtr _PKCS11_CTX_create(const std::string &module);
 
 /**
  * Stores a private key inside the HSM.
@@ -174,7 +173,7 @@ P11_PKCS11_CTX_Ptr _PKCS11_CTX_create(const std::string &module);
  * @param label The label of the key to store.
  * @param id The ID of the key to store.
  */
-void _PKCS11_store_private_key(P11_Session &session,
+void _PKCS11_store_private_key(P11Session_SharedPtr session,
                                EVP_PKEY *pk,
                                const std::string &label,
                                const std::string &id);
@@ -187,25 +186,26 @@ void _PKCS11_store_private_key(P11_Session &session,
  * @param label The label of the key to store.
  * @param id The ID of the key to store.
  */
-void _PKCS11_store_public_key(P11_Session &session,
+void _PKCS11_store_public_key(P11Session_SharedPtr session,
                               EVP_PKEY *pk,
                               const std::string &label,
                               const std::string &id);
 
 /**
- * Generates a key inside the HSM.
+ * Generates an RSA key inside the HSM.
  *
  * @param session The session used to generate the key.
  * @param bits The size of the modulus in bits.
  * @param label The label of the generated key.
  * @param id The ID of the generated key.
  *
- * \note Due to limited support offered by LibP11, this function only generates RSA keys.
+ * \note Due to limited support offered by LibP11, this function only
+ * generates RSA keys.
  */
-void _PKCS11_generate_key(P11_Session &session,
-                          unsigned int bits,
-                          const std::string &label,
-                          const std::string &id);
+void _PKCS11_generate_rsa_key(P11Session_SharedPtr session,
+                              unsigned int bits,
+                              const std::string &label,
+                              const std::string &id);
 
 }  // namespace p11
 }  // namespace mococrw
--- a/src/mococrw/p11_wrap.h
+++ b/src/mococrw/p11_wrap.h
@@ -110,7 +110,7 @@ public:
     /**
      * Finds the slot of a token on the HSM.
      */
-    P11_PKCS11_SLOT_SharedPtr findSlot();
+    P11_PKCS11_SLOT_SharedPtr findFirstSlot();
 
 private:
     P11_PKCS11_CTX_SharedPtr _ctx;             // The context associated with the slot information.
--- a/src/mococrw/p11_wrap.h
+++ b/src/mococrw/p11_wrap.h
@@ -18,6 +18,7 @@
  */
 #pragma once
 
+#include "openssl_wrap.h"
 #include "p11_lib.h"
 #include "util.h"
 
@@ -26,9 +27,9 @@ namespace mococrw
 namespace p11
 {
 /**
- * Template to wrap OpenSSLs "_free" functions
+ * Template to wrap LibP11 free/destroy functions
  * into a functor so that a std::unique_ptr
- * can use them.
+ * can use them upon being out-of-scope.
  */
 template <class R, class P, R(Func)(P *)>
 struct P11Deleter
@@ -45,43 +46,112 @@ struct P11Deleter
 /**
  * An Exception for LibP11 errors.
  *
- * This exception is thrown by all methods when an OpenSSL error occurs.
+ * This exception is thrown by all methods when a LibP11 error occurs.
  */
-class P11Exception final : public std::exception
+class P11Exception final : public openssl::OpenSSLException
 {
+    // Note, pkcs11_CTX_new() internally loads LibP11 error strings,
+    // i.e., it invokes ERR_load_PKCS11_strings(). We therefore
+    // don't trigger the loading ourselves.
+
 public:
     template <class StringType>
-    explicit P11Exception(StringType &&message) : _message{std::forward<StringType>(message)}
+    explicit P11Exception(StringType &&message) : OpenSSLException(message)
     {
     }
 
     /**
      * Generate an exception with the defalt LibP11 error-string
-     * as message.
+     * as message. Internally, LibP11 uses OpenSSL's error handling framework,
+     * and therefore we simply extend OpenSSLException.
      *
      */
-    P11Exception() : _message{generateP11ErrorString()} {}
+    P11Exception() : OpenSSLException() {}
 
-    const char *what() const noexcept override { return _message.c_str(); }
-
-private:
-    static std::string generateP11ErrorString();
-    std::string _message;
+    /**
+     * Error Strings.
+     *
+     * These are declared here to reduce string duplication as well as facilitate
+     * tests that check exceptions.
+     */
+    static const std::string nullCtxExceptionString;
+    static const std::string nullSlotListExceptionString;
+    static const std::string nullTokenExceptionString;
+    static const std::string mismatchLoginExceptionString;
 };
 
 /*
  * Wrap all the pointer-types returned by LibP11.
  */
 
-using P11_PKCS11_CTX_Ptr =
-        std::unique_ptr<PKCS11_CTX, P11Deleter<void, PKCS11_CTX, lib::LibP11::P11_PKCS11_CTX_free>>;
+using P11_PKCS11_CTX_Ptr = std::shared_ptr<PKCS11_CTX>;
+using P11_PKCS11_SLOT_LIST_PTR = std::shared_ptr<PKCS11_SLOT>;
 
-/*
- * The memory referred to by these pointer types live out of scope of their pointers.
- * Therefore, simply wrap the C pointer, without use of smart pointer (this is a bit ugly).
+/* The memory referred to by these pointer types live out-of-scope of their pointers.
+ * Therefore, we leverage the aliasing constructor of shared pointers when the following
+ * smart pointer types are created.
  */
-using P11_PKCS11_SLOT_PTR = PKCS11_SLOT *;
-using P11_PKCS11_TOKEN_PTR = PKCS11_TOKEN *;
+using P11_PKCS11_SLOT_PTR = std::shared_ptr<PKCS11_SLOT>;
+using P11_PKCS11_TOKEN_PTR = std::shared_ptr<PKCS11_TOKEN>;
+
+/**
+ * Details information related to a slot list, obtained via enumeration of HSM.
+ */
+class P11_SlotInfo
+{
+public:
+    P11_SlotInfo(P11_PKCS11_CTX_Ptr ctx);
+
+    struct SlotListDeleter
+    {
+        SlotListDeleter(P11_PKCS11_CTX_Ptr ctx, int numSlots) : _ctx(ctx), _numSlots(numSlots) {}
+        void operator()(PKCS11_SLOT *);
+
+        P11_PKCS11_CTX_Ptr _ctx;
+        int _numSlots;
+    };
+
+    /**
+     * Finds the slot of a token on the HSM.
+     */
+    P11_PKCS11_SLOT_PTR findSlot();
+
+private:
+    P11_PKCS11_CTX_Ptr _ctx;             // The context associated with the slot information.
+    P11_PKCS11_SLOT_LIST_PTR _slotList;  // Array consisting of slot descriptors on the HSM.
+    unsigned int _numSlots;              // The size of the array, i.e., the number of slots.
+};
+
+/**
+ * Session mode denotes whether to open a session in read/write mode.
+ */
+enum SessionMode { ReadOnly = 0, ReadWrite = 1 };
+
+class P11_Session
+{
+public:
+    P11_Session(P11_PKCS11_SLOT_PTR slot, const std::string &pin, SessionMode mode);
+    ~P11_Session();  // After session is made out-of-scope, trigger logout.
+
+    /* Returns the token descriptor associated with the passed slot. */
+    P11_PKCS11_TOKEN_PTR getTokenFromSlot();
+
+private:
+    P11_PKCS11_SLOT_PTR _slot;
+
+    /* Opens a sessions with the passed slot. */
+    void openSession(SessionMode mode);
+
+    /* Performs a user login to the HSM. This function performs a user login and not an SO login. */
+    void login(const std::string &pin);
+
+    /* Performs a user logout from the HSM. This function only performs a user logout, and not an SO
+     * logout. */
+    void logout();
+
+    /* Returns true if a user is currently logged in. */
+    bool isLoggedIn();
+};
 
 /* Below is the "wrapped" LibP11 library. By convention, all functions start with an
  * underscore to visually distinguish them from the methods of the class P11Lib and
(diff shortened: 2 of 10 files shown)


istepic:
I think it would make more sense if you pushed a PR that integrates ENGINE_* API to MoCOCrW. For example, how can we create AsymmetricKeypair MoCOCrW object if we use HSM as a source. Once we get an understanding of how that can be done, then we can fill the gaps with this PR.

johnfxgalea-e:
> I think it would make more sense if you pushed a PR that integrates ENGINE_* API to MoCOCrW. For example, how can we create AsymmetricKeypair MoCOCrW object if we use HSM as a source. Once we get an understanding of how that can be done, then we can fill the gaps with this PR.

Currently, key generation is achieved via LibP11, and not the ENGINE_* API. However, I think I can add another sub PR so you get a better idea. I'm on it!

istepic:
> > I think it would make more sense if you pushed a PR that integrates ENGINE_* API to MoCOCrW. For example, how can we create AsymmetricKeypair MoCOCrW object if we use HSM as a source. Once we get an understanding of how that can be done, then we can fill the gaps with this PR.
> 
> Currently, key generation is achieved via LibP11, and not the ENGINE_* API. However, I think I can add another sub PR so you get a better idea. I'm on it!

Sorry for not being clear, I was speaking about just loading the key from HSM with the ENGINE_*. In that case, some parts of this PR might be unnecessary.
Edit: Sorry, now I see key loading is not part of this PR so that's a bad example.

johnfxgalea-e:
> > > I think it would make more sense if you pushed a PR that integrates ENGINE_* API to MoCOCrW. For example, how can we create AsymmetricKeypair MoCOCrW object if we use HSM as a source. Once we get an understanding of how that can be done, then we can fill the gaps with this PR.
> > 
> > 
> > Currently, key generation is achieved via LibP11, and not the ENGINE_* API. However, I think I can add another sub PR so you get a better idea. I'm on it!
> 
> Sorry for not being clear, I was speaking about just loading the key from HSM with the ENGINE_*. In that case, some parts of this PR might be unnecessary. Edit: Sorry, now I see key loading is not part of this PR so that's a bad example.

Does this other PR help? : https://github.com/bmwcarit/MoCOCrW/pull/116

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * Loads the PKCS11 module to use.
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ * @param module The PKCS#11 module filename.
+ */
+void _PKCS11_CTX_load(PKCS11_CTX *ctx, const std::string &module);
+
+struct P11_SlotInfo
```
wusto: Don't like this approach.

This requires the user to think of freeing memory, which is proofed to be error prone.
Please add a destructor, which frees the memory, when the object goes out of scope.

Alternatively wrap it in a unique pointer, which calls the release_all_slots method.

if you compare it with openssl_wrap.h, there no function ever returns an object, which can leak memory.

I guess the common way in this library is to return a smart pointer which frees the memory when going out of scope.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+ */
+
+using P11_PKCS11_CTX_Ptr =
+        std::unique_ptr<PKCS11_CTX, P11Deleter<void, PKCS11_CTX, lib::LibP11::P11_PKCS11_CTX_free>>;
+
+/*
+ * The memory referred to by these pointer types live out of scope of their pointers.
+ * Therefore, simply wrap the C pointer, without use of smart pointer (this is a bit ugly).
+ */
+using P11_PKCS11_SLOT_PTR = PKCS11_SLOT *;
```
wusto: Not sure, if this improves readability.
At least I would assume something like a smart pointer reading "P11_PKCS11_SLOT_PTR".

And whats the rational not assigning it to a smart pointer?
Assigning a noop function as destructor should do the trick and helps to unify the interface for users.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+    static std::string generateP11ErrorString();
+    std::string _message;
+};
+
+/*
+ * Wrap all the pointer-types returned by LibP11.
+ */
+
+using P11_PKCS11_CTX_Ptr =
+        std::unique_ptr<PKCS11_CTX, P11Deleter<void, PKCS11_CTX, lib::LibP11::P11_PKCS11_CTX_free>>;
```
wusto: Is P11_PKCS11_CTX_free enough or do you need to unload it before?

I am wondering if the unload function needs to be public for the p11_wrap?
Is there any use-case you can think of?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+using P11_PKCS11_SLOT_PTR = PKCS11_SLOT *;
+using P11_PKCS11_TOKEN_PTR = PKCS11_TOKEN *;
+
+/* Below is the "wrapped" LibP11 library. By convention, all functions start with an
+ * underscore to visually distinguish them from the methods of the class P11Lib and
+ * from the native LibP11 methods.
+ */
+
+/**
+ * Returns a new PKCS11 context for HSM interaction.
```
wusto: There is also a return statement for doxygen:
https://www.doxygen.nl/manual/commands.html#cmdreturn

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+ */
+void _PKCS11_release_all_slots(PKCS11_CTX *ctx, P11_SlotInfo &slotInfo);
+
+/**
+ * Finds a token on the HSM according to the passed slot information \p slotInfo.
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ * @param slotInfo Slot information obtained via _PKCS11_enumerate_slots().
+ */
+P11_PKCS11_SLOT_PTR _PKCS11_find_token(PKCS11_CTX *ctx, P11_SlotInfo &slotInfo);
```
wusto: Please return a smart pointer

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * Returns the token descriptor associated with \p slot.
+ *
+ * @param slot The slot of the returned token.
+ */
+P11_PKCS11_TOKEN_PTR _PKCS11_getTokenFromSlot(PKCS11_SLOT *slot);
```
wusto: s.a.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+#include <cppc/checkcall.hpp>
+#include "mococrw/openssl_wrap.h"
+#include "mococrw/p11_wrap.h"
+
+namespace mococrw
+{
+namespace p11
+{
+std::string P11Exception::generateP11ErrorString()
```
wusto: Duplication from std::string OpenSSLException::generateOpenSSLErrorString()

What's the benefit of having this function twice?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+     * Generate an exception with the defalt LibP11 error-string
+     * as message.
+     *
+     */
+    P11Exception() : _message{generateP11ErrorString()} {}
+
+    const char *what() const noexcept override { return _message.c_str(); }
+
+private:
+    static std::string generateP11ErrorString();
```
wusto: as stated in the cpp file, this the same as in openssl.
Is it possible to borrow the function (e.g. as a friend class)?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,227 @@
...
+    P11_SlotInfo();
+    P11_SlotInfo(PKCS11_SLOT *slots, unsigned int numSlots);
+};
+
+/**
+ * Obtains information of all current slots on the HSM.
+ *
+ * @param ctx The current context obtained via _PKCS11_CTX_new().
+ */
+P11_SlotInfo _PKCS11_enumerate_slots(PKCS11_CTX *ctx);
```
wusto: To unify the interface within MoCOCrW I would prefer a smart pointer here.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,283 @@
...
+struct P11ExceptionErrorPolicy
+{
+    template <class Rv>
+    static void handleError(const Rv &);
+};
+
+/**
+ * Throw a P11 exception upon error.
+ *
+ * This method gets the error string and throws an exception
```
wusto: "gets" is misleading here.
I guess the better description would be:
P11Exception generates the error string from the openssl error string in the constructor.
```suggestion
 * This method fetches the error string from openssl in P11Exception's constructor and throws an exception
```

src/p11_wrap.cpp
```diff
@@ -0,0 +1,283 @@
...
+}
+
+void _PKCS11_release_all_slots(PKCS11_CTX *ctx, P11_SlotInfo &slotInfo)
+{
+    if (ctx == nullptr) {
+        throw P11Exception("NULL passed for context when releaseing slots");
+    }
+
+    if (slotInfo._slots == nullptr) {
+        throw P11Exception("NULL list when releaseing slots");
```
wusto:
```suggestion
        throw P11Exception("NULL list when releasing slots");
```

src/p11_wrap.cpp
```diff
@@ -0,0 +1,283 @@
...
+    // Based on the information obtained by slot enumaration, we instantiate
+    // a corresponding P11_SlotInfo object as a return value.
+    P11_SlotInfo slotInfo(slotsp, nslotsp);
+    return slotInfo;
+}
+
+void _PKCS11_release_all_slots(PKCS11_CTX *ctx, P11_SlotInfo &slotInfo)
+{
+    if (ctx == nullptr) {
+        throw P11Exception("NULL passed for context when releaseing slots");
```
wusto:
```suggestion
        throw P11Exception("NULL passed for context when releasing slots");
```

tests/unit/p11_lib_mock.h
```diff
@@ -0,0 +1,158 @@
...
+    MOCK_METHOD1(P11_PKCS11_CTX_unload, void(PKCS11_CTX *ctx));
+    MOCK_METHOD1(P11_PKCS11_CTX_free, void(PKCS11_CTX *ctx));
+};
+
+/**
+ * Wrap instances of the LibP11Mock
+ * inside static members of this class.
+ *
+ * This gets rid of the need for a singleton and all
+ * the realted problems.
```
wusto:
```suggestion
 * the related problems.
```

tests/unit/p11_lib_mock.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+/**
+ * Provide implementations for the OpenSSLLib members that forward
+ * to the mock object.
+ *
+ * This translation unit is inserted into the bulid process for unit-tests
+ * via CMake. The library gets these definitions from a corresponding class
+ * in the src-dir.
+ */
+// TODO unit test the initialization
```
wusto: Either remove the comment of fix it.

tests/unit/test_p11wrapper.cpp
```diff
@@ -0,0 +1,403 @@
...
+{
+    P11_SlotInfo slotInfo(::testutils::somePKCS11SlotPtr(), 2);
+    EXPECT_THROW(_PKCS11_release_all_slots(nullptr, slotInfo), P11Exception);
+}
+
+/* Test _PKCS11_release_all_slots() with bad slot information. */
+TEST_F(P11WrapperTest, badSlotRelease2)
+{
+    P11_SlotInfo slotInfo;
+    EXPECT_THROW(_PKCS11_release_all_slots(::testutils::somePKCS11CtxPtr(), slotInfo),
```
wusto: Is there a way to check for the "right" exception to be thrown?
There is the ctx and slotInfo check.

tests/unit/test_p11wrapper.cpp
```diff
@@ -0,0 +1,403 @@
...
+/* Test _PKCS11_enumerate_slots with valid context. */
+TEST_F(P11WrapperTest, goodSlotEnumeration)
+{
+    EXPECT_CALL(_mock(), P11_PKCS11_enumerate_slots(::testutils::somePKCS11CtxPtr(), _, _))
+            .WillOnce(Return(0));
+
+    EXPECT_NO_THROW(_PKCS11_enumerate_slots(::testutils::somePKCS11CtxPtr()));
+}
+
+/* Test _PKCS11_release_all_slots() with NULL context. */
```
wusto: Code duplication.
Can you please provide the test parameters to a single function?

tests/unit/test_p11wrapper.cpp
```diff
@@ -0,0 +1,403 @@
...
+{
+    EXPECT_CALL(_mock(), P11_PKCS11_open_session(::testutils::somePKCS11SlotPtr(), 1))
+            .WillOnce(Return(-1));
+
+    EXPECT_THROW(_PKCS11_open_session(::testutils::somePKCS11SlotPtr(), SessionMode::ReadWrite),
+                 P11Exception);
+}
+
+/* Test successful _PKCS11_open_session(). */
+TEST_F(P11WrapperTest, goodOpenSession)
```
wusto: Please use parameters and avoid code duplication.

wusto (changes requested):
General:
Please try to parametrize your test functions instead of code duplication.
Please try to strictly stick to the interface covention of the openssl wrapper. Afaik there are no raw pointers returned anywhere.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

tests/unit/p11_lib_mock.h
johnfxgalea-e: Done.

tests/unit/p11_lib_mock.cpp
johnfxgalea-e: Done.

tests/unit/test_p11wrapper.cpp
johnfxgalea-e: I had a look at this... There is a kind of hack to do this: https://stackoverflow.com/questions/23270078/test-a-specific-exception-type-is-thrown-and-the-exception-has-the-right-propert

The approach revolves around catching the exception and checking the msg string, followed by raising the exception again to satisfy the expectation. I think this should be okay.

tests/unit/test_p11wrapper.cpp
johnfxgalea-e: Done.

tests/unit/test_p11wrapper.cpp
johnfxgalea-e: Done.

tests/unit/test_p11wrapper.cpp
johnfxgalea-e: Done

src/p11_wrap.cpp
johnfxgalea-e: Done. I did s/gets/fetches but refrained from delving deeper into implementation details for abstraction.

src/p11_wrap.cpp
johnfxgalea-e: Done. Resolved via inheritance.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done. Resolved via inheritance.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done. This function has been removed and encapsulated in sloftinfo class.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,211 @@
...
+     * as message. Internally, LibP11 uses OpenSSL's error handling framework,
+     * and therefore we simply extend OpenSSLException.
+     *
+     */
+    P11Exception() : OpenSSLException() {}
+
+    /**
+     * Error Strings.
+     *
+     * These are declared here to reduce string duplication as well as facilitate
```
garw: I don't really get this.
What are those? Why are they public?

src/mococrw/p11_lib.h
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Prefix all method names with 'P11_' to distinguish them from their
+ * LibP11 counterparts.
+ */
+class LibP11
+{
+public:
+    /* Initialisation */
+    static PKCS11_CTX *P11_PKCS11_CTX_new(void) noexcept;
+    static int P11_PKCS11_CTX_load(PKCS11_CTX *ctx, const char *ident) noexcept;
```
istepic, line 42: Can't this be a part of constructor? Also, specifying the right token could be an argument to constructor.

src/mococrw/p11_lib.h
```diff
@@ -0,0 +1,85 @@
...
+    static int P11_PKCS11_generate_key(PKCS11_TOKEN *token,
+                                       int algorithm,
+                                       unsigned int bits,
+                                       char *label,
+                                       unsigned char *id,
+                                       size_t id_len) noexcept;
+
+    /* Clean-up Functions */
+    static void P11_PKCS11_CTX_unload(PKCS11_CTX *ctx) noexcept;
+    static void P11_PKCS11_CTX_free(PKCS11_CTX *ctx) noexcept;
```
istepic, line 81: Destructor?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,211 @@
...
+    static const std::string nullSlotListExceptionString;
+    static const std::string nullTokenExceptionString;
+    static const std::string mismatchLoginExceptionString;
+};
+
+/*
+ * Wrap all the pointer-types returned by LibP11.
+ */
+
+using P11_PKCS11_CTX_Ptr = std::shared_ptr<PKCS11_CTX>;
```
garw: The shared_ptr aliase are usually called _SharedPtr in this lib.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,211 @@
...
+class P11_SlotInfo
+{
+public:
+    P11_SlotInfo(P11_PKCS11_CTX_Ptr ctx);
+
+    struct SlotListDeleter
+    {
+        SlotListDeleter(P11_PKCS11_CTX_Ptr ctx, int numSlots) : _ctx(ctx), _numSlots(numSlots) {}
+        void operator()(PKCS11_SLOT *);
+
```
garw, line 104: should this be private?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,211 @@
...
+enum SessionMode { ReadOnly = 0, ReadWrite = 1 };
+
+class P11_Session
+{
+public:
+    P11_Session(P11_PKCS11_SLOT_PTR slot, const std::string &pin, SessionMode mode);
+    ~P11_Session();  // After session is made out-of-scope, trigger logout.
+
+    /* Returns the token descriptor associated with the passed slot. */
+    P11_PKCS11_TOKEN_PTR getTokenFromSlot();
```
garw: This is a session object. This function takes no slots. How does it get the token *fromSlot*?

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,211 @@
...
+ * Generates a key inside the HSM.
+ *
+ * @param session The session used to generate the key.
+ * @param bits The size of the modulus in bits.
+ * @param label The label of the generated key.
+ * @param id The ID of the generated key.
+ *
+ * \note Due to limited support offered by LibP11, this function only generates RSA keys.
+ */
+void _PKCS11_generate_key(P11_Session &session,
```
garw: If this only generates RSA keys, then it should maybe be cvalled PKCS11_generate_RSA_key?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+        throw P11Exception(P11Exception::nullSlotListExceptionString);
+    }
+
+    SlotListDeleter deleter(_ctx, _numSlots);
+    _slotList = P11_PKCS11_SLOT_LIST_PTR(slots, deleter);
+}
+
+void P11_SlotInfo::SlotListDeleter::operator()(PKCS11_SLOT *slotList)
+{
+    if (slotList == nullptr) {
```
garw: why is this an error? Calling delete on a nullptr is considered safe.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,211 @@
...
+    P11_PKCS11_SLOT_LIST_PTR _slotList;  // Array consisting of slot descriptors on the HSM.
+    unsigned int _numSlots;              // The size of the array, i.e., the number of slots.
+};
+
+/**
+ * Session mode denotes whether to open a session in read/write mode.
+ */
+enum SessionMode { ReadOnly = 0, ReadWrite = 1 };
+
+class P11_Session
```
garw: Do we think it is reasonable that the session objects are copyable? This is probably something that shouldn't be copyable.

src/mococrw/p11_wrap.h
```diff
@@ -0,0 +1,211 @@
...
+using P11_PKCS11_SLOT_PTR = std::shared_ptr<PKCS11_SLOT>;
+using P11_PKCS11_TOKEN_PTR = std::shared_ptr<PKCS11_TOKEN>;
+
+/**
+ * Details information related to a slot list, obtained via enumeration of HSM.
+ */
+class P11_SlotInfo
+{
+public:
+    P11_SlotInfo(P11_PKCS11_CTX_Ptr ctx);
```
garw: This gets a context. Should this be copyable? How is this related to the context? Does it take ownership of this? Does it just prevent the context from being destroyed by taking a reference? This probably needs a bit of documentation what the design intent is here.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+{
+    PKCS11_SLOT *slots;
+
+    // Enumerate slots to obtain slot information.
+    P11IsNonNegative::callChecked(
+            lib::LibP11::P11_PKCS11_enumerate_slots, _ctx.get(), &slots, &_numSlots);
+
+    // If the returned slot list is NULL, we cannot continue!
+    if (slots == nullptr) {
+        throw P11Exception(P11Exception::nullSlotListExceptionString);
```
garw: Why do we use these rather static error messages? I'd rather like to see situation specific messages in the respective circumstances. "Found null slot list" will not tell me anything. I'd like to see here that it failed to enumerate the PKCS#11 slot and got null for the slotlist back…

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if (slotList == nullptr) {
+        throw P11Exception(P11Exception::nullSlotListExceptionString);
+    }
+
+    lib::LibP11::P11_PKCS11_release_all_slots(_ctx.get(), slotList, _numSlots);
+}
+
+P11_PKCS11_SLOT_PTR P11_SlotInfo::findSlot()
+{
+    PKCS11_SLOT *raw_slot = P11CallPtr::callChecked(
```
garw: auto here?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+            lib::LibP11::P11_PKCS11_find_token, _ctx.get(), _slotList.get(), _numSlots);
+
+    // Leverage aliasing.
+    return P11_PKCS11_SLOT_PTR(_slotList, raw_slot);
```
garw: I don't get the comment and I don't understand what this is doing here. This creates a new pointer to raw_slot or what? Is this even supposed to be deleted when it came from find_token? If yes, is it supposed to be deleted by delete or some library specific deleter?

src/mococrw/p11_wrap.h
garw: Are these objects supposed to be deleted by C++ delete or by some library specific delete function?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+ */
+
+P11_Session::P11_Session(P11_PKCS11_SLOT_PTR slot, const std::string &pin, SessionMode mode)
+        : _slot(slot)
+{
+    openSession(mode);
+    login(pin);
+}
+
+P11_Session::~P11_Session() { logout(); }
```
garw: Why don't we close the session?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+/***********************************************************************************
+ * Implementation of P11_Session
+ */
+
+P11_Session::P11_Session(P11_PKCS11_SLOT_PTR slot, const std::string &pin, SessionMode mode)
+        : _slot(slot)
+{
+    openSession(mode);
+    login(pin);
```
garw, line 115: Are we sure we always need to login? Don't we want to make this explicit instead of the session constructor?

src/mococrw/p11_wrap.h
garw: Why is this called `P11_Session`. Normally, this lib uses CamelCase for type names…

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+{
+    P11IsNonNegative::callChecked(
+            lib::LibP11::P11_PKCS11_open_session, _slot.get(), mode /* Denotes read/write mode. */);
+}
+
+void P11_Session::login(const std::string &pin)
+{
+    if (isLoggedIn()) {
+        // Weird: We are trying to log in, when we are logged in already.
+        throw P11Exception(P11Exception::mismatchLoginExceptionString);
```
garw: Are we sure this should be an error? We could just ignore it…

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    P11IsNonNegative::callChecked(lib::LibP11::P11_PKCS11_login,
```
garw, line 140: Maybe we want some better error handling here. Wrong PIN will probably occur some times. Do we get an error message out of callChecked that will be understandable enough?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+void P11_Session::logout()
+{
+    if (!isLoggedIn()) {
+        // Weird: We are trying to log out, when we are logged out already.
+        throw P11Exception(P11Exception::mismatchLoginExceptionString);
+    }
+
+    P11IsNonNegative::callChecked(lib::LibP11::P11_PKCS11_logout, _slot.get());
```
garw: Your current setup creates a destructor that throws. This is very discouraged in C++.

src/mococrw/p11_wrap.h
garw: I have the feeling we should have a chat about this class and its intention…

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/***********************************************************************************
+ * Implementation of API functions.
+ */
+
+/* Destroys a PKCS11 context by first unloading it, and then freeing it. */
+static void _PKCS11_CTX_destroy(PKCS11_CTX *ctx)
+{
+    if (ctx == nullptr) {
```
garw: s.a. 
free on nullptr is usually considered safe

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+        throw P11Exception(P11Exception::nullCtxExceptionString);
+    }
+    lib::LibP11::P11_PKCS11_CTX_unload(ctx);
+    lib::LibP11::P11_PKCS11_CTX_free(ctx);
+}
+
+P11_PKCS11_CTX_Ptr _PKCS11_CTX_create(const std::string &module)
+{
+    PKCS11_CTX *ctx = P11CallPtr::callChecked(lib::LibP11::P11_PKCS11_CTX_new);
```
garw, line 166: Why is no RAII object created here already? This would eliminate the need for this try  catch block

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+        // Free the ctx if loading failed and then throw.
+        lib::LibP11::P11_PKCS11_CTX_free(ctx);
+        throw;
+    }
+
+    // Create shared pointer which will unload and free the ctx automatically.
+    return P11_PKCS11_CTX_Ptr{ctx, _PKCS11_CTX_destroy};
+}
+
+static std::string parseID(const std::string &idHexString)
```
garw: maybe rather add some generic unhex utility function somehwhere? Why do we need this very special thing here? Is the result really always a string or binary data?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+using P11CallPtr =
+        ::cppc::CallCheckContext<::cppc::IsNotNullptrReturnCheckPolicy, P11ExceptionErrorPolicy>;
+using P11IsNonNegative =
+        ::cppc::CallCheckContext<::cppc::IsNotNegativeReturnCheckPolicy, P11ExceptionErrorPolicy>;
+
+/***********************************************************************************
+ * Implementation of P11_SlotInfo
+ */
+
+P11_SlotInfo::P11_SlotInfo(P11_PKCS11_CTX_Ptr ctx) : _ctx(ctx), _slotList(nullptr), _numSlots(0)
```
istepic: Can this be somehow a part of P11_Session class? I'm not sure why user of this  API needs to think about constructing 2 different objects before doing anything with HSM.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+    return parsedID;
+}
+
+void _PKCS11_store_private_key(P11_Session &session,
+                               EVP_PKEY *pk,
+                               const std::string &label,
+                               const std::string &id)
+{
+    std::string parsedStrID = parseID(id);
+    unsigned char *parsedID =
```
garw: This looks a bit like this should've been put into a vector of unsigned char?

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+                                  parsedStrID.size());
+}
+
+void _PKCS11_store_public_key(P11_Session &session,
+                              EVP_PKEY *pk,
+                              const std::string &label,
+                              const std::string &id)
+{
+    auto parsedStrID = parseID(id);
+    unsigned char *parsedID =
```
garw: s.a.

src/p11_wrap.cpp
```diff
@@ -0,0 +1,276 @@
...
+                                  parsedStrID.size());
+}
+
+void _PKCS11_generate_key(P11_Session &session,
+                          unsigned int bits,
+                          const std::string &label,
+                          const std::string &id)
+{
+    std::string parsedStrID = parseID(id);
+    unsigned char *parsedID =
```
garw: s.a.

istepic (commented):
My only concern is complexity of this thing. It's great that you tried to wrap libp11 like we wrap openssl but I wonder if that was really necessary and if this adds more complexity to the API than needed. It'd be great to see API providing something like `ECCKeypair ecc_keypair = AsymmetricKey::fromHSM(ECC_KEY, slot, pin, key_label);` (if you planned to do this in later PR's, I'd suggest you do it in this one) and 1 class that takes care of finding the slot, initialising context, logging in and out etc...

garw (changes requested):

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: It is passed via the constructor. I fixed the comment so it is clearer. I also renamed this function for the same purpose.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: It was something which I did think of in my initial design, but I think this current approach is better. The reason is that enumeration is done once for slot selection, whereas sessions are created for every operation. In other words, we do not want to perform slot enumeration every time we perform an operation on the HSM.

src/mococrw/p11_lib.h
johnfxgalea-e: Ah, maybe it was not clear - the `p11_lib.h` is a very raw and thin layer to LibP11. It takes a similar approach to what has been done for OpenSSL with the introduction of `openssl_lib.h`. In turn, `p11_wrap.h` uses `p11_lib.h` to enable C++ features. It is at this layers, where translation to shared pointer of  PKCS11_CTX takes place, along with the specification of a custom deleter. In future PRs, functions of `p11_wrap.h` will be used internally by MoCOCrW to implement interfaces like the one you mentioned (i.e., `*fromHSM`). If you are interested in a complete picture, see the RFC PR to get an idea https://github.com/bmwcarit/MoCOCrW/pull/106.

I updated the commit message to make it a bit clearer.

W.R.T why I did not add the user interface in this PR - I think this PR is already big and complex enough. It will be harder to review and get merged if more code is added, imo but I am happy to explain further if there are any questions.

src/mococrw/p11_lib.h
johnfxgalea-e: s.a

src/p11_wrap.cpp
istepic: Consider using isEnumerated flag in constructor then to simplify design.

src/mococrw/p11_lib.h
istepic: If _lib.h and _wrap.h are not supposed to be used by end-users, I wonder why do we not put those in src/mococrw/private folder then.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: No, they have custom deleters, which are specified upon their construction. I added a comment to make this clearer.

src/p11_wrap.cpp
johnfxgalea-e: Let's ignore. Done.

src/mococrw/p11_lib.h
johnfxgalea-e: I followed the directory path structure as openssl_lib.h, since they are similar

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: I saw this as an overkill, because I want the resource release of the context to also include unloading as well as free.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: This is the shared pointer's aliasing constructor. Basically since `raw_slot` is a pointer to memory owned by `_slotList`, the alias constructor is used to add the new shared_pointer, associated to the `raw_slot` pointer, in the share group containing `_slotList`.

See https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: They were placed there to facilitate exception testing as requested by a previous reviewer. I now reverted.

src/p11_wrap.cpp
johnfxgalea-e: Done

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/mococrw/p11_wrap.h
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: Done.

src/p11_wrap.cpp
johnfxgalea-e: This is just a simple helper function to convert IDs written in hex string by the user, and converts it to data so that it can be passed to libp11. The comment below explains this quite clearly imo.

As for:

> add some generic unhex utility function somehwhere

Not sure what the goal of this is. The code already makes use of a " generic unhex utility function", provided by the Boost API.

src/p11_wrap.cpp
johnfxgalea-e: The idea here was to login and logout only when a session is open, rather than stay in a persistent state of login throughout execution. Putting it in the constructor, and then logout in the destructor facilitates this process.

src/mococrw/p11_wrap.h
johnfxgalea-e: Sure. However, in essence, this class, and others, are there to put into effect the RAII principle.

src/p11_wrap.cpp
johnfxgalea-e: I'm not sure if this is worth the effort, and I'm not sure I like the prospect of having this class's constructor stateful based on a flag.

This API is an internal one and is not a concern for the user.

src/p11_wrap.cpp
johnfxgalea-e: Closure is done internally by libp11, when the slot is release: 

See https://github.com/OpenSC/libp11/blob/be13962a05c0e8c630bfd74eecd45f24cf5011c0/src/p11_slot.c#L433

src/p11_wrap.cpp
johnfxgalea-e: LibP11 internally already has error strings defined, which are used by P11Exception.

In the case where login fails, see:

https://github.com/OpenSC/libp11/blob/be13962a05c0e8c630bfd74eecd45f24cf5011c0/src/p11_slot.c#L208
https://github.com/OpenSC/libp11/blob/be13962a05c0e8c630bfd74eecd45f24cf5011c0/src/p11_ckr.c#L102

johnfxgalea-e:
@wusto gave an approve offline as he cannot login today. @wusto please approve when you can. Thanks

wusto (approved):
Looks good to me.
Someone else should please have a second look.

johnfxgalea-e:
Currently under internal architectural review. Not progressing further for the time being with this PR.

PR #114, opened 2022-07-07
Add HSM Test Setup to CI

johnfxgalea-e:
Adds setup and tear down for HSM tests to CI. In particular commands are added to initialise and destroy a SoftHSM2 token prior to and after the invocation of ctests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,11 +91,11 @@ jobs:
           docker run --rm -e SOFTHSM2_CONF="${{ matrix.softhsm2_conf }}" \
             -u "$UID"${{ matrix.softhsm2_grp }} \
             -v "$PWD:/src:rw" -w /src/build ${{ matrix.docker_tag }}  bash -c \
-            'if [ ${{ matrix.hsm_flag }} = "HSM_ENABLED=ON" ]; then \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -10,6 +10,10 @@ env:
   DOCKER_TAG: buildenv
   DOCKER_FILE_PATH: ./docker-build-env/Dockerfile
 
+  TOKEN_LABEL: token-label
+  USER_PIN: 1234
+  SO_PIN: 4321
```
istepic, line 18: Can we keep these the same just to avoid any problems this might cause? For example, libp11 doesn't have different engine control commands for PIN and SO_PIN, only `PIN: Specifies the pin code`.

istepic (changes requested):

.github/workflows/ci.yml
johnfxgalea-e: I'm not sure I agree. If different pins causes problems, then we would want to know about such problems (hopefully, via the CI) and fix them, rather than silent them. Besides, I would think it is unlikely for SO and User Pins to be the same on actual HSMs in the real-world.

istepic (approved):

wusto (approved):

PR #115, opened 2022-07-07
Make CI consider L/ASAN options

johnfxgalea-e:
Adds a suppression list to avoid LSAN from complaining about known false positives. To achieve this, we rely on the settings of LSAN and ASAN options.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,15 +12,10 @@ env:
 
   DOCKER_WITH_HSM_TAG: buildenv_with_hsm
   DOCKER_WITH_HSM_FILE_PATH: ./docker-build-env/Dockerfile-With-HSM
- 
+
   TOKEN_LABEL: token-label
   USER_PIN: 1234
   SO_PIN: 4321
-  
-  # Include suppression list to avoid false positives reported by LSAN.
-  LSAN_OPTIONS: "suppressions=/src/lsan_suppression.txt"
-  # Disable fast unwind to have clean stack-traces for better suppression.
-  ASAN_OPTIONS: "fast_unwind_on_malloc=false"
 
 jobs:
   check-format:
@@ -50,6 +45,16 @@ jobs:
         hsm_flag: ["HSM_ENABLED=ON", "HSM_ENABLED=OFF"]
 
         include:
+          - build_type: "ASAN"
+            # Disable fast unwind to have clean stack-traces for better suppression.
+            ASAN_OPTIONS: "fast_unwind_on_malloc=false"
+            # Include suppression list to avoid false positives reported by LSAN.
+            LSAN_OPTIONS: "suppressions=/src/lsan_suppression.txt"
+
+          - build_type: ""
+            ASAN_OPTIONS: ""
+            LSAN_OPTIONS: ""
+
           - hsm_flag: "HSM_ENABLED=ON"
             docker_tag: "buildenv_with_hsm"
             softhsm2_conf: "/usr/share/softhsm/softhsm2.conf"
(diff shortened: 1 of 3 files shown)


lsan_suppression.txt
```diff
@@ -0,0 +1,2 @@
+# Leak arises from OpenSSL usage of MoCOCrW and SoftHSM2.
```
wusto: This is too brief to judge.
Please add a technical explanation why this leak cannot be solved and is accepted.

.github/workflows/ci.yml
```diff
@@ -84,7 +89,8 @@ jobs:
 
       - name: "Run tests"
         run: |
-          docker run --rm -e SOFTHSM2_CONF="${{ matrix.softhsm2_conf }}" \
+          docker run --rm -e ASAN_OPTIONS="${{ env.ASAN_OPTIONS }}" -e LSAN_OPTIONS="${{ env.LSAN_OPTIONS }}" \
```
wusto: What happens if ASAN options and LSAN options are provided but the build-type is a no-asan one?

I am not sure, if I like the overloading and somehow tend to create an extra step which is dependent on the build-type.

docker-build-env/Dockerfile
```diff
@@ -11,6 +11,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-rec
         googletest \
         libboost-all-dev \
         libssl-dev \
+        # For llvm-symbolizer (used by ASAN).
+        llvm \
```
wusto, line 17: Wasn't ASAN already enabled in previous builds?
Is there a reason why it is needed now?

wusto (changes requested):

.github/workflows/ci.yml
johnfxgalea-e: They are ignored. Do you prefer if then else statements, with and without OPTION specification instead?

docker-build-env/Dockerfile
johnfxgalea-e: This is needed for LLVM's symboliser to print clearer stack-traces required for more accurate suppression.

wusto: Do you want to add this to your comment?
Iirc Ivan already asked the same. So it is not obvious.

.github/workflows/ci.yml
wusto: yes. That would be great.
Then it is clear what is ran.

wusto (changes requested):

docker-build-env/Dockerfile
johnfxgalea-e: Done.

lsan_suppression.txt
johnfxgalea-e: It is not MoCOCwR code, as far as I am aware. Added more clarification to the comment.

.github/workflows/ci.yml
johnfxgalea-e: I fixed this with matrix CI code instead, in order to avoid clunky if statements.

lsan_suppression.txt
```diff
@@ -0,0 +1,2 @@
+# Leak arises from OpenSSL p11 and SoftHSM2, where an engine is allocated during initilisation.
```
istepic: It's not clear where and in what scenario it leaks. A trace would help.

istepic (approved):
This leak doesn't look problematic to me. Consider fixing/raising issue upstream since MoCOCrW is not using OPENSSL_init_crypto.

lsan_suppression.txt
johnfxgalea-e: Makes sense. Will do!

johnfxgalea-e: Done.

wusto (approved):

PR #116, opened 2022-07-08
Adds main abstract classes related to HSMs

johnfxgalea-e:
Adds two main abstract classes concerning HSMs. The `HSM` class is the highest abstract class in the hierarchy; all HSM implementations should inherit it either directly or indirectly. The second abstract class is the `HsmEngine` class which extends `HSM` and adds functionality based on OpenSSL's `ENGINE_* API`.

Since these two classes are both abstract, no concrete implementation of an HSM is provided in this PR. Such an implementation will be provided in a subsequent PR to ease review.

### DIFF
--- a/src/hsm.cpp
+++ b/src/hsm.cpp
@@ -1,0 +1,51 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2022 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include "mococrw/hsm.h"
+#include <boost/format.hpp>
+
+namespace mococrw
+{
+using namespace openssl;
+
+HSM::HSM() {}
+
+HsmEngine::HsmEngine(const std::string &id, const std::string &modulePath, const std::string &pin)
+        : _id(id), _modulePath(modulePath), _pin(pin)
+{
+    // Fetch engine via ID.
+    _engine = _ENGINE_by_id(_id);
+
+    _ENGINE_ctrl_cmd_string(_engine.get(), "MODULE_PATH", _modulePath);
+    _ENGINE_ctrl_cmd_string(_engine.get(), "PIN", _pin);
+    _ENGINE_init(_engine.get());
+}
+
+HsmEngine::~HsmEngine() { _ENGINE_finish(_engine.get()); }
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPublicKey(const std::string &keyID)
+{
+    return _ENGINE_load_public_key(_engine.get(), keyID);
+}
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
+{
+    return _ENGINE_load_private_key(_engine.get(), keyID);
+}
+
+}  // namespace mococrw
--- a/src/hsm.cpp
+++ b/src/hsm.cpp
@@ -17,16 +17,18 @@
  * #L%
  */
 #include "mococrw/hsm.h"
-#include <boost/format.hpp>
 
 namespace mococrw
 {
 using namespace openssl;
 
-HSM::HSM() {}
+const std::string &HSM::getName() { return _name; }
 
-HsmEngine::HsmEngine(const std::string &id, const std::string &modulePath, const std::string &pin)
-        : _id(id), _modulePath(modulePath), _pin(pin)
+HsmEngine::HsmEngine(const std::string &name,
+                     const std::string &id,
+                     const std::string &modulePath,
+                     const std::string &pin)
+        : HSM(name), _id(id), _modulePath(modulePath), _pin(pin)
 {
     // Fetch engine via ID.
     _engine = _ENGINE_by_id(_id);
--- a/src/mococrw/hsm.h
+++ b/src/mococrw/hsm.h
@@ -1,0 +1,111 @@
+/*
+ * #%L
+ * %%
+ * Copyright (C) 2022 BMW Car IT GmbH
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+
+#include <boost/optional/optional.hpp>
+#include "openssl_wrap.h"
+
+namespace mococrw
+{
+/**
+ * Driver for Hardware Security Modules (HSMs).
+ */
+class HSM
+{
+public:
+    HSM();
+    virtual ~HSM() {}
+
+protected:
+    /**
+     * Returns the name of the HSM driver.
+     */
+    virtual const std::string getName() = 0;
+
+    /**
+     *  Loads public key from HSM.
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string &keyID) = 0;
+
+    /**
+     * Stores public key to HSM.
+     */
+    virtual void storePublicKey(EVP_PKEY *key,
+                                const std::string &label,
+                                const std::string &keyID) = 0;
+
+    /**
+     * Loads private key from HSM.
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPrivateKey(const std::string &keyID) = 0;
+
+    /**
+     * Stores private key to HSM.
+     */
+    virtual void storePrivateKey(EVP_PKEY *key,
+                                 const std::string &label,
+                                 const std::string &keyID) = 0;
+
+    /**
+     * Generates a key pair via HSM.
+     *
+     * \note Currently, RSA keys are only supported.
+     */
+    virtual void generateKey(unsigned int bits,
+                             const std::string &label,
+                             const std::string &id) = 0;
+
+    // Many of protected functions provided by the HSM class are seen
+    // as internal, not to be used by the User of MoCOCrW but specific
+    // friends:
+    friend class AsymmetricPublicKey;
+    friend class AsymmetricKeypair;
+    friend class RSASpec;
+};
+
+/**
+ * HSM driver implemented using OpenSSL's engine interface.
+ */
+class HsmEngine : public HSM
+{
+public:
+    HsmEngine(const std::string &id, const std::string &modulePath, const std::string &pin);
+    virtual ~HsmEngine();
+
+    ENGINE *internal();
+    const ENGINE *internal() const;
+
+protected:
+    /** Pointer to OpenSSL ENGINE. */
+    openssl::SSL_ENGINE_Ptr _engine;
+    /** Engine ID. */
+    const std::string &_id;
+    /** Path to Module. */
+    const std::string &_modulePath;
+    /** Pin to access PKCS11 Engine. */
+    const std::string &_pin;
+
+    virtual const std::string getName() override = 0;
+
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string &keyID) override;
+
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPrivateKey(const std::string &keyID) override;
+};
+
+}  // namespace mococrw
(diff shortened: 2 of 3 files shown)


src/mococrw/hsm.h
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Generates a key pair via HSM.
+     *
+     * \note Currently, RSA keys are only supported.
+     */
+    virtual void generateKey(unsigned int bits,
+                             const std::string &label,
+                             const std::string &id) = 0;
+
+    // Many of protected functions provided by the HSM class are seen
```
wusto: I guess the friend class belong logically to a different PR.

Please add them where the friend classes access the protected functions.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,111 @@
...
+    virtual ~HSM() {}
+
+protected:
+    /**
+     * Returns the name of the HSM driver.
+     */
+    virtual const std::string getName() = 0;
+
+    /**
+     *  Loads public key from HSM.
```
wusto, line 46: Wouldn't it make sense to add descriptions for the parameters or at least a link to the official documentation, which describes the mechanisms behind the arguments?

src/mococrw/hsm.h
```diff
@@ -0,0 +1,111 @@
...
+    /** Pointer to OpenSSL ENGINE. */
+    openssl::SSL_ENGINE_Ptr _engine;
+    /** Engine ID. */
+    const std::string &_id;
+    /** Path to Module. */
+    const std::string &_modulePath;
+    /** Pin to access PKCS11 Engine. */
+    const std::string &_pin;
+
+    virtual const std::string getName() override = 0;
```
wusto: So this class is still an abstract one?

From the class is description it reads like this is the actual implementation.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,111 @@
...
+protected:
+    /**
+     * Returns the name of the HSM driver.
+     */
+    virtual const std::string getName() = 0;
+
+    /**
+     *  Loads public key from HSM.
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string &keyID) = 0;
```
wusto, line 50: What's the rational to return a openssl::SSL_EVP_PKEY_Ptr and not an object of type AsymmetricPublicKey or AsymmetricKeypair?

Usually this requires then an additional layer of abstraction to cast/transform it into the key object(s) used by MoCOCrW.

wusto (changes requested):

src/mococrw/hsm.h
johnfxgalea-e: Makes sense. I will remove.

src/mococrw/hsm.h
johnfxgalea-e: This is an internal API which is to be used by AsymmetricPublicKey/AsymmetricKeypair (this is where the friend relationship comes in and why these functions are marked protected).

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#pragma once
+
+#include <boost/optional/optional.hpp>
```
istepic: Where is this used?

src/hsm.cpp
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+#include "mococrw/hsm.h"
+#include <boost/format.hpp>
```
istepic: Where is this used?

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+
+namespace mococrw
+{
+/**
+ * Driver for Hardware Security Modules (HSMs).
+ */
+class HSM
+{
+public:
+    HSM();
```
istepic: Not needed.

src/hsm.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace mococrw
+{
+using namespace openssl;
+
+HSM::HSM() {}
```
istepic: Not needed.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+ */
+#pragma once
+
+#include <boost/optional/optional.hpp>
+#include "openssl_wrap.h"
+
+namespace mococrw
+{
+/**
+ * Driver for Hardware Security Modules (HSMs).
```
istepic: Can you update/be more verbose in this comment? This class has no implementation so I wouldn't say it's a driver for anything.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+ * HSM driver that leverages OpenSSL's ENGINE_* API interface.
+ */
+class HsmEngine : public HSM
+{
+public:
+    HsmEngine(const std::string &id, const std::string &modulePath, const std::string &pin);
+    virtual ~HsmEngine();
+
+    ENGINE *internal();
+    const ENGINE *internal() const;
```
istepic: Please rethink why would we want to give ptr to ENGINE to clients. This is currently not used anywhere and the association between `ENGINE*` and `openssl::SSL_ENGINE_Ptr` is not clear to me.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+    /** Pointer to OpenSSL ENGINE. */
+    openssl::SSL_ENGINE_Ptr _engine;
+    /** Engine ID. */
+    const std::string &_id;
+    /** Path to Module. */
+    const std::string &_modulePath;
+    /** Pin to access PKCS11 Engine. */
+    const std::string &_pin;
+
+    virtual const std::string getName() override = 0;
```
istepic: It's tough to see use-case for this function and why it's not implemented.

garw (approved):
lg2m

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPublicKey(const std::string &keyID) override;
```
istepic: [Avoid using virtual and override.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final)

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPrivateKey(const std::string &keyID) override;
```
istepic: s.a.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+namespace mococrw
+{
+/**
+ * Driver for Hardware Security Modules (HSMs).
+ */
+class HSM
+{
+public:
+    HSM();
+    virtual ~HSM() {}
```
istepic: nitpick: Usually `= default;` is used in this codebase.

src/mococrw/hsm.h
```diff
@@ -0,0 +1,120 @@
...
+    virtual openssl::SSL_EVP_PKEY_Ptr loadPrivateKey(const std::string &keyID) = 0;
+
+    /**
+     * Stores private key to HSM.
+     *
+     * @param key The private key to store.
+     * @param label The label of the key to store.
+     * @param keyID The ID of the key to store.
+     */
+    virtual void storePrivateKey(EVP_PKEY *key,
```
istepic, line 77: Why is `EVP_PKEY*` used here and `openssl::SSL_EVP_PKEY_Ptr` in the function above? Preferably we would use only one for consistency unless this is how it's supposed to work.

src/hsm.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    // Fetch engine via ID.
+    _engine = _ENGINE_by_id(_id);
+
+    _ENGINE_ctrl_cmd_string(_engine.get(), "MODULE_PATH", _modulePath);
+    _ENGINE_ctrl_cmd_string(_engine.get(), "PIN", _pin);
+    _ENGINE_init(_engine.get());
+}
+
+HsmEngine::~HsmEngine() { _ENGINE_finish(_engine.get()); }
```
istepic, line 41: What about _ENGINE_free()?

istepic (dismissed):
Looks good overall but I would like to see a slightly expanded version of this PR which tests at least one of the HSMEngine methods. You did provide implementation for HSMEngine so this class is not really abstract and it would be good to see if construction/destruction of HSMEngine works as intended. An example wouldn't hurt but that can be done later.

src/mococrw/hsm.h
wusto: Can you please recheck. Seems like a rebase issue.

wusto (changes requested):

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: That's fair. I based it on a pattern of other existing classes in the code-base (e.g., `BioObject`), but I agree I don't need this functionality for now... Removing.

src/mococrw/hsm.h
johnfxgalea-e: I will implement this function by having the string be passed as a parameter via the construction.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/hsm.cpp
johnfxgalea-e: That function is triggered via the smart pointer automatically

src/hsm.cpp
johnfxgalea-e: N/A

src/hsm.cpp
johnfxgalea-e: Done

src/mococrw/hsm.h
johnfxgalea-e: Removed. Done

src/mococrw/hsm.h
johnfxgalea-e: Done.

src/mococrw/hsm.h
johnfxgalea-e: I tried to make the code comment clearer to denote that this class is also abstract.

src/mococrw/hsm.h
johnfxgalea-e: This is the pattern that the rest of the code base follows for intermediary layers such as the one in this PR. For example, `keygen()` takes the raw OpenSSL pointer as parameter, but returns the wrapped version of EVP_PKEY:

`SSL_EVP_PKEY_Ptr _EVP_PKEY_keygen(EVP_PKEY_CTX *ctx)`

See: https://github.com/bmwcarit/MoCOCrW/blob/361f880ae068a0a346e9bb9301e0025ff97e2892/src/key.cpp#L120

wusto (approved):

johnfxgalea-e:
> Looks good overall but I would like to see a slightly expanded version of this PR which tests at least one of the HSMEngine methods. You did provide implementation for HSMEngine so this class is not really abstract...

Not sure I follow. HSMEngine is not a concrete class, and indeed abstract.

src/hsm.cpp
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ * #L%
+ */
+#include "mococrw/hsm.h"
+
+namespace mococrw
+{
+using namespace openssl;
+
+const std::string &HSM::getName() { return _name; }
```
istepic, line 25: I still don't see a use-case for this _name attribute but I guess we'll see that in the next PRs.

istepic (commented):
> > Looks good overall but I would like to see a slightly expanded version of this PR which tests at least one of the HSMEngine methods. You did provide implementation for HSMEngine so this class is not really abstract...
> 
> Not sure I follow. HSMEngine is not a concrete class, and indeed abstract.

Depends on the [definition](https://en.cppreference.com/w/cpp/language/abstract_class) you use I guess. Typically it's nice to see some tests with implementation in the same PR but since I know you have something in your backpocket I'm going to remain neutral here.

src/hsm.cpp
johnfxgalea-e: This is not a vital function and not to be given significant importance. It is more with the intention of having a descriptive name, as a string, associated with concrete HSM classes (e.g., for logging).

PR #117, opened 2022-08-08
Fix front HSM interface

johnfxgalea-e:
Converts the HSMEngine class, which makes use of OpenSSL's ENGINE API to interact with HSMs, from an abstract to a concrete class. Virtual functions which cannot be implemented due to current lack of support (e.g., key storage) are removed for now.

Adds readPrivateKeyFromHSM() and readPublicKeyFromHSM() to the MoCOCrW interface. Tests for loading keys are also included in this PR.

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -30,6 +30,12 @@ if(BUILD_TESTING) #set by Ctest. Also set in the integration build environment.
 
     add_executable(openssltest test_opensslwrapper.cpp ${MOCK_SOURCES})
 
+    if(HSM_ENABLED)
+        add_executable(hsmtest test_hsm.cpp
+                               "${SRC_DIR}/hsm.cpp"
+                               ${MOCK_SOURCES})
+    endif()
+
     add_executable(asn1timetests test_asn1time.cpp ${REAL_SOURCES})
     add_executable(keytests test_key.cpp ${REAL_SOURCES})
     add_executable(csrtests test_csr.cpp "${SRC_DIR}/key.cpp" ${REAL_SOURCES})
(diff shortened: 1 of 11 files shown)


tests/unit/CMakeLists.txt
```diff
@@ -140,6 +146,11 @@ if(BUILD_TESTING) #set by Ctest. Also set in the integration build environment.
     # Cannot link to imported OpenSSL and leverage implicit include-dir-propagation, as openssltest uses mocks.
     target_include_directories(openssltest PUBLIC ${OPENSSL_INCLUDE_DIR})
 
+    if(HSM_ENABLED)
```
garw, line 149: Not sure why its split for the rest, but I think we should better group the directives per binary.

tests/unit/test_hsm.cpp
```diff
@@ -0,0 +1,105 @@
...
+    ON_CALL(_mock(), SSL_ERR_error_string(_, nullptr))
+            .WillByDefault(Return(const_cast<char*>(_defaultErrorMessage.c_str())));
+    // TODO: Get rid of the uninteresting calls by default here somehow...
+}
+
+void HSMTest::TearDown() { openssl::OpenSSLLibMockManager::destroy(); }
+
+TEST_F(HSMTest, testHSMEngine)
+{
+    std::string engineID = "engine_id";
```
garw: Stick to one init style?

wusto (approved):
lgtm. Someone else should please have another look.

garw (changes requested):
content-wise it looks good to me. just some nitpicks.
Do we already have integration tests against softhsm? I fear I lost the overview here…

tests/unit/test_hsm.cpp
johnfxgalea-e: Done

tests/unit/CMakeLists.txt
johnfxgalea-e: Let's punt this for now as that clean-up does not really concern this PR's Contribution.

garw (approved):

PR #119, opened 2022-12-06
Apply libp11 patch for EC keygen on HSM and adapt CI

istepic:
Libp11 (0.4.12) is patched to support EC and RSA keypair generation on HSMs
through the ENGINE openssl interface. This custom patched libp11 is installed
in the docker container used in the CI. CI setup has been adapted as follows:
- Reworked docker setup to allow adding files to the image.
  This is not possible when using " - < " and reading from STDIN
  because there is no Docker Build Context there which is necessary
  for ADD and COPY commands.

- Built libp11 the same way upstream builds it in their CI. This
  requires libtool package.


dockerfiles/hsm-support/Dockerfile
```diff
@@ -2,14 +2,27 @@ FROM buildenv
...
+    wget https://github.com/OpenSC/libp11/releases/download/libp11-0.4.12/libp11-0.4.12.tar.gz && \
+    tar xvf libp11-0.4.12.tar.gz && \
+    cd libp11-0.4.12 && \
+    git apply /tmp/keygen.patch && \
+    echo "Successfully patched libp11" && \
+    autoreconf --verbose --install --force && \
+    ./configure --enable-strict && \
+    make -j$(nproc) && \
+    make check && \
+    make install
```
garw: you probably want to do all of this in a subdir of /tmp that can later be removed again to reduce the size of the resulting container.

dockerfiles/hsm-support/keygen.patch
```diff
@@ -0,0 +1,766 @@
+diff --git a/src/eng_back.c b/src/eng_back.c
```
garw: Can we maybe add a complete git style patch here with commit message?

.github/workflows/ci.yml
```diff
@@ -8,10 +8,10 @@ on:
 
 env:
   DOCKER_TAG: buildenv
-  DOCKER_FILE_PATH: ./docker-build-env/Dockerfile
+  DOCKER_FILE_PATH: ./dockerfiles/base
 
   DOCKER_WITH_HSM_TAG: buildenv_with_hsm
-  DOCKER_WITH_HSM_FILE_PATH: ./docker-build-env/Dockerfile-With-HSM
+  DOCKER_WITH_HSM_FILE_PATH: ./dockerfiles/hsm-support
```
garw: minor: The variable name is now slightly lying to you as it isn't the file path anymore but rather like some directory.

garw (changes requested):

.github/actions/build-docker/action.yml
```diff
@@ -17,5 +17,5 @@ runs:
     - name: "Build Docker image"
       shell: bash
       run: |
-        docker build -t ${{ inputs.docker_tag }} - < ${{ inputs.docker_file_path }}
+        docker build -t ${{ inputs.docker_tag }} ${{ inputs.docker_file_path }}
```
johnfxgalea-e: Nit: s/docker_file_path/docker_ctx_path ?

.github/workflows/ci.yml
johnfxgalea-e: Nit: I agree. Maybe s/FILE/CONTEXT?

dockerfiles/hsm-support/Dockerfile
```diff
@@ -2,14 +2,27 @@ FROM buildenv
...
         # softhsm2: includes both softhsm2-util and libsofthsm2
         softhsm2 \
+        # libp11 needs this
+        libtool \
         && rm -rf /var/lib/apt/lists/*
+
+ADD keygen.patch /tmp/
+
+RUN cd /tmp && \
+    wget https://github.com/OpenSC/libp11/releases/download/libp11-0.4.12/libp11-0.4.12.tar.gz && \
```
johnfxgalea-e: Nit: I prefer not keeping URLS in command but in a separate variable for easy change and clarity.

Maybe use docker's Arg? This will have the added benefit of build time customisation.

johnfxgalea-e (changes requested):
Overall, good job! I am being very picky with my requested changes and are mostly nits.

dockerfiles/hsm-support/keygen.patch
istepic: Done

dockerfiles/hsm-support/Dockerfile
istepic: Done

.github/workflows/ci.yml
istepic: Agreed, I used `folder` instead. Context might be too abstract for a casual reader.

dockerfiles/hsm-support/Dockerfile
istepic: Done

garw (approved):
lgtm now

johnfxgalea-e (approved):

PR #120, opened 2022-12-06
Add keypair generation on HSMs

istepic:
Introduced keypair generation on HSMs through the OpenSSL's ENGINE interface.
This is based on a temporary solution where libp11 is patched to support keypair
generation.

### DIFF
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -99,12 +99,11 @@ int main(void)
     std::vector<uint8_t> message = utility::fromHex("deadbeef");
     HsmEngine hsmEngine(id, modulePath, pin);
 
-
-    /************** ECDSA signature **************/
+    /************** ECC key generation and ECDSA **************/
     std::string keyIDEcc("5567");
     auto ecdsaDigestType = DigestTypes::SHA512;
     ECCSpec ecspec;
-    auto eccPrivKey = AsymmetricPrivateKey::genKeyOnHsmGetPrivate(
+    auto eccPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
             hsmEngine, ecspec, keyIDEcc, "token-label", "DobarKey");
     auto ecdsaSigFormat = ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS;
 
@@ -119,6 +118,7 @@ int main(void)
      */
     ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, signature, message);
 
+    /************** Get pubkey from HSM, store it in PEM file for later use **************/
     /* Use-Case 2: The public key used for verification is stored in the HSM */
     auto pubKeyEcc = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDEcc);
     ecdsaVerify(pubKeyEcc, ecdsaDigestType, ecdsaSigFormat, signature, message);
@@ -132,10 +132,10 @@ int main(void)
     ecdsaVerify(pubKeyEccFromPEm, ecdsaDigestType, ecdsaSigFormat, signature, message);
     /*********************************************/
 
-    /************** RSA key generation **************/
+    /************** RSA key generation and loading **************/
     std::string keyIDRsa("8890");
     mococrw::RSASpec rsaSpec;
-    auto rsaPrivKey = AsymmetricPrivateKey::genKeyOnHsmGetPrivate(
+    auto rsaPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
             hsmEngine, rsaSpec, keyIDRsa, "token-label", "BarfoKey");
 
     /* Read public key from HSM */
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -158,17 +158,18 @@ int main(void)
     // Information for engine loading and key management.
     std::string id("pkcs11");
     std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
+    std::string tokenLabel("token-label");
     // Don't hardcode the pin in your application, this is just for demonstration purposes
     std::string pin("1234");
+    HsmEngine hsmEngine(id, modulePath, tokenLabel, pin);
     std::vector<uint8_t> message = utility::fromHex("deadbeef");
-    HsmEngine hsmEngine(id, modulePath, pin);
 
     /************** ECC key generation and ECDSA **************/
     std::string keyIDEcc("123");
     auto ecdsaDigestType = DigestTypes::SHA512;
     ECCSpec ecspec;
-    auto eccPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
-            hsmEngine, ecspec, "token-label", keyIDEcc, "DobarKey");
+    auto eccPrivKey =
+            AsymmetricPrivateKey::generateKeyOnHsm(hsmEngine, ecspec, keyIDEcc, "DobarKey");
     auto ecdsaSigFormat = ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS;
 
     /* The argument hashFunction is optional. Default is SHA256
@@ -207,8 +208,8 @@ int main(void)
     /**
      * Generate an RSA keypair and load the public part
      */
-    auto rsaPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
-            hsmEngine, rsaSpec, "token-label", keyIDRsa, "BarfoKey");
+    auto rsaPrivKey =
+            AsymmetricPrivateKey::generateKeyOnHsm(hsmEngine, rsaSpec, keyIDRsa, "BarfoKey");
     auto pubKeyRsa = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDRsa);
 
     /**
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -111,8 +111,8 @@ int main(void)
      * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
     auto signature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
 
-    /* Use-Case 1: You want to check your own signature:
-     * we can use here the private key, as it also contains the public key.
+    /* Use-Case 1: You want to check your own signature.
+     * We can use here the private key, as it also contains the public key.
      * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
      * we do an implicit upcast here.
      */
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -17,7 +17,9 @@
  * #L%
  */
 #include <mococrw/asymmetric_crypto_ctx.h>
+#include <mococrw/hash.h>
 #include <mococrw/hsm.h>
+#include <mococrw/key.h>
 #include <iostream>
 
 /* This example demonstrates how to create a PKCS11 engine object,
@@ -26,6 +28,67 @@
 
 using namespace mococrw;
 
+std::vector<uint8_t> ecdsaSign(const AsymmetricPrivateKey &privKey,
+                               const DigestTypes digestType,
+                               const ECDSASignatureFormat sigFormat,
+                               const std::vector<uint8_t> &message)
+{
+    std::shared_ptr<MessageSignatureCtx> signCtx;
+    try {
+        /* The argument hashFunction is optional. Default is SHA256
+         * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
+        signCtx = std::make_shared<ECDSASignaturePrivateKeyCtx>(privKey, digestType, sigFormat);
+    } catch (MoCOCrWException &e) {
+        std::cerr << "Please check your ECC key. Failure creating context." << std::endl;
+        std::cerr << e.what();
+        exit(EXIT_FAILURE);
+    }
+
+    std::vector<uint8_t> signature;
+    try {
+        signature = signCtx->signMessage(message);
+    } catch (const MoCOCrWException &e) {
+        /* Possible reasons:
+         * - error in openssl (sign, padding, ...)
+         * - Hash function's digest size doesn't match the message's digest size
+         * - Invalid signature format set
+         */
+        std::cerr << "Failure occurred during signing." << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+    return signature;
+}
+
+void ecdsaVerify(const AsymmetricPublicKey &pubKey,
+                 const DigestTypes digestType,
+                 const ECDSASignatureFormat sigFormat,
+                 const std::vector<uint8_t> &signature,
+                 const std::vector<uint8_t> &message)
+{
+    std::shared_ptr<MessageVerificationCtx> verifyCtx;
+    try {
+        verifyCtx = std::make_shared<ECDSASignaturePublicKeyCtx>(pubKey, digestType, sigFormat);
+    } catch (const MoCOCrWException &e) {
+        std::cerr << "Please check your ECC key. Failure creating context." << std::endl;
+        std::cerr << e.what();
+        exit(EXIT_FAILURE);
+    }
+    try {
+        verifyCtx->verifyMessage(signature, message);
+    } catch (const MoCOCrWException &e) {
+        /* Possible reasons:
+         * - error in openssl (sign, padding, ...)
+         * - Hash function's digest size doesn't match the message's digest size
+         * - Invalid signature format set
+         * - Signature can't be parsed
+         * - Invalid signature
+         */
+        std::cerr << "Verification failed!" << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
 int main(void)
 {
     // Information for engine loading and key management.
@@ -33,13 +96,26 @@ int main(void)
     std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
     std::string pin("1234");
     std::string keyID("1001");
-
-    // Step 2: Initialise the PKCS11 Engine.
+    std::vector<uint8_t> message = utility::fromHex("deadbeef");
     HsmEngine hsmEngine(id, modulePath, pin);
 
-    // Step 4: Load the key from the HSM and print it.
-    auto pubKey = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyID);
-    std::cout << "Loaded key: \n\n" << pubKey.publicKeyToPem() << "\n\n" << std::endl;
+    /************** ECDSA signature **************/
+    auto ecdsaDigestType = DigestTypes::SHA512;
+    ECCSpec ecspec;
+    auto eccPrivKey = AsymmetricPrivateKey::genKeyOnHsmGetPrivate(
+            hsmEngine, ecspec, "5567", "token-label", "DobarKey");
+    auto ecdsaSigFormat = ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS;
+
+    /* The argument hashFunction is optional. Default is SHA256
+     * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
+    auto signature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
+
+    /* we can use here the private key, as it also contains the public key.
+     * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
+     * we do an implicit upcast here.
+     */
+    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, signature, message);
+    /*********************************************/
 
     return 0;
 }
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -89,32 +89,61 @@ void ecdsaVerify(const AsymmetricPublicKey &pubKey,
         exit(EXIT_FAILURE);
     }
 }
+
 int main(void)
 {
     // Information for engine loading and key management.
     std::string id("pkcs11");
     std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
     std::string pin("1234");
-    std::string keyID("1001");
     std::vector<uint8_t> message = utility::fromHex("deadbeef");
     HsmEngine hsmEngine(id, modulePath, pin);
 
+
     /************** ECDSA signature **************/
+    std::string keyIDEcc("5567");
     auto ecdsaDigestType = DigestTypes::SHA512;
     ECCSpec ecspec;
     auto eccPrivKey = AsymmetricPrivateKey::genKeyOnHsmGetPrivate(
-            hsmEngine, ecspec, "5567", "token-label", "DobarKey");
+            hsmEngine, ecspec, keyIDEcc, "token-label", "DobarKey");
     auto ecdsaSigFormat = ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS;
 
     /* The argument hashFunction is optional. Default is SHA256
      * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
     auto signature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
 
-    /* we can use here the private key, as it also contains the public key.
+    /* Use-Case 1: You want to check your own signature:
+     * we can use here the private key, as it also contains the public key.
      * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
      * we do an implicit upcast here.
      */
     ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, signature, message);
+
+    /* Use-Case 2: The public key used for verification is stored in the HSM */
+    auto pubKeyEcc = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDEcc);
+    ecdsaVerify(pubKeyEcc, ecdsaDigestType, ecdsaSigFormat, signature, message);
+
+    /* Use-Case 3: You want to write the public key to a PEM file and use it later
+     * for verification
+     */
+    auto pubKeyPem = eccPrivKey.publicKeyToPem();
+    /* Omitted: Write PEM data to a file and read it again */
+    auto pubKeyEccFromPEm = mococrw::AsymmetricPublicKey::readPublicKeyFromPEM(pubKeyPem);
+    ecdsaVerify(pubKeyEccFromPEm, ecdsaDigestType, ecdsaSigFormat, signature, message);
+    /*********************************************/
+
+    /************** RSA key generation **************/
+    std::string keyIDRsa("8890");
+    mococrw::RSASpec rsaSpec;
+    auto rsaPrivKey = AsymmetricPrivateKey::genKeyOnHsmGetPrivate(
+            hsmEngine, rsaSpec, keyIDRsa, "token-label", "BarfoKey");
+
+    /* Read public key from HSM */
+    auto pubKeyRsa = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDRsa);
+
+    /* Do whatever you want using the private and public key. See rsa-example.cpp or
+     * sig-example.cpp for further examples.
+     */
     /*********************************************/
 
     return 0;
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -28,6 +28,69 @@
 
 using namespace mococrw;
 
+std::vector<uint8_t> rsaSign(const AsymmetricPrivateKey &privKey,
+                             const DigestTypes digestType,
+                             std::shared_ptr<RSASignaturePadding> padding,
+                             const std::vector<uint8_t> &message)
+{
+    std::shared_ptr<MessageSignatureCtx> signCtx;
+
+    try {
+        /* Padding is optional. Default: PSSPadding with MGF1 as mask generation function.
+         * DigestType is used as hash function for the padding schemes and for MGF1 (if used) */
+        signCtx = std::make_shared<RSASignaturePrivateKeyCtx>(privKey, digestType, padding);
+    } catch (const MoCOCrWException &e) {
+        std::cerr << "Please check your RSA key. Failure creating context." << std::endl;
+        std::cerr << e.what();
+        exit(EXIT_FAILURE);
+    }
+
+    std::vector<uint8_t> signature;
+    try {
+        signature = signCtx->signMessage(message);
+    } catch (const MoCOCrWException &e) {
+        /* Possible reasons:
+         * - error in openssl (sign, padding, ...)
+         * - Hash function's digest size doesn't match the message's digest size
+         */
+        std::cerr << "Failure occurred during signing." << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+    return signature;
+}
+
+void rsaVerify(const AsymmetricPublicKey &pubKey,
+               std::shared_ptr<RSASignaturePadding> rsaPadding,
+               DigestTypes digestType,
+               const std::vector<uint8_t> &signature,
+               const std::vector<uint8_t> &message)
+{
+    std::shared_ptr<MessageVerificationCtx> verifyCtx;
+    try {
+        /* Padding is optional. Default: PSSPadding with MGF1 as mask generation function.
+         * DigestType is used as hash function for the padding schemes and for MGF1 (if used) */
+        verifyCtx = std::make_shared<RSASignaturePublicKeyCtx>(pubKey, digestType, rsaPadding);
+    } catch (const MoCOCrWException &e) {
+        std::cerr << "Please check your RSA key. Failure creating context." << std::endl;
+        std::cerr << e.what();
+        exit(EXIT_FAILURE);
+    }
+
+    try {
+        verifyCtx->verifyMessage(signature, message);
+    } catch (const MoCOCrWException &e) {
+        /* Possible reasons:
+         * - error in openssl (sign, padding, ...)
+         * - Hash function's digest size doesn't match the message's digest size
+         * - Invalid signature
+         */
+        std::cerr << "Verification failed!" << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
+
 std::vector<uint8_t> ecdsaSign(const AsymmetricPrivateKey &privKey,
                                const DigestTypes digestType,
                                const ECDSASignatureFormat sigFormat,
@@ -95,56 +158,72 @@ int main(void)
     // Information for engine loading and key management.
     std::string id("pkcs11");
     std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
+    // Don't hardcode the pin in your application, this is just for demonstration purposes
     std::string pin("1234");
     std::vector<uint8_t> message = utility::fromHex("deadbeef");
     HsmEngine hsmEngine(id, modulePath, pin);
 
     /************** ECC key generation and ECDSA **************/
-    std::string keyIDEcc("5567");
+    std::vector<uint8_t> keyIDECC{};
+    std::string keyLabelECC("ecc-key-label");
     auto ecdsaDigestType = DigestTypes::SHA512;
     ECCSpec ecspec;
     auto eccPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
-            hsmEngine, ecspec, keyIDEcc, "token-label", "DobarKey");
+            hsmEngine, ecspec, "token-label", keyLabelECC, {});
     auto ecdsaSigFormat = ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS;
 
     /* The argument hashFunction is optional. Default is SHA256
      * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
-    auto signature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
+    auto ECDSAsignature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
 
     /* Use-Case 1: You want to check your own signature.
      * We can use here the private key, as it also contains the public key.
      * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
      * we do an implicit upcast here.
      */
-    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, signature, message);
+    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, ECDSAsignature, message);
 
-    /************** Get pubkey from HSM, store it in PEM file for later use **************/
     /* Use-Case 2: The public key used for verification is stored in the HSM */
-    auto pubKeyEcc = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDEcc);
-    ecdsaVerify(pubKeyEcc, ecdsaDigestType, ecdsaSigFormat, signature, message);
+    auto pubKeyEcc = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyLabelECC, keyIDECC);
+    ecdsaVerify(pubKeyEcc, ecdsaDigestType, ecdsaSigFormat, ECDSAsignature, message);
 
     /* Use-Case 3: You want to write the public key to a PEM file and use it later
      * for verification
      */
     auto pubKeyPem = eccPrivKey.publicKeyToPem();
-    /* Omitted: Write PEM data to a file and read it again */
-    auto pubKeyEccFromPEm = mococrw::AsymmetricPublicKey::readPublicKeyFromPEM(pubKeyPem);
-    ecdsaVerify(pubKeyEccFromPEm, ecdsaDigestType, ecdsaSigFormat, signature, message);
+    /* Write PEM data to a file and read it again */
+    auto pubKeyEccFromPem = mococrw::AsymmetricPublicKey::readPublicKeyFromPEM(pubKeyPem);
+    ecdsaVerify(pubKeyEccFromPem, ecdsaDigestType, ecdsaSigFormat, ECDSAsignature, message);
     /*********************************************/
 
-    /************** RSA key generation and loading **************/
-    std::string keyIDRsa("8890");
+    /************** RSA key generation, loading and digital signatures **************/
+    std::vector<uint8_t> keyIDRSA{0x12, 0x34};
+    std::string keyLabelRSA{"rsa-key-label"};
+    auto mgf1DigestType = DigestTypes::SHA256;
     mococrw::RSASpec rsaSpec;
-    auto rsaPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
-            hsmEngine, rsaSpec, keyIDRsa, "token-label", "BarfoKey");
+    auto rsaSignatureDigestType = DigestTypes::SHA512;
+    auto mgf1 = std::make_shared<MGF1>(mgf1DigestType);
+    int saltLength = mococrw::Hash::getDigestSize(rsaSignatureDigestType);
+    auto padding = std::make_shared<PSSPadding>(mgf1, saltLength);
 
-    /* Read public key from HSM */
-    auto pubKeyRsa = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDRsa);
-
-    /* Do whatever you want using the private and public key. See rsa-example.cpp or
-     * sig-example.cpp for further examples.
+    /**
+     * Generate an RSA keypair and load the public part
      */
-    /*********************************************/
+    auto rsaPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
+            hsmEngine, rsaSpec, "token-label", keyLabelRSA, keyIDRSA);
+    auto pubKeyRsa = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyLabelRSA, keyIDRSA);
+
+    /**
+     * Do signing/verification
+     */
+    auto RSAsignature = rsaSign(rsaPrivKey, rsaSignatureDigestType, padding, message);
+    // We can use here the private key, as it also contains the public key.
+    rsaVerify(rsaPrivKey, padding, rsaSignatureDigestType, RSAsignature, message);
+    // ... or you can use the public key
+    rsaVerify(pubKeyRsa, padding, rsaSignatureDigestType, RSAsignature, message);
+    /**
+     * See rsa-example.cpp or sig-example.cpp for further examples.
+     */
 
     return 0;
 }
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -28,6 +28,69 @@
 
 using namespace mococrw;
 
+std::vector<uint8_t> rsaSign(const AsymmetricPrivateKey &privKey,
+                             const DigestTypes digestType,
+                             std::shared_ptr<RSASignaturePadding> padding,
+                             const std::vector<uint8_t> &message)
+{
+    std::shared_ptr<MessageSignatureCtx> signCtx;
+
+    try {
+        /* Padding is optional. Default: PSSPadding with MGF1 as mask generation function.
+         * DigestType is used as hash function for the padding schemes and for MGF1 (if used) */
+        signCtx = std::make_shared<RSASignaturePrivateKeyCtx>(privKey, digestType, padding);
+    } catch (const MoCOCrWException &e) {
+        std::cerr << "Please check your RSA key. Failure creating context." << std::endl;
+        std::cerr << e.what();
+        exit(EXIT_FAILURE);
+    }
+
+    std::vector<uint8_t> signature;
+    try {
+        signature = signCtx->signMessage(message);
+    } catch (const MoCOCrWException &e) {
+        /* Possible reasons:
+         * - error in openssl (sign, padding, ...)
+         * - Hash function's digest size doesn't match the message's digest size
+         */
+        std::cerr << "Failure occurred during signing." << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+    return signature;
+}
+
+void rsaVerify(const AsymmetricPublicKey &pubKey,
+               std::shared_ptr<RSASignaturePadding> rsaPadding,
+               DigestTypes digestType,
+               const std::vector<uint8_t> &signature,
+               const std::vector<uint8_t> &message)
+{
+    std::shared_ptr<MessageVerificationCtx> verifyCtx;
+    try {
+        /* Padding is optional. Default: PSSPadding with MGF1 as mask generation function.
+         * DigestType is used as hash function for the padding schemes and for MGF1 (if used) */
+        verifyCtx = std::make_shared<RSASignaturePublicKeyCtx>(pubKey, digestType, rsaPadding);
+    } catch (const MoCOCrWException &e) {
+        std::cerr << "Please check your RSA key. Failure creating context." << std::endl;
+        std::cerr << e.what();
+        exit(EXIT_FAILURE);
+    }
+
+    try {
+        verifyCtx->verifyMessage(signature, message);
+    } catch (const MoCOCrWException &e) {
+        /* Possible reasons:
+         * - error in openssl (sign, padding, ...)
+         * - Hash function's digest size doesn't match the message's digest size
+         * - Invalid signature
+         */
+        std::cerr << "Verification failed!" << std::endl;
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
+
 std::vector<uint8_t> ecdsaSign(const AsymmetricPrivateKey &privKey,
                                const DigestTypes digestType,
                                const ECDSASignatureFormat sigFormat,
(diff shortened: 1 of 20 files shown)


examples/hsm-example.cpp
```diff
@@ -26,20 +28,94 @@
...
+        exit(EXIT_FAILURE);
+    }
+}
 int main(void)
 {
     // Information for engine loading and key management.
     std::string id("pkcs11");
     std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
     std::string pin("1234");
     std::string keyID("1001");
```
johnfxgalea-e, line 98: This is no longer needed in this example.

johnfxgalea-e: This is no longer needed in this example.

examples/hsm-example.cpp
```diff
@@ -26,20 +28,94 @@
...
-
-    // Step 2: Initialise the PKCS11 Engine.
+    std::vector<uint8_t> message = utility::fromHex("deadbeef");
     HsmEngine hsmEngine(id, modulePath, pin);
 
-    // Step 4: Load the key from the HSM and print it.
-    auto pubKey = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyID);
```
johnfxgalea-e, line 41: Why not create a new example, in addition to this existing one, so that we still keep an example of using readPublicKeyFromHSM() for key loading?

Maybe create an file: "examples/hsm-gen-example.cpp", and include your example there?

examples/hsm-example.cpp
```diff
@@ -26,20 +28,94 @@
...
+    auto signature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
+
+    /* we can use here the private key, as it also contains the public key.
+     * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
+     * we do an implicit upcast here.
+     */
+    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, signature, message);
+    /*********************************************/
 
     return 0;
```
johnfxgalea-e, line 229: Nit: Maybe output something to stdout indicating success if all goes well?

src/hsm.cpp
```diff
@@ -45,4 +47,75 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
     return _ENGINE_load_private_key(_engine.get(), keyID);
 }
 
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::genKeyGetPublic(const RSASpec &spec,
```
johnfxgalea-e: Keep function names simple and intuitive. s/genKeyGetPublic/genPublicKey

src/hsm.cpp
```diff
@@ -45,4 +47,75 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
...
+                                                     const std::string &keyID,
+                                                     const std::string &tokenLabel,
+                                                     const std::string &keyLabel)
+{
+    genKey(spec, keyID, tokenLabel, keyLabel);
+    return loadPublicKey(keyID);
+}
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::genKeyGetPublic(const ECCSpec &spec,
```
johnfxgalea-e: Ditto

src/hsm.cpp
```diff
@@ -45,4 +47,75 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
...
+                                                     const std::string &keyID,
+                                                     const std::string &tokenLabel,
+                                                     const std::string &keyLabel)
+{
+    genKey(spec, keyID, tokenLabel, keyLabel);
+    return loadPublicKey(keyID);
+}
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::genKeyGetPrivate(const RSASpec &spec,
```
johnfxgalea-e: Ditto

src/hsm.cpp
```diff
@@ -45,4 +47,75 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
...
+                                                      const std::string &keyID,
+                                                      const std::string &tokenLabel,
+                                                      const std::string &keyLabel)
+{
+    genKey(spec, keyID, tokenLabel, keyLabel);
+    return loadPrivateKey(keyID);
+}
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::genKeyGetPrivate(const ECCSpec &spec,
```
johnfxgalea-e: Ditto

src/key.cpp
```diff
@@ -86,7 +86,7 @@ AsymmetricKeypair AsymmetricKeypair::generateEd25519()
     return generate(spec);
 }
 
-AsymmetricKeypair AsymmetricKeypair::generate(const AsymmetricKey::Spec &keySpec)
+AsymmetricKeypair AsymmetricKeypair::generate(const AsymmetricKey::Spec& keySpec)
```
johnfxgalea-e: Why are we getting these weird code changes all of a sudden? Is clang-format acting funny?

Seems like it might be due to DerivePointerAlignment, which precedes PointerAlignment and instead tries to determine the preferred alignment by analysing the file.

This needs to be fixed in a separate PR. I'll handle it.

src/key.cpp
```diff
@@ -107,15 +107,35 @@ AsymmetricPublicKey AsymmetricPublicKey::readPublicKeyFromPEM(const std::string
...
 
 #ifdef HSM_ENABLED
-AsymmetricPublicKey AsymmetricPublicKey::readPublicKeyFromHSM(HSM &hsm, const std::string &keyID)
+AsymmetricPublicKey AsymmetricPublicKey::readPublicKeyFromHSM(HSM& hsm, const std::string& keyID)
 {
     auto key = hsm.loadPublicKey(keyID);
     return AsymmetricPublicKey{std::move(key)};
 }
+
+AsymmetricPublicKey AsymmetricPublicKey::genKeyOnHsmGetPublic(HSM& hsm,
```
johnfxgalea-e: s/genKeyOnHsmGetPublic/genPublicKeyOnHsm

src/key.cpp
```diff
@@ -107,15 +107,35 @@ AsymmetricPublicKey AsymmetricPublicKey::readPublicKeyFromPEM(const std::string
...
+                                                              const ECCSpec& spec,
+                                                              const std::string& keyID,
+                                                              const std::string& tokenLabel,
+                                                              const std::string& keyLabel)
+{
+    auto key = hsm.genKeyGetPublic(spec, keyID, tokenLabel, keyLabel);
+    return AsymmetricPublicKey{std::move(key)};
+}
+
+AsymmetricPublicKey AsymmetricPublicKey::genKeyOnHsmGetPublic(HSM& hsm,
```
johnfxgalea-e: Ditto

src/key.cpp
```diff
@@ -186,11 +206,31 @@ AsymmetricKeypair AsymmetricKeypair::readPrivateKeyFromPEM(const std::string &pe
...
 
 #ifdef HSM_ENABLED
-AsymmetricKeypair AsymmetricKeypair::readPrivateKeyFromHSM(HSM &hsm, const std::string &keyID)
+AsymmetricKeypair AsymmetricKeypair::readPrivateKeyFromHSM(HSM& hsm, const std::string& keyID)
 {
     auto key = hsm.loadPrivateKey(keyID);
     return AsymmetricKeypair{std::move(key)};
 }
+
+AsymmetricKeypair AsymmetricKeypair::genKeyOnHsmGetPrivate(HSM& hsm,
```
johnfxgalea-e: s/genKeyOnHsmGetPrivate/genPrivateKeyOnHsm

src/key.cpp
```diff
@@ -186,11 +206,31 @@ AsymmetricKeypair AsymmetricKeypair::readPrivateKeyFromPEM(const std::string &pe
...
+                                                           const ECCSpec& spec,
+                                                           const std::string& keyID,
+                                                           const std::string& tokenLabel,
+                                                           const std::string& keyLabel)
+{
+    auto key = hsm.genKeyGetPrivate(spec, keyID, tokenLabel, keyLabel);
+    return AsymmetricKeypair{std::move(key)};
+}
+
+AsymmetricKeypair AsymmetricKeypair::genKeyOnHsmGetPrivate(HSM& hsm,
```
johnfxgalea-e: Ditto

src/mococrw/key.h
```diff
@@ -92,6 +92,18 @@ class AsymmetricPublicKey
      * object as a result.
      */
     static AsymmetricPublicKey readPublicKeyFromHSM(HSM& hsm, const std::string& keyID);
+
+    static AsymmetricPublicKey genKeyOnHsmGetPublic(HSM& hsm,
```
johnfxgalea-e: Kindly add doxygen comments for this function, and the other new ones.

tests/unit/openssl_lib_mock.h
```diff
@@ -35,6 +35,13 @@ namespace openssl
 class OpenSSLLibMockInterface
 {
 public:
+    virtual const char* SSL_EC_curve_nid2nist(int nid) = 0;
```
johnfxgalea-e: Where are the actual tests that use the mock? I would expect them to be added in MoCOCrW/tests/unit/test_hsm.cpp

johnfxgalea-e (changes requested):
Main Points:
1) Add Doxygen comments
2) Include mock tests
3) Make function names more concise and clear
4) Add new example, rather than replacing existing one.

There is an issue concerning clang-format and pointer alignment. I will fix this in the meantime.

examples/hsm-example.cpp
```diff
@@ -26,20 +28,94 @@
...
+    std::vector<uint8_t> message = utility::fromHex("deadbeef");
     HsmEngine hsmEngine(id, modulePath, pin);
 
-    // Step 4: Load the key from the HSM and print it.
-    auto pubKey = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyID);
-    std::cout << "Loaded key: \n\n" << pubKey.publicKeyToPem() << "\n\n" << std::endl;
+    /************** ECDSA signature **************/
+    auto ecdsaDigestType = DigestTypes::SHA512;
+    ECCSpec ecspec;
+    auto eccPrivKey = AsymmetricPrivateKey::genKeyOnHsmGetPrivate(
```
johnfxgalea-e: In addition, also include an RSA example for completeness? (I know EC is more important, but would be nice to have both, especially since these examples also serve as additional tests).

examples/hsm-example.cpp
garw: Yep. I also think adding this on top would be better. I'm fine with reusing this file though - it can demonstrate several things.

src/mococrw/key.h
garw: Keep the interface the same as for regular keys.
generateOnHsm(keyspec) or generateEccOnHsm(spec) generateRsaOnHsm(spec)
they always return a a private key. If you want to get a public key, you get this from the private key object. I don't see why the HSM should a have special generate method here that returns public keys. It is not there for normal keys and not needed.

garw: Why do we have a key generation method in the pubkey class? From a logical point of view, only a keypair can be created, not a pubkey.

src/mococrw/key.h
```diff
@@ -271,6 +283,17 @@ class AsymmetricKeypair : public AsymmetricPublicKey
      * object as a result.
      */
     static AsymmetricKeypair readPrivateKeyFromHSM(HSM& hsm, const std::string& keyID);
+
+    static AsymmetricKeypair genKeyOnHsmGetPrivate(HSM& hsm,
```
garw: s.a.

src/openssl_wrap.cpp
```diff
@@ -1597,5 +1597,16 @@ void _ENGINE_finish(ENGINE *e)
...
 
+void _ENGINE_ctrl_cmd(ENGINE *e, const std::string &cmdName, void *p)
+{
+    OpensslCallIsOne::callChecked(
+            lib::OpenSSLLib::SSL_ENGINE_ctrl_cmd, e, cmdName.c_str(), 0, p, nullptr, 1);
+}
+
+std::string _EC_curve_nid2nist(int nid)
+{
+    return std::string{lib::OpenSSLLib::SSL_EC_curve_nid2nist(nid)};
```
garw: We probably want error handling here and throw if the lookup wasn't successful?

To make things worse: std::string{nullptr} is undefined behavior, see https://en.cppreference.com/w/cpp/string/basic_string/basic_string case (5)

src/mococrw/hsm.h
```diff
@@ -53,6 +55,36 @@ class HSM
...
+                                                       const std::string &keyID,
+                                                       const std::string &tokenLabel,
+                                                       const std::string &keyLabel) = 0;
+
+    virtual openssl::SSL_EVP_PKEY_Ptr genKeyGetPrivate(const ECCSpec &spec,
+                                                       const std::string &keyID,
+                                                       const std::string &tokenLabel,
+                                                       const std::string &keyLabel) = 0;
+
+    virtual void genKey(const RSASpec &spec,
```
garw: same as for the key objects, We don't need to abbreviate names. This should be called `generateKey`

src/hsm.cpp
```diff
@@ -45,4 +47,75 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
...
+    genKey(spec, keyID, tokenLabel, keyLabel);
+    return loadPrivateKey(keyID);
+}
+
+void HsmEngine::genKey(const RSASpec &spec,
+                       const std::string &keyID,
+                       const std::string &tokenLabel,
+                       const std::string &keyLabel)
+{
+    PKCS11_RSA_KGEN rsa;
```
garw: please more verbose names, this looks like `pkcs11_rsa_spec`

src/hsm.cpp
```diff
@@ -45,4 +47,75 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
...
+    rsa.bits = spec.numberOfBits();
+    PKCS11_KGEN_ATTRS kg;
```
garw: s.a.

src/hsm.cpp
```diff
@@ -45,4 +47,75 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
...
+
+    _ENGINE_ctrl_cmd(_engine.get(), "KEYGEN", &kg);
+}
+
+void HsmEngine::genKey(const ECCSpec &spec,
+                       const std::string &keyID,
+                       const std::string &tokenLabel,
+                       const std::string &keyLabel)
+{
+    PKCS11_EC_KGEN ec;
```
garw: s.a.

src/mococrw/hsm.h
```diff
@@ -53,6 +55,36 @@ class HSM
      * @param keyID The ID of the private key to load.
      */
     virtual openssl::SSL_EVP_PKEY_Ptr loadPrivateKey(const std::string &keyID) = 0;
+
+    virtual openssl::SSL_EVP_PKEY_Ptr genKeyGetPublic(const RSASpec &spec,
```
garw: This should have doxygen user documentation here. What do the parameters mean? What kind of input is allowed? How does this translate into PKCS#11 as we have PKCS#11 in mind. What would we do with it otherwise? How do we explain that the load functions just support a key id? Should it maybe support more? At least the label?

src/mococrw/hsm.h
garw: The normal key generation flow is, generate and get the private key. Why do we have these void functions and the other special get variants?

examples/hsm-example.cpp
```diff
@@ -26,20 +28,94 @@
...
+
+    /* The argument hashFunction is optional. Default is SHA256
+     * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
+    auto signature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
+
+    /* we can use here the private key, as it also contains the public key.
+     * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
+     * we do an implicit upcast here.
+     */
+    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, signature, message);
```
garw: What I'm missing here is: get the private key and write out the public key as pem. Then reload it from PEM and use this for validation. This is what will (implicitly) happen a lot of times on our platforms.

tests/unit/openssl_lib_mock.h
garw: I agree here. We currently don't have any tests for the HSM code except for the example - which (iirc) also executed as a test but maybe a bit misleading.
This is actually a good place where this mock layer here comes in handy. You can actually write a test for your HSM class by mocking the openssl interface and then checking that is submits the right call to SSL_ENGINE_ctrl_cmd and even verify if the parameters look sane-ish.

src/openssl_wrap.cpp
```diff
@@ -1597,5 +1597,16 @@ void _ENGINE_finish(ENGINE *e)
     OpensslCallIsOne::callChecked(lib::OpenSSLLib::SSL_ENGINE_finish, e);
 }
 
+void _ENGINE_ctrl_cmd(ENGINE *e, const std::string &cmdName, void *p)
+{
+    OpensslCallIsOne::callChecked(
+            lib::OpenSSLLib::SSL_ENGINE_ctrl_cmd, e, cmdName.c_str(), 0, p, nullptr, 1);
+}
+
+std::string _EC_curve_nid2nist(int nid)
```
garw, line 1606: This function may be interesting also to others. Maybe we want to offer this also in the public interface as static method on ECC Keyspec?

garw (changes requested):

examples/hsm-example.cpp
wusto: Done

examples/hsm-example.cpp
wusto: I guess that one example is enough here. Otherwise this is lots of copy and paste from other example.cpp's.

I added some lines how to use it with RSA.

examples/hsm-example.cpp
wusto: Done

examples/hsm-example.cpp
wusto: We don't have this for our other examples. Usually everything crashes when an exception is raised.

src/hsm.cpp
wusto: Done

src/hsm.cpp
wusto: Done

src/hsm.cpp
wusto: Done

examples/hsm-example.cpp
istepic: Done in the same file to demonstrate the API in one place.

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/key.cpp
istepic: Done

src/key.cpp
istepic: Done

src/key.cpp
istepic: Done

src/key.cpp
istepic: Done

src/mococrw/key.h
istepic: Done

tests/unit/openssl_lib_mock.h
istepic: All code paths are covered by the examples. That's why I initially decided not to write unit tests, especially since example tests the functionality E2E and has more chance of reporting a problem.

src/mococrw/key.h
istepic: Done

istepic: Done

src/mococrw/key.h
istepic: Done

src/openssl_wrap.cpp
istepic: Done

src/mococrw/hsm.h

src/mococrw/hsm.h
istepic: Done. Still wondering if error reporting makes sense. Lots of things happen in libp11 and there's no clean way to pinpoint what went from MoCOCrW.

src/mococrw/hsm.h
istepic: Attempt to remove code duplication. Not needed anymore as API is simplified

tests/unit/openssl_lib_mock.h
istepic: Added some unit tests to test_opensslwrapper.cpp

istepic: Done-ish

src/openssl_wrap.cpp
istepic: Done

examples/hsm-example.cpp
```diff
@@ -26,20 +28,123 @@
...
+    auto eccPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
+            hsmEngine, ecspec, keyIDEcc, "token-label", "DobarKey");
+    auto ecdsaSigFormat = ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS;
+
+    /* The argument hashFunction is optional. Default is SHA256
+     * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
+    auto signature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
+
+    /* Use-Case 1: You want to check your own signature:
+     * we can use here the private key, as it also contains the public key.
```
johnfxgalea-e: Nit: s/w/W

src/mococrw/hsm.h
```diff
@@ -64,6 +94,16 @@ class HsmEngine : public HSM
     HsmEngine(const std::string &id, const std::string &modulePath, const std::string &pin);
     virtual ~HsmEngine();
 
+    openssl::SSL_EVP_PKEY_Ptr generateKey(const RSASpec &spec,
```
johnfxgalea-e: Why are these made public, in contrast to their protected access mod settings in HSM?

src/mococrw/openssl_wrap.h
```diff
@@ -1560,5 +1560,20 @@ SSL_EVP_PKEY_Ptr _ENGINE_load_public_key(ENGINE *e, const std::string &keyId);
  */
 void _ENGINE_finish(ENGINE *e);
 
+/**
+ * Send control command to an engine and pass arbitrary data through \p p
+ */
+void _ENGINE_ctrl_cmd(ENGINE *e, const std::string &cmdName, void *p);
+
+/**
+ * Convert integer nid to curve name
```
johnfxgalea-e: You missed \p for nid parameter:

tests/unit/test_hsm.cpp
```diff
@@ -103,3 +110,43 @@ TEST_F(HSMTest, testHSMEngine)
...
+{
+    std::string engineID("engine_id");
+    std::string modulePath("/test_path.so");
+    std::string pin("1234");
+    ECCSpec eccSpec;
+    int curve = int(mococrw::openssl::ellipticCurveNid::PRIME_256v1);
+    auto engine = ::testutils::someEnginePtr();
+    auto pkey = ::testutils::somePkeyPtr();
+
+    EXPECT_CALL(_mock(), SSL_ENGINE_by_id(StrEq(engineID.c_str())))
```
johnfxgalea-e: Seems like we can reduce code duplication as a lot of the code is shared with HSMTest.

A helper function would be useful as I predict this code is going to be used also for other tests in the future.

tests/unit/test_key.cpp
```diff
@@ -310,6 +310,51 @@ TEST_F(KeyHandlingTests, testKeyLoadPrivKeyFromHSM)
...
+TEST_F(KeyHandlingTests, testHSMKeyGenerationInvalidKeyId)
+{
+    ECCSpec eccSpec;
+    HSMMock hsmMock;
+    EXPECT_THROW(AsymmetricKeypair::generateKeyOnHsm(
+                         hsmMock, eccSpec, "100z", "token-label", "key-label"),
+                 MoCOCrWException);
+}
+
+TEST_F(KeyHandlingTests, testHSMKeyGenerationValidKeyId)
```
johnfxgalea-e: Nit: A test just like this but for RSA would be good for better coverage.

johnfxgalea-e (changes requested):

tests/unit/test_hsm.cpp
```diff
@@ -79,7 +87,36 @@ void HSMTest::SetUp()
     // TODO: Get rid of the uninteresting calls by default here somehow...
 }
 
-void HSMTest::TearDown() { openssl::OpenSSLLibMockManager::destroy(); }
+void HSMTest::TearDown()
```
johnfxgalea-e: Cannot you use this new helper `void HSMTest::initialiseEngine()` also for testHSMEngine now?

src/mococrw/openssl_wrap.h
```diff
@@ -1571,9 +1571,9 @@ void _ENGINE_ctrl_cmd(ENGINE *e, const std::string &cmdName, void *p);
 std::string _EC_curve_nid2nist(int nid);
 
 /**
- * Get number of bits key has from EVP_PKEY data structure
+ * Get number of bits key has from \p to EVP_PKEY data structure
```
johnfxgalea-e: I think you need to remove `to`, no?

johnfxgalea-e (changes requested):
Last few things

src/mococrw/openssl_wrap.h
johnfxgalea-e: @istepic I think you missed this one

examples/hsm-example.cpp
istepic: Done

src/mococrw/hsm.h

tests/unit/test_hsm.cpp
istepic: Done

tests/unit/test_hsm.cpp
istepic: Yeah. That makes testHSMEngine test redundant so I deleted it. initialiseEngine and testHSMEngine both test only the constructor.

src/mococrw/openssl_wrap.h
istepic: Done

tests/unit/test_key.cpp
istepic: Done

src/mococrw/openssl_wrap.h
istepic: Not sure what you mean

istepic: Done

johnfxgalea-e (approved):
LGTM

src/hsm.cpp
```diff
@@ -35,14 +37,55 @@ HsmEngine::HsmEngine(const std::string &id, const std::string &modulePath, const
...
     return _ENGINE_load_public_key(_engine.get(), keyID);
 }
 
-openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID)
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyID) const
 {
     return _ENGINE_load_private_key(_engine.get(), keyID);
 }
 
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
```
garw, line 84: this shouldn't be const even if it may work syntactically. Just from the idea: Creating a key changes the state of the HSM.

src/mococrw/hsm.h
```diff
@@ -40,19 +42,47 @@ class HSM
...
+    /**
+     * @brief Generate a RSA key pair on the HSM
+     *
+     * @param spec The RSA specification @ref RSASpec
+     * @param keyID The key identifier
+     * @param tokenLabel The token label
+     * @param keyLabel The key label
+     * @return AsymmetricKeypair @ref AsymmetricKeypair
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr generateKey(const RSASpec &spec,
```
garw: This shouldn't be const even if it works syntactically here. Generating a key definately changes the state of the HSM.
Additionaly: Order of parameter. The key is placed into a token, so I would put the token-label first and then keyId/keylabel.

src/mococrw/hsm.h
```diff
@@ -40,19 +42,47 @@ class HSM
...
+    /**
+     * @brief Generate a ECC key pair on the HSM
+     *
+     * @param spec The ECC specification @ref ECCSpec
+     * @param keyID The key identifier
+     * @param tokenLabel The token label
+     * @param keyLabel The key label
+     * @return AsymmetricKeypair @ref AsymmetricKeypair
+     */
+    virtual openssl::SSL_EVP_PKEY_Ptr generateKey(const ECCSpec &spec,
```
garw: s.a.

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
      * Loads a private key from an HSM, creating an @ref AsymmetricPublicKey
      * object as a result.
      */
-    static AsymmetricKeypair readPrivateKeyFromHSM(HSM &hsm, const std::string &keyID);
+    static AsymmetricKeypair readPrivateKeyFromHSM(const HSM &hsm, const std::string &keyID);
+
+    /**
+     * @brief Generates RSA keypair on HSM token according to the spec given. The keys are fetched
```
garw: Only RSA? This looks now generic to me?

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+     * _only_ by key ID so generating multiple keys with the same ID must be avoided!
+     * @param hsm Initialized HSM engine handle
+     * @param spec @ref RSASpec
```
garw, line 288: How about ECC spec?

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+     * there
+     */
+    template <typename T>
+    static AsymmetricKeypair generateKeyOnHsm(const HSM &hsm,
+                                              const T &spec,
+                                              const std::string &keyID,
+                                              const std::string &tokenLabel,
+                                              const std::string &keyLabel)
+    {
+        static_assert(std::is_base_of<AsymmetricKey::Spec, T>::value,
```
garw: Templates are bad for ABI stability and I wonder if it's really worth it here.

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
      */
-    static AsymmetricKeypair readPrivateKeyFromHSM(HSM &hsm, const std::string &keyID);
+    static AsymmetricKeypair readPrivateKeyFromHSM(const HSM &hsm, const std::string &keyID);
+
+    /**
+     * @brief Generates RSA keypair on HSM token according to the spec given. The keys are fetched
+     * _only_ by key ID so generating multiple keys with the same ID must be avoided!
+     * @param hsm Initialized HSM engine handle
+     * @param spec @ref RSASpec
+     * @param keyID ID of the key on HSM. We use IDs to fetch keys from HSM. _Only_ hex values are
```
garw: Can we improve the documentation here a bit? This is one of the main entry points of a developer to the whole HSM setup, so it wouldn't hurt to explain some of the basics here: 
* IDs are the main identification mechanism we use (btw - why?, shouldn't we offer label as well?)
* IDs are internally big integers and we represent them here as hex strings
There is a vague statement about double IDs to be "avoided" what is this supposed to mean? What happens? Should we explicitly prevent this by checking if the ID is already taken?

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+     * valid! Generating 2 keys with the same ID must be avoided. This is not prohibited by PKCS#11
+     * standard but since we currently only fetch by key ID, there is no way of ensuring that the
+     * correct key is fetched.
+     * @param tokenLabel Label of the token on HSM. This determines where they key shall be stored.
```
garw: Same as above: Let's be a bit more generous regarding information here.
* One HSM may support multiple (separated) storage locations that we call token
* This parameter selects the token to put the key into

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+                                              const std::string &tokenLabel,
+                                              const std::string &keyLabel)
+    {
+        static_assert(std::is_base_of<AsymmetricKey::Spec, T>::value,
+                      "T must be a subclass of AsymmetricKey::Spec");
+        // libp11 uses 128 byte buffer
+        if (keyID.length() > 127 || !std::all_of(keyID.begin(), keyID.end(), [](unsigned char c) {
+                return std::isxdigit(c);
+            })) {
+            throw MoCOCrWException(std::string("invalid keyID"));
```
garw: More verbose error description: What does a valid keyID entail? Only a-f0-9, at max 127 letters. 
In addition we probably want to enforce that its an even amount of letters, otherwise this does not lead to full bytes (to mention also in documentation?)

src/mococrw/util.h
```diff
@@ -23,7 +23,7 @@
 
 #include <openssl/crypto.h>
 
-#include "mococrw/error.h"
+#include "error.h"
```
garw: Why?

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+     * @param tokenLabel Label of the token on HSM. This determines where they key shall be stored.
+     * @param keyLabel Arbitrary key label
+     * @return AsymmetricKeypair @ref AsymmetricKeypair
+     * @throw MoCOCrWException Since most of the logic is happening outside of OpenSSL and inside
+     * libp11 and HSM module implementation, exception's what() tries to list the most common things
+     * that could go wrong. libp11 does log some things to stderr, check if there's more context
+     * there
+     */
+    template <typename T>
+    static AsymmetricKeypair generateKeyOnHsm(const HSM &hsm,
```
garw: Follow-up here: When the generation method won't be const anymore, it can't be const here.

examples/hsm-example.cpp
```diff
@@ -26,20 +28,123 @@
...
+    /* Use-Case 1: You want to check your own signature.
+     * We can use here the private key, as it also contains the public key.
+     * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
+     * we do an implicit upcast here.
+     */
+    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, signature, message);
+
+    /************** Get pubkey from HSM, store it in PEM file for later use **************/
+    /* Use-Case 2: The public key used for verification is stored in the HSM */
+    auto pubKeyEcc = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDEcc);
```
garw: Do I really need to get it separately or could I just call publicKeyToPem() on the private key?

src/hsm.cpp
```diff
@@ -17,6 +17,8 @@
  * #L%
  */
 #include "mococrw/hsm.h"
+#include "libp11.h"
```
garw, line 24: Include order: Our headers first to prevent "hidden" compile errors because some definitions have been always implicitly been available to the header.
Among ours: preferably alphabetic if there is no technical reason against it?

src/hsm.cpp
```diff
@@ -35,14 +37,55 @@ HsmEngine::HsmEngine(const std::string &id, const std::string &modulePath, const
...
 {
     return _ENGINE_load_private_key(_engine.get(), keyID);
 }
 
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
+                                                 const std::string &keyID,
+                                                 const std::string &tokenLabel,
+                                                 const std::string &keyLabel) const
+{
+    PKCS11_RSA_KGEN pkcs11_rsa_spec;
```
garw: naming convention of our code is camelCase.

src/hsm.cpp
```diff
@@ -35,14 +37,55 @@ HsmEngine::HsmEngine(const std::string &id, const std::string &modulePath, const
...
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const ECCSpec &spec,
+                                                 const std::string &keyID,
+                                                 const std::string &tokenLabel,
+                                                 const std::string &keyLabel) const
+{
+    PKCS11_EC_KGEN pkcs11_ec_spec;
+    std::string curve{};
+    try {
+        curve = _EC_curve_nid2nist(int(spec.curve()));
```
garw: This could now be just `spec.curveName()`?!

src/hsm.cpp
```diff
@@ -35,14 +37,55 @@ HsmEngine::HsmEngine(const std::string &id, const std::string &modulePath, const
...
+    return loadPrivateKey(keyID);
+}
+
+openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const ECCSpec &spec,
+                                                 const std::string &keyID,
+                                                 const std::string &tokenLabel,
+                                                 const std::string &keyLabel) const
+{
+    PKCS11_EC_KGEN pkcs11_ec_spec;
+    std::string curve{};
```
garw:
```suggestion
    std::string curve;
```
We just don't have this style in our code usually.

examples/hsm-example.cpp
```diff
@@ -26,20 +28,123 @@
...
+    /************** RSA key generation and loading **************/
+    std::string keyIDRsa("8890");
+    mococrw::RSASpec rsaSpec;
+    auto rsaPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
+            hsmEngine, rsaSpec, keyIDRsa, "token-label", "BarfoKey");
+
+    /* Read public key from HSM */
+    auto pubKeyRsa = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDRsa);
+
+    /* Do whatever you want using the private and public key. See rsa-example.cpp or
```
garw: Considering that running the examples is the simplest form of integration test that we so/have. Add some actual example code here even if it is not really required to demonstrate anything?

src/mococrw/hsm.h
```diff
@@ -40,19 +42,47 @@ class HSM
     friend class AsymmetricKeypair;
 
 protected:
+    /**
+     * @brief Generate a RSA key pair on the HSM
+     *
+     * @param spec The RSA specification @ref RSASpec
```
garw: Not sure how much value is in super repetitive param descriptions. If they speak for themselves, just drop them. Otherwise, see that some value is actually added.

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
      * Loads a private key from an HSM, creating an @ref AsymmetricPublicKey
      * object as a result.
      */
-    static AsymmetricKeypair readPrivateKeyFromHSM(HSM &hsm, const std::string &keyID);
+    static AsymmetricKeypair readPrivateKeyFromHSM(const HSM &hsm, const std::string &keyID);
+
+    /**
+     * @brief Generates RSA keypair on HSM token according to the spec given. The keys are fetched
+     * _only_ by key ID so generating multiple keys with the same ID must be avoided!
+     * @param hsm Initialized HSM engine handle
```
garw: What is initialized supposed to mean in this context here? Does this refer to some state of the HSM object? What does the caller need to do? If nothing, just drop it to avoid confusion.

src/mococrw/key.h
```diff
@@ -270,7 +270,51 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+                return std::isxdigit(c);
+            })) {
+            throw MoCOCrWException(std::string("invalid keyID"));
+        }
+        try {
+            auto key = hsm.generateKey(spec, keyID, tokenLabel, keyLabel);
+            return AsymmetricKeypair{std::move(key)};
+        } catch (const openssl::OpenSSLException &e) {
+            throw MoCOCrWException(
+                    std::string("Key generation failed for unknown reason. Likely reasons: invalid "
```
garw: Not sure if the "likely reasons" are actually helpful. Feels a bit like just blaming others while we don't actually know anything helpful.

src/openssl_wrap.cpp
```diff
@@ -1597,5 +1597,22 @@ void _ENGINE_finish(ENGINE *e)
...
+{
+    auto nist = OpensslCallPtr::callChecked(lib::OpenSSLLib::SSL_EC_curve_nid2nist, nid);
+    return std::string(nist);
```
garw: minor: `std::string()` shouldn't be required here. The constructor from `const char*` isn't explicit.

tests/unit/test_hsm.cpp
```diff
@@ -74,32 +83,45 @@ void HSMTest::SetUp()
...
+TEST_F(HSMTest, testHSMKeygen)
+{
+    ECCSpec eccSpec;
+    int curve = int(mococrw::openssl::ellipticCurveNid::PRIME_256v1);
+    auto engine = ::testutils::someEnginePtr();
+    auto pkey = ::testutils::somePkeyPtr();
+    auto hsm = initialiseEngine();
+    EXPECT_CALL(_mock(), SSL_EC_curve_nid2nist(curve)).WillOnce(Return("P-256"));
+    EXPECT_CALL(_mock(),
+                SSL_ENGINE_ctrl_cmd(engine, StrEq("KEYGEN"), 0 /*non-optional*/, _, nullptr, 1))
```
garw, line 124: If we would want to be more thorough, we would also match the struct that is passed to the engine command. But I guess for now its ok to have it like this. In particular, given that engines are deprecated and to be replaced by providers anyways.

tests/unit/test_key.cpp
```diff
@@ -373,6 +421,18 @@ TEST_F(KeyHandlingTests, testGetKeySpec)
     EXPECT_EQ(dynamic_cast<ECCSpec *>(Ed448Spec.get())->curve(), openssl::ellipticCurveNid::Ed448);
     EXPECT_EQ(dynamic_cast<ECCSpec *>(Ed25519Spec.get())->curve(),
               openssl::ellipticCurveNid::Ed25519);
+    EXPECT_EQ(dynamic_cast<ECCSpec *>(defaultSpec.get())->curveName(), "P-256");
+    EXPECT_EQ(dynamic_cast<ECCSpec *>(Sect571r1Spec.get())->curveName(), "B-571");
+    EXPECT_EQ(dynamic_cast<ECCSpec *>(Secp521Spec.get())->curveName(), "P-521");
+    /* These two curves are not working:
```
garw, line 419: This is to be expected as the Ed curves have a slightly different implementation in OpenSSL. Our implementations also look different. We shouldn't worry to much, or think about a better way to implement this function then…

tests/unit/test_opensslwrapper.cpp
```diff
@@ -340,3 +340,41 @@ TEST_F(OpenSSLWrapperTest, testEngineFinishThrowsException)
     EXPECT_CALL(_mock(), SSL_ENGINE_finish(engine)).WillOnce(Return(0));
     EXPECT_THROW(_ENGINE_finish(engine), OpenSSLException);
 }
+
+TEST_F(OpenSSLWrapperTest, testEngineCtrlCmd)
```
garw, line 344: I don't really think that we need tests for the wrap layer. At least we usually don't have this if the function just implements error handling using the CallChecked templates.

garw (changes requested):
Overall looks good. Yet, there is a list of minor issues to iron out.

examples/hsm-example.cpp
istepic: This is done in use-case 3. I cleaned it up a bit.

examples/hsm-example.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

src/mococrw/hsm.h

src/mococrw/hsm.h

src/mococrw/hsm.h

src/mococrw/key.h
istepic: Moved back to overloaded functions

src/mococrw/key.h
istepic: Done

src/mococrw/key.h
istepic: Done

src/mococrw/key.h
istepic: In theory we could offer a label identification as well. This is something we didn't really look into. Maybe because `key_id` is used for a function argument name that loads the keys but this `key_id` is obviously a more general concept.
EVP_PKEY *ENGINE_load_private_key(ENGINE *e, const char *key_id,
                                  UI_METHOD *ui_method, void *callback_data);
                                  
We can talk about this but this should into seperate PR.

src/mococrw/key.h
istepic: Done

src/mococrw/key.h
istepic: Done

src/mococrw/key.h
istepic: Reverted to overloaded functions

src/mococrw/key.h
istepic: This is a bit weird yes. I think we shouldn't restrict users to use keyID's that have odd number of hex characters. Standard linux tool for managing HSMs, pkcs11-tool, adds 4 zeros on the 4 MSbs. I added a little bit of code that loads a key with the ID 0x0123 if "123" ID is given. And a little bit of code that creates a key with the ID 0x0123 if "123" is given.

I think this is the sanest approach as we rely less on the implementation of lower-level modules.

src/mococrw/key.h
istepic: Done

src/mococrw/util.h
istepic: clangd was complaining for me

src/openssl_wrap.cpp
istepic: Done

tests/unit/test_opensslwrapper.cpp
istepic: Agreed but I guess these don't hurt much as well

src/mococrw/key.h
```diff
@@ -270,7 +271,55 @@ class AsymmetricKeypair : public AsymmetricPublicKey
      * Loads a private key from an HSM, creating an @ref AsymmetricPublicKey
      * object as a result.
      */
-    static AsymmetricKeypair readPrivateKeyFromHSM(HSM &hsm, const std::string &keyID);
+    static AsymmetricKeypair readPrivateKeyFromHSM(const HSM &hsm, const std::string &keyID);
```
dfuhrmann: Why is the key identified by tokenLabel, keyID and keyLabel in the generation API, but here only KeyID is used?

Would be cool to consolidate and minimize the number of artifacts you need to identify one key pair.

src/hsm.cpp
istepic: I changed this but clang-format prefers it this way for some reason.

src/mococrw/key.h
istepic: Good point. Key ID is the only artifact you need to identify a key pair. I guess adding keyLabel in generation API generates some confusion and we will probably create another PR with the intention to eliminate that confusion.

examples/hsm-example.cpp
```diff
@@ -26,20 +28,202 @@
...
+        exit(EXIT_FAILURE);
+    }
+}
+
 int main(void)
 {
     // Information for engine loading and key management.
     std::string id("pkcs11");
     std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
     std::string pin("1234");
```
ricleite, line 163: heh, should applications really hardcode this instead of `utility::cryptoRandomBytes`?

(@garw that would... ease some burdens ;) )

src/mococrw/key.h
```diff
@@ -270,7 +270,55 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+     * @param spec @ref RSASpec
+     * @param keyID ID of the key on HSM. _Only_ hex values are valid because key IDs are stored as
+     * raw bytes.
+     * @param tokenLabel Label of the token on HSM. An HSM can have multiple tokens where the keys
+     * are stored. This determines on which token the key shall be stored.
+     * @param keyLabel Arbitrary key label
+     * @return AsymmetricKeypair @ref AsymmetricKeypair
+     * @throw MoCOCrWException Since most of the logic is happening outside of OpenSSL and inside
+     * libp11 and HSM module implementation, we can't know exactly what went wrong. libp11 does log
+     * some things to stderr, check if there's more context there
```
ricleite, line 295: uff! not good

I would ask that the exceptions' `.what()` states the same thing that's here, as other mococrw exceptions always have useful info

(it's also a bit weird that libmococrw now does stdout/stderr due to this...)

src/mococrw/key.h
```diff
@@ -270,7 +270,55 @@ class AsymmetricKeypair : public AsymmetricPublicKey
...
+    /**
+     * @brief Generates RSA keypair on HSM token according to the spec given.
+     * @note libp11 supports fetching the keys only by keyID but generating multiple keys with the
+     * same key is not prohibited by PKCS#11 standard. Care should be taken when generating multiple
+     * keys with the same ID because there are no guarantees the correct one will be fetched given
+     * current libp11 implementation.
+     * @param hsm HSM engine handle
+     * @param spec @ref RSASpec
+     * @param keyID ID of the key on HSM. _Only_ hex values are valid because key IDs are stored as
+     * raw bytes.
```
ricleite: if only hex values are valid anyway, perhaps this shouldn't be an `std::string`

is there any hard reason to keep it like this? it is very easy as a user to make a mistake

ricleite: no limitations on the key size or encoding otherwise? weird, I'd expect an HSM to be a lot more picky

examples/hsm-example.cpp
```diff
@@ -26,20 +28,202 @@
...
+        std::cerr << e.what() << std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
+
 int main(void)
 {
     // Information for engine loading and key management.
     std::string id("pkcs11");
     std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
```
ricleite, line 160: TIL OpenSSL can `dlopen` stuff... interesting

examples/hsm-example.cpp
```diff
@@ -26,20 +28,202 @@
...
     std::string pin("1234");
-    std::string keyID("1001");
-
-    // Step 2: Initialise the PKCS11 Engine.
+    std::vector<uint8_t> message = utility::fromHex("deadbeef");
     HsmEngine hsmEngine(id, modulePath, pin);
```
ricleite: hm, the pin seems to be associated to the HSM, instead of the key itself

hypothetical scenario... how could an `application A` create a key in the HSM, and share it only with an `application B`, such that no `application C` (even if it can access the HSM as well) can use it?

it does not seem possible with this mococrw design, unless I'm missing something

ricleite (commented):
How/where to do HSM initialization? Or is this out of scope for mococrw?

examples/hsm-example.cpp
garw: I discussed the same thing with Ivan today. We'll move the "token" also into the HSMEngine. Then one HSMEngine represents on PKCS#11 token instead of the whole HSM. Then everything that belongs together token/PIN is in the same thing. This is not really clear from the current design.

examples/hsm-example.cpp
```diff
@@ -100,51 +163,67 @@ int main(void)
...
-    /************** RSA key generation and loading **************/
-    std::string keyIDRsa("8890");
+    /************** RSA key generation, loading and digital signatures **************/
+    std::string keyIDRsa("12");
+    auto mgf1DigestType = DigestTypes::SHA256;
     mococrw::RSASpec rsaSpec;
-    auto rsaPrivKey = AsymmetricPrivateKey::generateKeyOnHsm(
-            hsmEngine, rsaSpec, keyIDRsa, "token-label", "BarfoKey");
+    auto rsaSignatureDigestType = DigestTypes::SHA512;
+    auto mgf1 = std::make_shared<MGF1>(MGF1(mgf1DigestType));
```
garw: make_shared is a fowarding constructor. You shouldn't create the object itself in it. Otherwise, you don't really get the benefits of it and the object will most probably copy constructed/move constructed from your temporary object.

examples/hsm-example.cpp
```diff
@@ -100,51 +163,67 @@ int main(void)
...
+    int saltLength = mococrw::Hash::getDigestSize(rsaSignatureDigestType);
+    auto padding = std::make_shared<PSSPadding>(PSSPadding(mgf1, saltLength));
```
garw: s.a.

examples/hsm-example.cpp
```diff
@@ -100,51 +163,67 @@ int main(void)
...
     auto pubKeyRsa = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyIDRsa);
 
-    /* Do whatever you want using the private and public key. See rsa-example.cpp or
-     * sig-example.cpp for further examples.
+    /**
+     * Do signing/verification
+     */
+    auto RSAsignature = rsaSign(rsaPrivKey, rsaSignatureDigestType, padding, message);
+    /* We can use here the private key, as it also contains the public key.
+     * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
```
garw: Isn't is usually called generalitzation if you do an upcast? Anyways, I don't really think that this comment is needed. Downcasts are never implicit and need to be explicitly stated.

src/mococrw/key.h
garw: I kind of agree that this then maybe should better be a std::vector<uint8_t>

garw: This is probably implementation specific but not mentioned in the general PKCS#11 spec

src/mococrw/key.h
garw: see also Ricardo's comment.

src/mococrw/key.h
garw: Now that I thought about it, I do agree with Ricardo tbh. std::string is not the right datatype here.

src/mococrw/util.h
garw: I don't think that this is correct though.

garw (changes requested):
Let's briefly look through the remaining few findings and David's/Ricardo's comments tomorrow.

src/mococrw/key.h
dfuhrmann: @garw I understand that the HSM expect a certain amount of raw bytes as keyID. But for the user of this interface, I think it would be most convenient to be able to also pass a human readable string as a keyID, as this can be more easily handled / used on caller site.

Why not keep std::string as a data type then, and just change semantics? This data type is perfectly fine for both passing raw bytes, as well as (ASCII / UTF8) strings. So the caller could decide to pass a human readable string, and internally you just check for a maximum size of the data, and treat any bytes as raw data.

examples/hsm-example.cpp
istepic: Added a comment to  warn people not to hardcore pins

src/mococrw/key.h
istepic: Yeah, this is unfortunate but there's not much we can do about it as error handling between 3 independently developed libraries is always a mess. Not much is logged to stderr by libp11 though.

src/mococrw/key.h
istepic: This important point will be handled in a separate PR. I don't want it to get lost in this already huge PR so stay tuned.

examples/hsm-example.cpp
istepic: This will be handled in a separate PR. The scenario you described currently doesn't really work as you would expect.

examples/hsm-example.cpp
istepic: Done

examples/hsm-example.cpp
istepic: done

examples/hsm-example.cpp
istepic: Done

src/mococrw/key.h
dfuhrmann: All fine, thanks for the consideration.

src/mococrw/util.h
```diff
@@ -45,6 +45,15 @@ std::vector<uint8_t> fromHex(const std::string &hexData);
...
 std::vector<uint8_t> cryptoRandomBytes(size_t length);
 
+/**
+ * @brief Transforms given string to a string with percent encoding notation (see RFC 3986)
+ * i.e. "30AFF" -> "%03%0A%FF"
+ * @note This function does not enforce a string to contain only hex characters (only hex digits
+ * are used in percent encoding). If you have a different use case, change this or create your
+ * own function.
+ */
+void hexStringToPctEncoded(std::string &str);
```
garw: I don't think that this is a generally useful function. In URIs only non-ascii characters are escaped via this ampersand notation. The function here does no particular encoding but just adds an % every second character.  We should have this as a private function. Moreover, functions with side-effects are generally discouraged. This should be a string -> string function.

garw (changes requested):
mostly ok now w.r.t. what we discussed. Just this one utility function is a bit odd and shouldn't become part of our public interface.

garw (approved):
lg2m now

PR #122, opened 2022-12-09
Openssl1.1 dilithium

wusto:
As openssl is not supporting dilithium at the moment, the reference
implementation from the NIST PQC is used.

For this purpose the crystalls dilithium libraries without avx support
are linked.

The current version of MoCOCrW uses a EVP_PKEY centric implementation
for asymmetric cryptographic operations. To avoid changes on the
existing classes the relevant classes are cloned for the dilithium.

Functionality is added or removed wherever necessary. But the changes
were done with the plan in mind, to be able to change back to the
EVP_PKEY centric implementation once openssl supports dilithium.


CMakeLists.txt
```diff
@@ -31,6 +31,9 @@ set(SECURITY_COMPILER_FLAGS
 # By default, we do not build MoCOCrW with LibP11.
 option(HSM_ENABLED "Enable HSM features" OFF)
 
+# By default, we do not build MoCOCrW with LibP11.
```
garw: copy&pasta?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+public:
+    DilithiumKeyImpl(const std::vector<uint8_t> key_data, AsymmetricKey::KeyTypes key_type)
+            : _keyType(key_type), _key_data{std::move(key_data)}
+    {
+        if (!hasValidKeySize()) {
+            throw MoCOCrWException("Key data does not match the expected size.");
+        }
+    }
+
+    static std::shared_ptr<DilithiumKeyImpl> parseAsn1PublicKey(
```
garw: as previously discussed. For conformity with the other functions, I would rather like to have this as `fromDer` than `parseAsn1`. Strictly speaking ASN1 isn't a serialization in this sense.

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+            const std::vector<uint8_t> &pkcs8PrivKey);
+    /**
+     * @returns The key type
+     */
+    AsymmetricKey::KeyTypes getType() const { return _keyType; };
+
+    /**
+     * @returns The size of the raw key data
+     */
+    uint getKeySize() const noexcept { return _key_data.size(); }
```
garw: size_t? Why is this singular function noexcept here? If we start this, all functions that are actually noexcpt should be marked with it?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+     * @returns The raw key data
+     */
+    const std::vector<uint8_t> getKeyData() const { return _key_data; };
+
+    /**
+     * Returns the public key. If a private key is stored the public key is extracted and returned
+     *
+     * @return The public key
+     * @throws If the stored key size doesn't match the size of a private or public key
+     * @throws If the extraction functin to get the public key from a private key returns an error
```
garw: Run a spellchecker?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+     * @throws If the key type is not supported
+     */
+    DilithiumKeyImpl getPublicKey() const;
```
garw, line 100: hmhm. What is the purpose of this? Now we have DilithiumKeyImpl. Is this for public or for private keys or for both? How do we distinguish? How to prevent runtime errors?
Normally, I would say one (same) object for both is bad design as it shadows logic errors at compile time and moves them to runtime (which is usually a bad thing). As this is an intermediate solution, we can maybe live with it for a while if it is clear enough and hidden behind the other key objects.

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+     * @brief Checks if the key has a valid key size
+     *
+     * @return true if the key matches either the length of the private or public key of the given
+     * key type
+     * @return false otherwise
+     */
+    bool hasValidKeySize() const;
+
+    /**
+     * Checks if the key size matches the size of a private key. Does not do any validity checks.
```
garw: :(

I'm not sure if this is a sane way of handling this.

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+     * The key size is dependent on the selected key type (Dilithium3 or Dilithium5).
+     *
+     * @return True if the key size matches the size of private keys.
+     * @throws If the key size neither matches a private or public key size.
+     * @throws if the key type is not supported
+     */
+    bool isPrivateKey() const;
+
+    inline bool operator==(DilithiumKeyImpl &key) { return _key_data == key.getKeyData(); }
```
garw: parameter should be const. Moreover, we often use the convention in operations that the operand is called `rhs` for right hand-side of the operator.

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+{
+public:
+    DilithiumAsymmetricPublicKey(std::shared_ptr<DilithiumKeyImpl> key) : _key{std::move(key)} {}
+
+    static DilithiumAsymmetricPublicKey fromDER(std::vector<uint8_t> &asn1Data);
+
+    /**
+     * Getters for the internal dilithium implementation objet.
+     */
+    inline DilithiumKeyImpl *internal() { return _key._internal().get(); }
```
garw, line 198: Why does this return raw pointers and the private one shared pointers?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+ *
+ * @see AsymmetricKey
+ *
+ * This shall be replaced by AsymmetricKey once Dilithium is supported by the openssl version
+ * used by MoCOCrW.
+ * The interface is similar to the one of AsymmetricKey.
+ *
+ * For documentation see AsymmetricKey
+ */
+class DilithiumAsymmetricKey
```
garw, line 151: Hmhm. Isn't this class a lot of duplication? In the other code, the `AsymmetricKey` class is a thin wrapper of an OpenSSL key. Here we already have this `DilithiumKeyImpl`. Do we really need this in addition?

It also looks to me like duplication of the Keypair and the Pubkey.

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+/**
+ * @brief Specification for Dilithium keys
+ *
+ * This class allows to generate Dilithium keypairs.
+ *
+ */
+class DilithiumSpec final : public DilithiumAsymmetricKey::Spec
+{
+public:
+    static constexpr AsymmetricKey::KeyTypes defaultDilithiumType =
```
garw: what is constexpr supposed to do in this context here?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+    ~DilithiumSigningCtx();
+    /**
+     * @brief Copy Constructor
+     */
+    DilithiumSigningCtx(const DilithiumSigningCtx &other);
+
+    /**
+     * @brief Copy Assignment
+     */
+    DilithiumSigningCtx &operator=(const DilithiumSigningCtx &other);
```
garw, line 331: Do we really want those? Tbh, I would just delete everything except for the move constructor. Don't we do this for the other contexts?

src/mococrw/key.h
```diff
@@ -43,7 +42,7 @@ class AsymmetricKey
     /**
      * Supported asymmetric key types.
      */
-    enum class KeyTypes : int { RSA, ECC, ECC_ED };
+    enum class KeyTypes : int { RSA, ECC, ECC_ED, DILITHIUM2, DILITHIUM3, DILITHIUM5 };
```
garw: Are these algorithms really that substantially different? I aren't these just different kind of sizes/security parameters. I would just have one and then distinguish the different security levels in the creators. We just have one for all of ECC with its widely varying curves…

src/mococrw/openssl_wrap.h
```diff
@@ -1196,6 +1216,11 @@ SSL_X509_CRL_Ptr _PEM_read_bio_X509_CRL(BIO *bp);
  */
 SSL_X509_CRL_Ptr _d2i_X509_CRL_bio(BIO *bp);
 
+/**
+ * Reads a DER encoded CRL from a BIO object.
```
garw: CRL?

src/mococrw/openssl_wrap.h
```diff
@@ -355,6 +366,15 @@ void _EVP_PKEY_CTX_set_ec_paramgen_curve_nid(EVP_PKEY_CTX *ctx, int nid);
  */
 void _EVP_PKEY_CTX_set_ec_param_enc(EVP_PKEY_CTX *ctx, int param_enc);
 
+/**
+ * @brief Parses a private key ASN.1 structure (see RFC 5958 )
```
garw: DER-encoded?
space before `)`

tests/unit/test_dilithium.cpp
```diff
@@ -0,0 +1,1260 @@
...
+    auto ctx = DilithiumSigningCtx(privKey);
+    auto res = ctx.signMessage(message);
+
+    auto pub_key_raw = privKey._internal()->getPublicKey();
+    auto pub_key = DilithiumAsymmetricPublicKey(std::make_shared<DilithiumKeyImpl>(pub_key_raw));
+
+    auto a = pub_key._internal()->getPublicKey().getKeyData();
+    auto b = pub_key_raw.getKeyData();
+    // test getter for public Key
+    EXPECT_EQ(pub_key._internal()->getPublicKey().getKeyData(), pub_key_raw.getKeyData());
```
garw: What exactly does this test? Doesn't look like something that I would expect in a simple sign&verify test.

examples/sig-example.cpp
```diff
@@ -21,6 +21,10 @@
 #include <mococrw/hash.h>
 #include <iostream>
 
+#ifdef DILITHIUM_ENABLED
```
garw, line 25: Wouldn't it make more sense to move this into its own example file instead of this ifdef madness here?

examples/sig-example.cpp
```diff
@@ -252,4 +297,21 @@ int main(void)
...
     edDsaVerify(edPrivKey, signature, message);
     /*********************************************/
+
+#ifdef DILITHIUM_ENABLED
+    /************** dilithium signature **********/
+    /* Please note that the interface might change slightly once openssl officially supports
+     * dilithium
+     */
+    auto keySpec = DilithiumSpec(AsymmetricKey::KeyTypes::DILITHIUM3);
+    auto dilithiumKeypair = DilithiumAsymmetricPrivateKey::generate(keySpec);
```
garw, line 308: I guess the interesting part for our colleagues that should be demonstrates is how to load a key for verification from disk?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+#include "mococrw/dilithium.h"
```
garw: include order?

C++ best practise is usually: First your local things, then the standard includes (in order to prevent leaking definitions of the standard headers into your local headers by coincidence).

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+#include <memory>
+#include <vector>
+#include "mococrw/asymmetric_crypto_ctx.h"
+#include "mococrw/error.h"
+#include "mococrw/key.h"
+#include "mococrw/openssl_wrap.h"
+
+extern "C" {
+#include <dilithium-3.1/api.h>
+#include <openssl/asn1.h>
```
garw, line 33: Normally, this library discourages direct use of OpenSSL headers in the implementations. We discussed that this case is special but a comment to make this clear would be good here.

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+#include <openssl/objects.h>
+#include <openssl/x509.h>
+}
+
+namespace mococrw
+{
+using KeyTypes = AsymmetricKey::KeyTypes;
+
+/*
+ * The dilithium private key struct as returned by the backend
```
garw: *cough* This comment is an info leak. Remove it. Refer to the RFCs

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+ * (iaik: http://javadoc.iaik.tugraz.at/iaik_jce/current/iaik/pkcs/pkcs8/PrivateKeyInfo.html)
+ *
+ * This is the internal struct inside the OCTET_STRING
```
garw: Can you refer to the attributes by name instead of their type? 
```suggestion
 * This is the internal structure contained in the PrivateKey element of the PrivateKeyInfo structure.
```

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+    ASN1_OCTET_STRING *privKey;
+    ASN1_OCTET_STRING *pubKey;
+    ASN1_INTEGER *dilithiumForm;
+    ASN1_BOOLEAN bool1;
+    ASN1_BOOLEAN bool2;
+};
+
+typedef struct dilithium_privkey_st DILITHIUM_PRIV;
+
+// clang-format off
```
garw, line 69: Why is clang format off here? The current format also looks really weird.

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+                               DILITHIUM_PRIV_INTERNAL,
+                               dilithium_priv)
+
+using DILITHIUM_PRIV_Ptr = std::unique_ptr<DILITHIUM_PRIV,
+                                           openssl::SSLDeleter<DILITHIUM_PRIV, dilithium_priv_free>>;
+using DILITHIUM_PRIV_SharedPtr = utility::SharedPtrTypeFromUniquePtr<DILITHIUM_PRIV_Ptr>;
+// clang-format on
+
+/*
+ * The dilithium public key struct as returned by the backend
```
garw: s.a.

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+
+IMPLEMENT_ASN1_FUNCTIONS_fname(DILITHIUM_PUB, DILITHIUM_PUB_INTERNAL, dilithium_pub)
+
+using DILITHIUM_PUB_Ptr = std::unique_ptr<DILITHIUM_PUB,
+                                          openssl::SSLDeleter<DILITHIUM_PUB, dilithium_pub_free>>;
+using DILITHIUM_PUB_SharedPtr = utility::SharedPtrTypeFromUniquePtr<DILITHIUM_PUB_Ptr>;
+// clang-format on
+
+KeyTypes getKeyTypeFromDilithiumForm(ASN1_INTEGER *asn1Int)
```
garw:
```suggestion
KeyTypes getKeyTypeFromDilithiumForm(const ASN1_INTEGER *asn1Int)
```
?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+        const std::vector<uint8_t> &x509PubKey)
+{
+    const uint8_t *p = x509PubKey.data();
+    auto pubKey = openssl::_d2i_X509_PUBKEY(p, x509PubKey.size());
+
+    const unsigned char *pk = NULL;
+    int ppklen;
+    // Get the nested bit string
+    // we can use get0 here, as pk is not used outside of this function
+    // (d2i_dilithium_pub and the underlying ASN1_item_d2i_ex do an alloc according to the manpage)
```
garw: "do an alloc" This is super vague. What we need here is that the other functiony *copy* the result out of the parsed data. We do not care if they do some kind of allocation at point?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+    if (!X509_PUBKEY_get0_param(nullptr, &pk, &ppklen, nullptr, pubKey.get())) {
+        printf("Can't retreive public key data.\n");
```
garw: debug leftover?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+        throw MoCOCrWException("Cannot retreive public key bistring from x509 public key object.");
+    }
+
+    // parse the nested bitstring
+    DILITHIUM_PUB_Ptr nestedPubKey = DILITHIUM_PUB_Ptr(d2i_dilithium_pub(nullptr, &pk, ppklen));
```
garw: auto return type?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+    int ppklen;
+    // Get the nested bit string
+    // we can use get0 here, as pk is not used outside of this function
+    // (d2i_dilithium_pub and the underlying ASN1_item_d2i_ex do an alloc according to the manpage)
+    if (!X509_PUBKEY_get0_param(nullptr, &pk, &ppklen, nullptr, pubKey.get())) {
+        printf("Can't retreive public key data.\n");
+        throw MoCOCrWException("Cannot retreive public key bistring from x509 public key object.");
+    }
+
+    // parse the nested bitstring
```
garw: Comments that just state the same the code also states just clutter the code

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+    int ppklen;
+    // get the nested octet string
+    // we can use get0 here, as pk is not used outside of this function
+    // (d2i_dilithium_priv and the underlying ASN1_item_d2i_ex do an alloc according to the manpage)
+    if (!PKCS8_pkey_get0(nullptr, &pk, &ppklen, nullptr, p8inf.get())) {
+        throw MoCOCrWException("Cannot get private key data.\n");
+    }
+
+    // parse the nested octet string
+    DILITHIUM_PRIV_Ptr nestedPrivKey(d2i_dilithium_priv(nullptr, &pk, ppklen));
```
garw: style consistency with previous: Assignment vs constructor?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+        case KeyTypes::DILITHIUM3:
+            publicKeyGetterFunc = pqcrystals_dilithium3_ref_keypair_public_from_private;
+            break;
+        case KeyTypes::DILITHIUM5:
+            publicKeyGetterFunc = pqcrystals_dilithium5_ref_keypair_public_from_private;
+            break;
+        default:
+            throw MoCOCrWException("Unsupported key type");
+    }
+    if (publicKeyGetterFunc(public_key.data(), _key_data.data())) {
```
garw: if you just call the getter func in the switch, you can eliminate calling it here and the std::function definition. What is the purpose here? To me, It just makes the code harder to read.

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+    }
+    return DilithiumKeyImpl(public_key, _keyType);
+}
+
+bool DilithiumKeyImpl::hasValidKeySize() const
+{
+    if (getKeySize() != getPrivKeySize(_keyType) && getKeySize() != getPubKeySize(_keyType)) {
+        return false;
+    }
+    return true;
```
garw:
```suggestion
    return getkeySize() == getPrivKeySize(_keyType) || getKeySize() == getPubkeySize(_keyType)
```
However, I think I already commented this in the header. I find a function that does a limited syntactical check like that one here rather weird. One should at least know if one wants a public or a private key. This information should be available in all scopes when checking for the correct size is relevant?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+}
+
+bool DilithiumKeyImpl::isPrivateKey() const
+{
+    if (getKeySize() != getPrivKeySize(_keyType)) {
```
garw: Not sure this is a sane way of testing this.

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+DilithiumAsymmetricKeypair DilithiumAsymmetricKeypair::generate(
+        const DilithiumAsymmetricKey::Spec &spec)
+{
+    return DilithiumAsymmetricKeypair(spec.generate());
+}
+
+DilithiumAsymmetricKey DilithiumSpec::generate() const
+{
+    std::vector<uint8_t> privateKey(getPrivKeySize(_keyType));
+    // The public key is expected. Nullptr would lead to errors
```
garw: What does this mean? I think you want to tell me that the dilithium creation functions expect buffers for a private and a public key object?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+                break;
+            case KeyTypes::DILITHIUM5:
+                signature.resize(pqcrystals_dilithium5_BYTES);
+                sign_function = pqcrystals_dilithium5_ref_signature;
+                break;
+            default:
+                throw MoCOCrWException("Invalid key type for dilithium");
+        }
+
+        if (sign_function(
```
garw: Same as above.  Here it is imho even a bit more weird because parts of it (buffer resizing) are in the swich and then you call the sign function later? Imho it just makes the code harder to read because the task is split apart when reading.

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+                    reinterpret_cast<uint8_t *>(signature.data()),
```
garw: Why do we have these `reinterpret_cast`s here? `std::vector<uint8_t>::data()` returns a `uint8_t*`?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+            }
+            break;
+        default:
+            throw MoCOCrWException("The used dilithium key type is not supported.");
+    }
+
+    return DilithiumAsymmetricKey(std::make_shared<DilithiumKeyImpl>(privateKey, _keyType));
+}
+
+template <class Key>
```
garw: What is the purpose of this template? Do we need a ~10 lines template class to inherit a constructor and one member?

src/dilithium.cpp
```diff
@@ -0,0 +1,448 @@
...
+                verifyFunction = pqcrystals_dilithium3_ref_verify;
+                break;
+            case KeyTypes::DILITHIUM5:
+                verifyFunction = pqcrystals_dilithium5_ref_verify;
+                break;
+            default:
+                throw MoCOCrWException("Invalid key type for dilithium");
+        }
+
+        if (verifyFunction(signature.data(),
```
garw: see above.

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,347 @@
...
+    ~DilithiumVerificationCtx();
+    /**
+     * @brief Copy Constructor
+     */
+    DilithiumVerificationCtx(const DilithiumVerificationCtx &other);
+
+    /**
+     * @brief Copy Assignment
+     */
+    DilithiumVerificationCtx &operator=(const DilithiumVerificationCtx &other);
```
garw, line 377: s.a.

tests/unit/test_dilithium.cpp
```diff
@@ -0,0 +1,1260 @@
...
+
+TEST(DilithiumTest, CheckPrivateKeyGeneration)
+{
+    auto spec = DilithiumSpec(AsymmetricKey::KeyTypes::DILITHIUM3);
+    auto priv_key_ctx = DilithiumAsymmetricKeypair::generate(spec);
+    std::string messageStr = "You better believe I am loaded with both!";
+    signAndVerify(priv_key_ctx, messageStr);
+}
+
+struct BackendTestData
```
garw: Name? Maybe DER TestKeys?

tests/unit/test_dilithium.cpp
```diff
@@ -0,0 +1,1260 @@
...
+{
+};
+
+INSTANTIATE_TEST_SUITE_P(
+        BackendData,
+        DilithiumBackendDataTest,
+        testing::Values(
+                BackendTestData{
+
+                        // DILITHIUM 3
```
garw: Maybe just put large data into files? I think we already have some helpers to load data from files for other of the tests somewhere?

garw (changes requested):

src/mococrw/dilithium.h
wusto: I stuck to the same API as in key.h. Not sure why it is designed like that there.

src/mococrw/dilithium.h
wusto: The idea was to duplicate the interface of `key.h`. Surely it would be possible to add this functionality into `DilithiumKeyImpl`, however the rational was to be able to replace `dilithium.h` with `key.h` once openssl supports dilithium.
But I am open for improvements. :)
Same for the other classes.
I could come up with a wrapping interface for the classes in `key.h` without doing an API break.

src/mococrw/dilithium.h
wusto: In the other contexts the copy and assignment operator are defined.
For having a consistent interface I implemented them here, too.

examples/sig-example.cpp
wusto: For the moment most probably. In the future, when openssl supports dilithium, we can remove the ifdef.
But I am open for both.

src/dilithium.cpp
wusto: yes :)
Tried to improve it.

src/dilithium.cpp
wusto: I wanted to keep the switch/case part as small as possible and tried to avoid code-duplication. But I am also fine with moving code into the switch/case statement.

src/mococrw/dilithium.h
wusto: Implemented the same functions as in asymmetric_crypto_ctx.cpp

src/dilithium.cpp
wusto: clang-format has problems parsing the macros correctly and then indents the source in a strange way. That's the reason why I turn it off for the macro blocks.

dockerfiles/feature-support/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    autoreconf --verbose --install --force && \
+    ./configure --enable-strict && \
+    make -j$(nproc) && \
+    make check && \
+    make install && \
+    rm -rf /tmp/libp11
+
+# Install patched libdilithium
+RUN mkdir /tmp/libdilithium
+ADD dilithium-patches/0001-CMakeLists.txt-Add-BUILD_TESTING-compile-flag.patch   /tmp/libdilithium
```
garw: can we maybe consolidate this into one copy? Adding layers to containers is costly if you add a lot. Maybe hadolint has a good suggestion here as well?

examples/sig-example.cpp
garw: As discussed: Ok as we hope to get rid of the custom things anyways and make dilithium a first-class feature once supported by OpenSSL

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,396 @@
...
+ * Addittionally some convenient functions are added.
+ */
+class DilithiumKeyImpl
+{
+public:
+    enum class DilithiumParameterSet : int { DILITHIUM2, DILITHIUM3, DILITHIUM5 };
+
+    DilithiumKeyImpl(const std::vector<uint8_t> key_data,
+                     DilithiumKeyImpl::DilithiumParameterSet paramSet,
+                     bool is_private_key)
```
garw: if you want to make this nicer readable, you create an enum class KeyType { PRIVATE, PUBLIC }. Then, someone reading the code doesn"t see …, true) but …, KeyType::PRIVATE).

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,396 @@
...
+    /**
+     * @brief Get the selected dilithium parameter set (one of DILITHIUM{2,3,5})
+     *
+     * @return The parameter set chosen
+     */
+    DilithiumParameterSet getDilithiumParameterSet() const { return _paramSet; }
+
+    inline bool operator==(const DilithiumKeyImpl &rhs) const
+    {
+        return _key_data == rhs.getKeyData() && _is_private_key == rhs.isPrivateKey() &&
```
garw: I'm not a 100% sure but I'd imagine that this create a copy of the key data in rhs.getKeyData(). It may feel a bit strange but one can actually access private data objects of `rhs` in this context here because we're in `DilithiumKeyImpl`, so you can actually write `_key_data == rhs._key_data`

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,396 @@
...
+               _paramSet == rhs.getDilithiumParameterSet();
+    }
+
+private:
+    const AsymmetricKey::KeyTypes _keyType = AsymmetricKey::KeyTypes::DILITHIUM;
+    const std::vector<uint8_t> _key_data;
```
garw: Name consistency. This library useses _camelCase for internal variables. There shouldn't be any underscore notation anywhere.

src/mococrw/dilithium.h
garw: Adapt the doxygen to the new situation?

src/mococrw/dilithium.h
garw: hm. The `AsymmetricKey` class is a bit weird because its more or less some plain encapsulation of the OpenSSL things and then just serve the Public and the Keypair classes. The customer code only interacts with the latter two and just uses some of the definitions from `AsymmetricKey`. Maybe we can put the definitions already into AsymmetricKey and otherwise omit this whole class here? I'm not sure how valuable the contribution of another indirection is here.

src/mococrw/dilithium.h
garw: hm. We should be careful with continuing inconsistencies… This is not a good thing to do. Maybe we can eliminate the DilithiumAsymmetricKey here altogether and get rid of this on the way…

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,392 @@
...
+class DilithiumAsymmetricKeypair : public DilithiumAsymmetricPublicKey
+{
+public:
+    DilithiumAsymmetricKeypair(std::shared_ptr<DilithiumKeyImpl> keypair)
+            : DilithiumAsymmetricPublicKey{std::move(keypair)}
+    {
+    }
+
+    /**
+     * @brief Parses an ASN.1 struct in DER format at returns a private key object
```
garw:
```suggestion
     * @brief Parses an ASN.1 struct in DER format and returns a private key object
```

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,392 @@
...
+     * @throws MoCOCrWException if the struct cannot be parsed or if the key is invalid
+     */
+    static DilithiumAsymmetricKeypair readPrivateKeyfromDER(const std::vector<uint8_t> &asn1Data);
+
+    /**
+     * Generate an asymmetric keypair with given Spec.
+     *
+     * @see DilithiumSpec
+     *
+     * @throws This method may throw an OpenSSLException if OpenSSL
```
garw: Does it?  Looks to me it throws MoCoCrWException right now?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,392 @@
...
+ */
+using DilithiumAsymmetricPrivateKey = DilithiumAsymmetricKeypair;
+
+/**
+ * @brief Specification class for Dilithium keys (clone of AsymmetricKey::Spec)
+ *
+ * @see AsymmetricKey::Spec for mor information
+ *
+ */
+class DilithiumAsymmetricKey::Spec
```
garw: Have the `DilithiumSpec` as a standalone class and make it derived from `AsymmetricKey::Spec` in the future? Here, it is just boilerplate and doesn't serve any real purpose?

src/mococrw/dilithium.h
garw: ok

src/dilithium.cpp
garw: ok but shouldn't the clang-format on then be put directly after `IMPLEMENT_ASN1_FUNCTIONS_fname`?

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+#include "mococrw/key.h"
+#include "mococrw/openssl_wrap.h"
+
+namespace mococrw
+{
+using KeyTypes = AsymmetricKey::KeyTypes;
+
+/*
+ * The dilithium private key struct (see RFC5208 and iaik:
+ * http://javadoc.iaik.tugraz.at/iaik_jce/current/iaik/pkcs/pkcs8/PrivateKeyInfo.html)
```
garw: please remove references to iaik. I don't think this is relevant here.

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+                               dilithium_priv)
+
+using DILITHIUM_PRIV_Ptr = std::unique_ptr<DILITHIUM_PRIV,
+                                           openssl::SSLDeleter<DILITHIUM_PRIV, dilithium_priv_free>>;
+using DILITHIUM_PRIV_SharedPtr = utility::SharedPtrTypeFromUniquePtr<DILITHIUM_PRIV_Ptr>;
+// clang-format on
+
+/*
+ * The dilithium public key struct (see RFC5280 and iaik:
+ * http://javadoc.iaik.tugraz.at/iaik_jce/current/iaik/pkcs/pkcs8/PrivateKeyInfo.html)
```
garw: s.a.

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+
+IMPLEMENT_ASN1_FUNCTIONS_fname(DILITHIUM_PUB, DILITHIUM_PUB_INTERNAL, dilithium_pub)
+
+using DILITHIUM_PUB_Ptr = std::unique_ptr<DILITHIUM_PUB,
+                                          openssl::SSLDeleter<DILITHIUM_PUB, dilithium_pub_free>>;
+using DILITHIUM_PUB_SharedPtr = utility::SharedPtrTypeFromUniquePtr<DILITHIUM_PUB_Ptr>;
+// clang-format on
+
+DilithiumKeyImpl::DilithiumParameterSet getKeyTypeFromDilithiumParameterSet(
```
garw: prefix local functions with _?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,392 @@
...
+ * This class is a replacement of EVP_PKEY as long as dilithium is not supported by the openssl
+ * version used by MoCOCrW. It stores the key as std::vector<uint8_t> and the key type.
+ * Addittionally some convenient functions are added.
+ */
+class DilithiumKeyImpl
+{
+public:
+    enum class DilithiumParameterSet : int { DILITHIUM2, DILITHIUM3, DILITHIUM5 };
+
+    DilithiumKeyImpl(const std::vector<uint8_t> key_data,
```
garw: Probably this class wants a destructor that invokes the secure vector delete from our utilities on the key_data.

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,392 @@
...
+ * This shall be replaced by AsymmetricKeypair once Dilithium is supported by the openssl version
+ * used by MoCOCrW.
+ * The interface is similar to the one of AsymmetricKeypair.
+ *
+ * For documentation see AsymmetricKeypair
+ */
+class DilithiumAsymmetricKeypair : public DilithiumAsymmetricPublicKey
+{
+public:
+    DilithiumAsymmetricKeypair(std::shared_ptr<DilithiumKeyImpl> keypair)
```
garw, line 252: Probably we can make this constructor private. This shouldn't be used. Only the reader or generate function should be used to create new key objects. Less potential for future incompatibilities…

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,392 @@
...
+     * @returns The size of the raw key data
+     */
+    uint getKeySize() const { return _key_data.size(); }
+
+    /**
+     * Returns the raw key data
+     *
+     * @returns The raw key data
+     */
+    const std::vector<uint8_t> getKeyData() const { return _key_data; };
```
garw: hm. Now that I thought about this. Maybe it would be better to return a ptr here. This currently creates wild copies of (potentially) private key data in memory that isn't later blanked out. It's not really relevant to our main use-cases but I don't think its good to leave it if we offer this functionality at all. Maybe we just offer a way to get the key data byte buffer directly for the implementation functions?

garw (changes requested):

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+    }
+}
+
+std::shared_ptr<DilithiumKeyImpl> DilithiumKeyImpl::readPublicKeyFromDER(
+        const std::vector<uint8_t> &x509PubKey)
+{
+    const uint8_t *p = x509PubKey.data();
+    auto pubKey = openssl::_d2i_X509_PUBKEY(p, x509PubKey.size());
+
+    const unsigned char *pk = NULL;
```
istepic: Nitpick: nullptr

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+    int ppklen;
+    // Get the nested bit string
+    // we can use get0 here, as pk is not used outside of this function
+    // (d2i_dilithium_pub and the underlying ASN1_item_d2i_ex copies the data according to the
+    // manpage)
+    if (!X509_PUBKEY_get0_param(nullptr, &pk, &ppklen, nullptr, pubKey.get())) {
```
istepic, line 179: Why not use add_function.py for this too?

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+    // Parse the RFC 5958 ASN.1 DER form
+    auto p8inf = openssl::_SSL_d2i_PKCS8_PRIV_KEY_INFO(pkcs8PrivKey.data(), pkcs8PrivKey.size());
+
+    const unsigned char *pk;
+    int ppklen;
+    // get the nested octet string
+    // we can use get0 here, as pk is not used outside of this function
+    // (d2i_dilithium_priv and the underlying ASN1_item_d2i_ex copies the data according to the
+    // manpage)
+    if (!PKCS8_pkey_get0(nullptr, &pk, &ppklen, nullptr, p8inf.get())) {
```
istepic, line 213: Why not use add_function.py for this too?

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+            false);
+}
+
+std::shared_ptr<DilithiumKeyImpl> DilithiumKeyImpl::readPrivateKeyFromDER(
+        const std::vector<uint8_t> &pkcs8PrivKey)
+{
+    // Parse the RFC 5958 ASN.1 DER form
+    auto p8inf = openssl::_SSL_d2i_PKCS8_PRIV_KEY_INFO(pkcs8PrivKey.data(), pkcs8PrivKey.size());
+
+    const unsigned char *pk;
```
istepic: initialize

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+            paramSet,
+            true);
+}
+
+DilithiumKeyImpl DilithiumKeyImpl::getPublicKey() const
+{
+    if (!isPrivateKey()) {
+        return *this;
+    }
+    std::vector<uint8_t> public_key(getPubKeySize(_paramSet));
```
istepic: camelCase

src/mococrw/dilithium.h
wusto: As KeyTypes is already used in the AsymmetricKey class I decided to use AsymmetricType instead. If you have a better word for naming, I am open for improvements.
Alternatives could be:
* PrivacyType
* PrivacyClass
* PrivatenessType

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+}
+
+bool DilithiumKeyImpl::isPrivateKey() const { return _is_private_key; }
+
+std::unique_ptr<DilithiumAsymmetricKey::Spec> DilithiumAsymmetricKey::getKeySpec() const
+{
+    return std::make_unique<DilithiumSpec>(_internal()->getDilithiumParameterSet());
+}
+
+DilithiumAsymmetricPublicKey DilithiumAsymmetricPublicKey::readPublicKeyfromDER(
```
istepic, line 286: I guess we can mark this method as const.

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+        const std::vector<uint8_t> &asn1Data)
+{
+    return DilithiumAsymmetricPublicKey(DilithiumKeyImpl::readPublicKeyFromDER(asn1Data));
+}
+
+DilithiumAsymmetricKeypair DilithiumAsymmetricKeypair::readPrivateKeyfromDER(
```
istepic, line 292: s.a.

src/dilithium.cpp
```diff
@@ -0,0 +1,482 @@
...
+        const std::vector<uint8_t> &asn1Data)
+{
+    return DilithiumAsymmetricKeypair(DilithiumKeyImpl::readPrivateKeyFromDER(asn1Data));
+}
+
+DilithiumAsymmetricKeypair DilithiumAsymmetricKeypair::generate(
```
istepic: s.a.

src/mococrw/dilithium.h
wusto: This requires two friend classes. Did it as extra commit. If you like it, I will fixup the commit, if not, it can be easily removed.

.github/workflows/ci.yml
```diff
@@ -81,8 +87,8 @@ jobs:
         if: ${{ matrix.hsm_flag == 'HSM_ENABLED=ON' }}
```
istepic: Rename this step and change the condition. This is probably why some checks fail now.

.github/workflows/ci.yml
```diff
@@ -45,6 +46,7 @@ jobs:
         compiler: ["g++", "clang++"]
         build_type: ["", "ASAN"]
         hsm_flag: ["HSM_ENABLED=ON", "HSM_ENABLED=OFF"]
+        dilithium_flag: ["DILITHIUM_ENABLED=ON", "DILITHIUM_ENABLED=OFF"]
```
istepic, line 49: We should refactor the CI because this approach means exponential growth of number of checks. Maybe create another job build-and-test-hsm and build-and-test-dilithium. We will still end up with 16 jobs but it will be a bit cleaner.

istepic (changes requested):

examples/sig-example.cpp
```diff
@@ -252,4 +298,30 @@ int main(void)
      */
     edDsaVerify(edPrivKey, signature, message);
     /*********************************************/
+
+#ifdef DILITHIUM_ENABLED
+    /************** dilithium signature **********/
+    /* Please note that the interface might change slightly once openssl officially supports
+     * dilithium
```
ricleite, line 305: that's a.. very important note

it'll be a headache if the interface does change in a non-backwards-compatible way

examples/sig-example.cpp
```diff
@@ -252,4 +298,30 @@ int main(void)
...
+    /* Now the public key, the message and the signature are read from files */
+    auto pubKeyData = utility::bytesFromFile<uint8_t>("dilithium3-public-key.der");
+    auto dilithiumPubKey = DilithiumAsymmetricPublicKey::readPublicKeyfromDER(pubKeyData);
+
+    message = utility::bytesFromFile<uint8_t>("message.raw");
+    signature = utility::bytesFromFile<uint8_t>("dilithium3-test-signature.raw");
+
+    dilithiumVerify(dilithiumPubKey, signature, message);
+    /*********************************************/
+#endif  // DILITHIUM_ENABLED
```
ricleite, line 326: no example with `verifyDigest`?

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,378 @@
...
+     * @param x509PubKey The public key object as defined in RFC5280
+     * @return A shared pointer to a dilithium implementation object
+     */
+    static std::shared_ptr<DilithiumKeyImpl> readPublicKeyFromDER(
+            const std::vector<uint8_t> &x509PubKey);
+
+    /**
+     * @brief Parses an ASN.1 PKCS#8 private key object (see RFC5208)
+     *
+     * @param x509PubKey The private key object as defined in RFC5208
```
ricleite: nitpick: `x509PubKey` != `pkcs8PrivKey`

(I also wonder about the names themselves... but it's not my area of expertise)

src/mococrw/dilithium.h
```diff
@@ -0,0 +1,378 @@
...
+     * @brief Copy Constructor
+     */
+    DilithiumVerificationCtx(const DilithiumVerificationCtx &other);
+
+    /**
+     * @brief Copy Assignment
+     */
+    DilithiumVerificationCtx &operator=(const DilithiumVerificationCtx &other);
+
+    void verifyMessage(const std::vector<uint8_t> &signature, const std::vector<uint8_t> &message);
```
ricleite, line 395: hm, no `verifyDigest`?

I'll guess then that `verifyMessage` does not do any internal `digest` with a known hashing algorithm (e.g., sha256). If this is not true, it could be a problem for applications

src/mococrw/util.h
```diff
@@ -32,6 +33,28 @@ namespace utility
...
+
+    file.seekg(0, std::ios::end);
+    auto size = file.tellg();
+    file.seekg(0, std::ios::beg);
+
+    std::vector<T> buffer;
+    buffer.resize(size);
+    file.read(reinterpret_cast<char *>(buffer.data()), size);
+    return buffer;
+}
```
ricleite: from the error message and content, I'm not sure whether it's a good idea to put this in the public API

it does look like something that ought to be used only internally by mococrw

ricleite (commented):
Light review, it mostly LGTM from an application PoV

examples/sig-example.cpp
wusto: The reference implementation of dilithium does not offer such a function. All the functionality provided expects a message to be passed by reference. Thus verifyDigest is not implemented for dilithium.

src/mococrw/dilithium.h
wusto: thx.
The naming comes from the standards providing the fundamental layout for storing keys.

src/mococrw/dilithium.h
wusto: see above.

src/mococrw/util.h
wusto: hmm. Jep. Not an easy decision.
In the past all the examples had the keys stored in the source file. For dilithium this was omitted as the keys are really big. Instead the keys are stored in files and read from disk. However this requires some function to read data from disk. The only other example reading data from files is the cert test. However there are functions which reads the data directly from disk.
To keep the interface for dilithium the same as for the other asymmetric keys, it was decided to make this function public.

And if this function is used, the namespace mococrw:utility has to be used. So the programmer should expect a crypto related error message. ;-)

examples/sig-example.cpp
wusto: Unfortunately I don't know how the openssl interface will look like. So most probably there will be some minor changes.
We try to keep the interface as stable as possible, so that recompilation when "auto" is used wherever feasible should be possible.

istepic (approved):

src/dilithium.cpp
wusto: then the "using"-lines are indented, too. So I put the line below.

src/dilithium.cpp
wusto: I decided to not to, as the functionality will be removed once openssl supports dilithium.

src/dilithium.cpp
wusto: I decided to not to, as the functionality will be removed once openssl supports dilithium.

src/dilithium.cpp
wusto: I orientated on the functions in key.h for consistency and these function does not have a const qualifier.

src/dilithium.cpp
wusto: I orientated on the functions in key.h for consistency and these function does not have a const qualifier.

.github/workflows/ci.yml
wusto: as discussed offline, I leave it like this to reduce code duplication in the jobs.

src/mococrw/dilithium.h
wusto: as discussed I leave the structure as it is, to simplify the adaptions once openssl supports dilithium.

src/mococrw/dilithium.h
wusto: as discussed I leave the structure as it is (copied from key.h), to simplify the adaptions once openssl supports dilithium.

src/mococrw/dilithium.h
```diff
@@ -40,19 +40,22 @@ class DilithiumKeyImpl
...
+                     DilithiumParameterSet paramSet,
+                     AsymmetricType is_private_key)
+            : _keyData{std::move(keyData)}, _asymType(is_private_key), _paramSet(paramSet)
     {
         if (!hasValidKeySize()) {
             throw MoCOCrWException("Key data does not match the expected size.");
         }
     }
 
+    ~DilithiumKeyImpl() { utility::vectorCleanse<uint8_t>(*_keyData.get()); };
```
garw: The .get() shouldn't be necessary. shared_ptr<T>::operator*() returns T&

garw (commented):
looks ok to me. One more comment that you can feel free to ignore.

Once the clarification about the signMessage function is in, we can merge this from my pov.

garw (approved):
lgtm now

PR #123, opened 2022-12-13
Arrange Pointer Alignment Setting in Clang Format Config

johnfxgalea-e:
Arranges Pointer alignment setting in clang-format such that alignment is done on the right. Also, removes DerivePointerAlignment so that formatting is deterministic.

Code-base is formatted accordingly.


garw (approved):
lg2m

miroshko (approved):

PR #130, opened 2023-01-17
hsm: add ecies to hsm-example

istepic:
Heavily inspired by ecies-example.cpp


wusto (approved):
lgtm.

garw (commented):
Not sure I understand the purpose of this. We have a demo how to load and use keys from an HSM and we have a demo that does ECIES based on AssymmetricKey objects. There is nothing special about ECIES+HSM. What shall this illustrate here?

PR #131, opened 2023-01-23
hsm: move unrelated code from hsm-example to separate integration test

istepic:
To keep hsm-example clean and easy to read, using only HSM API is demonstrated. Other things that were meant for integration testing are moved to a seperate file.

### DIFF
--- a/examples/hsm-example.cpp
+++ b/examples/hsm-example.cpp
@@ -16,142 +16,14 @@
  * limitations under the License.
  * #L%
  */
-#include <mococrw/asymmetric_crypto_ctx.h>
-#include <mococrw/hash.h>
 #include <mococrw/hsm.h>
 #include <mococrw/key.h>
-#include <iostream>
 
 /* This example demonstrates how to create a PKCS11 engine object,
- * and store a key inside the HSM.
+ * and store/load a key inside the HSM.
  */
 
 using namespace mococrw;
-
-std::vector<uint8_t> rsaSign(const AsymmetricPrivateKey &privKey,
-                             const DigestTypes digestType,
-                             std::shared_ptr<RSASignaturePadding> padding,
-                             const std::vector<uint8_t> &message)
-{
-    std::shared_ptr<MessageSignatureCtx> signCtx;
-
-    try {
-        /* Padding is optional. Default: PSSPadding with MGF1 as mask generation function.
-         * DigestType is used as hash function for the padding schemes and for MGF1 (if used) */
-        signCtx = std::make_shared<RSASignaturePrivateKeyCtx>(privKey, digestType, padding);
-    } catch (const MoCOCrWException &e) {
-        std::cerr << "Please check your RSA key. Failure creating context." << std::endl;
-        std::cerr << e.what();
-        exit(EXIT_FAILURE);
-    }
-
-    std::vector<uint8_t> signature;
-    try {
-        signature = signCtx->signMessage(message);
-    } catch (const MoCOCrWException &e) {
-        /* Possible reasons:
-         * - error in openssl (sign, padding, ...)
-         * - Hash function's digest size doesn't match the message's digest size
-         */
-        std::cerr << "Failure occurred during signing." << std::endl;
-        std::cerr << e.what() << std::endl;
-        exit(EXIT_FAILURE);
-    }
-    return signature;
-}
-
-void rsaVerify(const AsymmetricPublicKey &pubKey,
-               std::shared_ptr<RSASignaturePadding> rsaPadding,
-               DigestTypes digestType,
-               const std::vector<uint8_t> &signature,
-               const std::vector<uint8_t> &message)
-{
-    std::shared_ptr<MessageVerificationCtx> verifyCtx;
-    try {
-        /* Padding is optional. Default: PSSPadding with MGF1 as mask generation function.
-         * DigestType is used as hash function for the padding schemes and for MGF1 (if used) */
-        verifyCtx = std::make_shared<RSASignaturePublicKeyCtx>(pubKey, digestType, rsaPadding);
-    } catch (const MoCOCrWException &e) {
-        std::cerr << "Please check your RSA key. Failure creating context." << std::endl;
-        std::cerr << e.what();
-        exit(EXIT_FAILURE);
-    }
-
-    try {
-        verifyCtx->verifyMessage(signature, message);
-    } catch (const MoCOCrWException &e) {
-        /* Possible reasons:
-         * - error in openssl (sign, padding, ...)
-         * - Hash function's digest size doesn't match the message's digest size
-         * - Invalid signature
-         */
-        std::cerr << "Verification failed!" << std::endl;
-        std::cerr << e.what() << std::endl;
-        exit(EXIT_FAILURE);
-    }
-}
-
-std::vector<uint8_t> ecdsaSign(const AsymmetricPrivateKey &privKey,
-                               const DigestTypes digestType,
-                               const ECDSASignatureFormat sigFormat,
-                               const std::vector<uint8_t> &message)
-{
-    std::shared_ptr<MessageSignatureCtx> signCtx;
-    try {
-        /* The argument hashFunction is optional. Default is SHA256
-         * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
-        signCtx = std::make_shared<ECDSASignaturePrivateKeyCtx>(privKey, digestType, sigFormat);
-    } catch (MoCOCrWException &e) {
-        std::cerr << "Please check your ECC key. Failure creating context." << std::endl;
-        std::cerr << e.what();
-        exit(EXIT_FAILURE);
-    }
-
-    std::vector<uint8_t> signature;
-    try {
-        signature = signCtx->signMessage(message);
-    } catch (const MoCOCrWException &e) {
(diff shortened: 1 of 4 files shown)


examples/hsm-example.cpp
```diff
@@ -162,68 +34,29 @@ int main(void)
...
     mococrw::RSASpec rsaSpec;
-    auto rsaSignatureDigestType = DigestTypes::SHA512;
-    auto mgf1 = std::make_shared<MGF1>(mgf1DigestType);
-    int saltLength = mococrw::Hash::getDigestSize(rsaSignatureDigestType);
-    auto padding = std::make_shared<PSSPadding>(mgf1, saltLength);
-
-    /**
-     * Generate an RSA keypair and load the public part
-     */
     auto rsaPrivKey =
```
garw, line 53: is it possible to call `publicKeyToPem()`? I would imagine so. This is in most of the cases probably more straight-forward than calling the additional readPublicKeyFromHSM?

The example should probably also show how to load a previously created *private* key later for doing crypto with it.

tests/integration/hsm-integration-test.cpp
```diff
@@ -0,0 +1,452 @@
...
+int main(void)
+{
+    // Information for engine loading and key management.
+    std::string id("pkcs11");
+    std::string modulePath("/usr/lib/softhsm/libsofthsm2.so");
+    std::string tokenLabel("token-label");
+    // Don't hardcode the pin in your application, this is just for demonstration purposes
+    std::string pin("1234");
+    HsmEngine hsmEngine(id, modulePath, tokenLabel, pin);
+    std::vector<uint8_t> message = utility::fromHex("deadbeef");
```
garw, line 276: I guess we would like to have some intermediate output what is tested next to have a better impression what actually failed when it fails.

tests/integration/hsm-integration-test.cpp
```diff
@@ -0,0 +1,452 @@
...
+    ECCSpec ecspec;
+    auto eccPrivKey =
+            AsymmetricPrivateKey::generateKeyOnHSM(hsmEngine, ecspec, keyLabelECC, keyIDECC);
+    auto ecdsaSigFormat = ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS;
+
+    /* The argument hashFunction is optional. Default is SHA256
+     * The default signature format is ECDSASignatureFormat::ASN1_SEQUENCE_OF_INTS */
+    auto ECDSAsignature = ecdsaSign(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, message);
+
+    /* Use-Case 1: You want to check your own signature.
```
garw: Does this trigger verfication in OpenSSL SW or via PKCS#11 C_Verify? Please check and add to the comment what we expect is tested here.

tests/integration/hsm-integration-test.cpp
```diff
@@ -0,0 +1,452 @@
...
+     * We can use here the private key, as it also contains the public key.
+     * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
+     * we do an implicit upcast here.
+     */
+    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, ECDSAsignature, message);
+
+    /* Use-Case 2: The public key used for verification is stored in the HSM */
```
garw: Isn't this effectively the same as UC 1? If not, stress in the comment what happens differently in the background.

tests/integration/hsm-integration-test.cpp
```diff
@@ -0,0 +1,452 @@
...
+    auto pubKeyEcc = AsymmetricPublicKey::readPublicKeyFromHSM(hsmEngine, keyLabelECC, keyIDECC);
+    ecdsaVerify(pubKeyEcc, ecdsaDigestType, ecdsaSigFormat, ECDSAsignature, message);
+
+    /* Use-Case 3: You want to write the public key to a PEM file and use it later
+     * for verification
+     */
+    auto pubKeyPem = eccPrivKey.publicKeyToPem();
```
garw, line 312: This is now definitely a SW verification. So we can consider this as a test that checks if Crypto Inside/outside of the HSM plays nicely together.

tests/integration/hsm-integration-test.cpp
```diff
@@ -0,0 +1,452 @@
...
+    ecdsaVerify(pubKeyEccFromPem, ecdsaDigestType, ecdsaSigFormat, ECDSAsignature, message);
+    /*********************************************/
+
+    /************** RSA key generation, loading and digital signatures **************/
+    std::vector<uint8_t> keyIDRSA{0x12, 0x34};
+    std::string keyLabelRSA{"rsa-key-label"};
+    auto mgf1DigestType = DigestTypes::SHA256;
+    mococrw::RSASpec rsaSpec;
+    auto rsaSignatureDigestType = DigestTypes::SHA512;
+    auto mgf1 = std::make_shared<MGF1>(mgf1DigestType);
```
garw: Maybe for the purpose of this test, we should get rid of all these details and just use the defaults.

tests/integration/hsm-integration-test.cpp
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    /* The underlying openssl function recognizes the used serialization format of the ephemeral key
+     * reading the first byte of the serialized data, which defines the format.
+     * The variable eccSpec contains the used elliptic curve of the ephemeral key */
+    auto _ephemeralKey = AsymmetricPublicKey::fromECPoint(eccSpec, eciesData.ephemeralKey);
+
+    /* Get the decryption context */
+    auto decCtx =
+            ECIESCtxBuilder{}  // This is optional. The default is X963 with SHA512
```
garw: Maybe we should not specify all the details in the context of this test here and just use the defaults where possible. Here, the focus is on the Mococrw <> HSM interplay and not at showcasing the configurability of ECIES.

garw (changes requested):
I think this is a good overall approach. A few comments that imho would help to simplify this here a bit and put more focus on what this is actually about.

examples/hsm-example.cpp
istepic: Not sure why publicKeyToPem makes sense in this context.

Loading private key is now demonstrated.

tests/integration/hsm-integration-test.cpp
istepic: Here's an example output now:

```
root@a54673e28eb0:/ws/build/tests/integration# ./hsm-integration-test
1. Testing digital signatures using ECC keys on HSM:
Generating ECC keys on HSM...Success
Signing a message using an ECC private key generated on HSM...Success
Verifying the message with the corresponding public key...Success
Explicitly loading a public key from HSM and trying to verify a signature...Success
Transforming a public key from HSM to a PKCS8 format that can be written to a PEM file...Success
Constructing a public key from PKCS8 format...Success
Doing the verification with public key from PKCS8 format (this verification is done in software)...Success

2. Testing digital signatures using RSA keys on HSM:
Generating ECC keys on HSM...Success
Explicitly loading a public key from HSM and trying to verify a signature...Success
Signing a message using an RSA private key generated on HSM...Success
Verifying the message with the corresponding public key...Success

3. Testing ECIES scheme using ECC keys on HSM:
Encrypting the message...Success
Decrypting the message...Success
```

tests/integration/hsm-integration-test.cpp
istepic: Agreed.

tests/integration/hsm-integration-test.cpp
istepic: Done.

tests/integration/hsm-integration-test.cpp
istepic: I did check this with SoftHSM compiled with debug symbols a while back. We can't really test that here tho. Comment added.

tests/integration/hsm-integration-test.cpp
istepic: It is but the key is explicitly loaded by loadPublicKey function.

tests/integration/hsm-integration-test.cpp
istepic: Done

tests/integration/hsm-integration-test.cpp
garw: Looks unchanged to me?

garw (approved):

examples/hsm-example.cpp
```diff
@@ -162,68 +34,32 @@ int main(void)
...
-
-    /* Use-Case 1: You want to check your own signature.
-     * We can use here the private key, as it also contains the public key.
-     * In MoCOCrW the AsymmetricPrivateKey is a specialisation of AsymmetricPublicKey. Thus
-     * we do an implicit upcast here.
-     */
-    ecdsaVerify(eccPrivKey, ecdsaDigestType, ecdsaSigFormat, ECDSAsignature, message);
-
-    /* Use-Case 2: The public key used for verification is stored in the HSM */
+    // Just the public key can be loaded explicitly
```
johnfxgalea-e: Nit: Remove `Just` because then you mention later on that you can load a handle of the private key.

johnfxgalea-e (commented):
LGTM

johnfxgalea-e (approved):

PR #132, opened 2023-02-08
CMAKE: Add CMake Header Configuration

johnfxgalea-e:
Configures a header to define MACROs for optional features provided by MoCOCrW. As a result, it avoids the need for users to add target_compile_definitions commands to enable the features.


examples/sig-example.cpp
```diff
@@ -17,6 +17,7 @@
  * #L%
  */
 
+#include <MoCOCrWConfig.h>
```
garw: This should be handled internally in the mococrw headers.

src/CMakeLists.txt
```diff
@@ -52,7 +52,11 @@ if(DILITHIUM_ENABLED)
     )
 endif()
 
+# Set up configuration header for optional features.
+configure_file(${MOCOCRW_PROJECT_ROOT}/cmake/in/MoCOCrWConfig.h.in ${CMAKE_BINARY_DIR}/MoCOCrWConfig.h)
```
garw: We should put this into a special dir that follows the usual scheme, so mococrw/Config.h.

src/mococrw/key.h
```diff
@@ -17,6 +17,7 @@
  * #L%
  */
 #pragma once
+#include <MoCOCrWConfig.h>
```
garw: Should be (as the other ones) mococrw/

src/CMakeLists.txt
```diff
@@ -98,16 +102,16 @@ add_library(${LIBRARY_NAME} SHARED ${LIBRARY_SOURCES} ${LIBRARY_PUBLIC_HEADERS})
...
 if(DILITHIUM_ENABLED)
     target_link_libraries(${LIBRARY_NAME} PRIVATE dilithium2_ref dilithium3_ref dilithium5_ref aes256ctr_ref fips202_ref)
 endif()
 
+#Include BINARY_DIR so that sources can find the config header.
+target_include_directories(${LIBRARY_NAME} PRIVATE ${CMAKE_BINARY_DIR})
+
 target_include_directories(${LIBRARY_NAME} INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
```
garw: Doesn't this just repeat the definition of l. 110?
It should work without it. Otherwise, it will not work with the generated exported target.

garw (changes requested):
I feel like this is still a bit problematic when installed. See the detailed comments.

CMakeLists.txt
```diff
@@ -124,12 +124,14 @@ set(MOCOCRW_INSTALL_BINDIR ${CMAKE_INSTALL_FULL_BINDIR})
...
 
 set(MOCOCRW_PROJECT_ROOT ${CMAKE_CURRENT_SOURCE_DIR})
+set(MOCOCRW_INCLUDE_CONFIG_DIR ${CMAKE_BINARY_DIR}/config_include)
 set(SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/src)
 #TODO set(DOC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/doc)
 set(TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR}/tests)
 
 # Enable all subprojects to find each other's include files
 include_directories(${SRC_DIR})
+include_directories(${MOCOCRW_INCLUDE_CONFIG_DIR})
```
garw, line 134: We usually strongly discourage the use of `include_directories` as this is global and easily shadows other problems. We do have it here for the rest already. So I guess this is (in principle) ok.

garw (commented):
The plan on your end is to include in the config in error.h because this is used everywhere already? 
Please write a bit more verbose commit message that talks about your intention.

A good commit message describes the problem, the solution approach + rationale for the main design.

CMakeLists.txt
johnfxgalea-e: Indeed, I don't like it either but figured this is how it has been done previously.

garw (approved):

miroshko (approved):

PR #136, opened 2023-02-20
HSM: Raise MoCOCrW Exception for unknown key ID

johnfxgalea-e:
When attempting to load a key with an unknown ID,
MoCOCrW raises an OpenSSLException. This is not
ideal, as such exceptions are usually raised for
internal problems within OpenSSL. For cases where
an ID is unknown, one would expect a MoCOCrWException
to be thrown back to the user.

To make matters worse, the current catch-all
approach of using OpenSSLException makes it
difficult to distinguish different types of
errors.

In order to specifically identify the error where the passed key is unknown,
we check that the error stems from the pkcs11 engine and that the reason is
"object not found". To obtain this information,
SSL_ERR_lib_error_string() and SSL_ERR_reason_error_string() wrappings
are introduced.

Inside HsmEngine::loadPrivateKey() and
HsmEngine::loadPublicKey(), the OpenSSLException
is caught and checked whether it has been raised
due to unknown ID. If so, a MoCOCrWException is
thrown. Otherwise, the original OpenSSLException
is thrown again.

### DIFF
--- a/src/hsm.cpp
+++ b/src/hsm.cpp
@@ -69,7 +69,19 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPublicKey(const std::string &keyLabel,
     auto keyIDPctEncoded = pctEncode(keyID);
     std::string pkcs11URI =
             "pkcs11:token=" + _tokenLabel + ";object=" + keyLabel + ";id=" + keyIDPctEncoded;
-    return _ENGINE_load_public_key(_engine.get(), pkcs11URI);
+    try {
+        return _ENGINE_load_public_key(_engine.get(), pkcs11URI);
+    } catch (const OpenSSLException &e) {
+        // The current OpenSSLException catch-all approach makes it difficult to distinguish
+        // different types of errors. In order to specifically identify the case where the passed
+        // key is unknown, we check that the error stems from the pkcs11 engine and that the
+        // reason is "object not found".
+        if (e.getLib() == "pkcs11 engine" && e.getReason() == "object not found") {
+            throw MoCOCrWException("Unable to load public key. Public key not found!");
+        }
+        // If not Unknown Key error, then throw again the original exception.
+        throw;
+    }
 }
 
 openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyLabel,
@@ -78,7 +90,14 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPrivateKey(const std::string &keyLabel,
     auto keyIDPctEncoded = pctEncode(keyID);
     std::string pkcs11URI =
             "pkcs11:token=" + _tokenLabel + ";object=" + keyLabel + ";id=" + keyIDPctEncoded;
-    return _ENGINE_load_private_key(_engine.get(), pkcs11URI);
+    try {
+        return _ENGINE_load_private_key(_engine.get(), pkcs11URI);
+    } catch (const OpenSSLException &e) {
+        if (e.getLib() == "pkcs11 engine" && e.getReason() == "object not found") {
+            throw MoCOCrWException("Unable to load private key. Private key not found!");
+        }
+        throw;
+    }
 }
 
 openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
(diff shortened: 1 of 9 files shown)


src/hsm.cpp
```diff
@@ -69,7 +69,21 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPublicKey(const std::string &keyLabel,
...
+    } catch (const OpenSSLException &e) {
+        // The current OpenSSLException catch-all approach makes it difficult to distinguish
+        // different types of errors. In order to specifically identify the case where the passed
+        // key is unknown, we check that the error stems from the pkcs11 engine and that the
+        // reason is "object not found".
+        if (e.getLib() == "pkcs11 engine" && e.getReason() == "object not found") {
+            throw MoCOCrWException("Unable to load public key. Public key not found!");
+        }
+        // If not Unknown Key error, then throw again the original exception.
+        throw e;
```
garw: just use throw without parameter to re-throw the current exception.

src/hsm.cpp
```diff
@@ -69,7 +69,21 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::loadPublicKey(const std::string &keyLabel,
     auto keyIDPctEncoded = pctEncode(keyID);
     std::string pkcs11URI =
             "pkcs11:token=" + _tokenLabel + ";object=" + keyLabel + ";id=" + keyIDPctEncoded;
-    return _ENGINE_load_public_key(_engine.get(), pkcs11URI);
+    openssl::SSL_EVP_PKEY_Ptr key;
+    try {
+        key = _ENGINE_load_public_key(_engine.get(), pkcs11URI);
```
garw: you can still just return here

src/mococrw/openssl_wrap.h
```diff
@@ -119,7 +119,8 @@ class OpenSSLException final : public std::exception
 {
 public:
     template <class StringType>
-    explicit OpenSSLException(StringType &&message) : _message{std::forward<StringType>(message)}
+    explicit OpenSSLException(StringType &&message)
+            : _message{std::forward<StringType>(message)}, _library{""}, _reason{""}
```
garw: std::string's don't need to be initialized with the empty string. Thats the default constructor.

src/mococrw/openssl_wrap.h
```diff
@@ -128,13 +129,22 @@ class OpenSSLException final : public std::exception
...
-    OpenSSLException() : _message{generateOpenSSLErrorString()} {}
+    OpenSSLException();
 
     const char *what() const noexcept override { return _message.c_str(); }
 
+    /**
+     * The following functions enable fine-tuned error handling via sub-library and reason
+     * information
+     */
+    const std::string &getLib() const { return _library; };
```
garw: Normally, we just return strings and not references. Returning references is potentially dangerous (even const ones). The overhead usually doesn't justify the memory management problems.

tests/unit/test_hsm.cpp
```diff
@@ -131,3 +137,23 @@ TEST_F(HSMTest, testHSMKeygen)
             .WillOnce(Return(pkey));
     EXPECT_NO_THROW(AsymmetricKeypair::generateKeyOnHSM(*hsm, eccSpec, keyLabel, keyId));
 }
+
+TEST_F(HSMTest, testHSMLoadUnknownPublicKey)
```
garw, line 141: Also test the private key loader?

garw (changes requested):
I kinda fear that the approach is a bit fragile but I guess it is the best we can do. So I think this is a reasonable way to approach the problem.
Just a few minor comments.

tests/unit/test_hsm.cpp
johnfxgalea-e: Done.

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/mococrw/openssl_wrap.h
johnfxgalea-e: Done.

src/hsm.cpp
johnfxgalea-e: Done.

src/hsm.cpp
johnfxgalea-e: Done.

garw (approved):

istepic (approved):

PR #137, opened 2023-02-20
HSM: Detect duplicate key generation

johnfxgalea-e:
The PKCS#11 standard is unclear on what happens when
a key is already stored in an HSM and an additional generate
operation is invoked to create a new key with the same label
and ID.

Consequently, this change adds a check to MoCOCrW in order to
detect duplicate key generation and raise an exception in such cases.

This is achived by attempting to load the key prior to invoking the
KEYGEN command. If the key is unknown at this point (determined by trying
to catch a MoCOCrWUnknownHsmKeyException), key generation is allowed to
be performed. Otherwise, a MoCOCrWHsmKeyDuplicateGenerationException
is raised.

This change also includes a patch to libp11 which adds a new ENGINE
command to open a RW session, a must for key generation in order to
avoid read-only errors. Prior to the patch, key generation
could fail if a read-only session was internally used for an earlier
operation.

### DIFF
--- a/dockerfiles/feature-support/hsm-patches/0002-Add-RW-Session-Engine-Command.patch
+++ b/dockerfiles/feature-support/hsm-patches/0002-Add-RW-Session-Engine-Command.patch
@@ -1,0 +1,154 @@
+From 4ef4c70d30dff8346a6291cfc840538e84c04459 Mon Sep 17 00:00:00 2001
+From: John Galea <John.Galea@bmw.de>
+Date: Mon, 27 Feb 2023 17:32:17 +0100
+Subject: [PATCH] Add RW Session Engine Command
+
+In order to support key generation, we add a new command that
+opens a RW session. Prior to this change, it was possible to
+make key generation fail due to the internal use of a read-only
+session.
+---
+ src/eng_back.c  | 71 ++++++++++++++++++++++++++++++++++++++++---------
+ src/eng_err.h   |  1 +
+ src/eng_front.c |  4 +++
+ src/engine.h    |  1 +
+ 4 files changed, 65 insertions(+), 12 deletions(-)
+
+diff --git a/src/eng_back.c b/src/eng_back.c
+index 868463e..809865d 100644
+--- a/src/eng_back.c
++++ b/src/eng_back.c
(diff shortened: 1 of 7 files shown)


src/mococrw/hsm.h
```diff
@@ -124,6 +124,9 @@ class HsmEngine : public HSM
     openssl::SSL_EVP_PKEY_Ptr generateKey(const ECCSpec &spec,
                                           const std::string &keyLabel,
                                           const std::vector<uint8_t> &keyID) override;
+
+    bool isUnknownKeyError(const openssl::OpenSSLException &e) const;
```
garw, line 128: Shouldn't those be private?

src/hsm.cpp
```diff
@@ -104,6 +130,12 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
                                                  const std::string &keyLabel,
                                                  const std::vector<uint8_t> &keyID)
 {
+    // Check for duplicate key generation:
+    checkForDuplicateKey(keyLabel, keyID);
+    checkForDuplicateKey("", keyID);
```
garw, line 135: What does this do? Does it search for the empty label or does it not include the label at all?

dockerfiles/feature-support/hsm-patches/0002-Add-RW-Session-Engine-Command.patch
```diff
@@ -0,0 +1,154 @@
...
+ 
+diff --git a/src/engine.h b/src/engine.h
+index 740f86e..19bf8dd 100644
+--- a/src/engine.h
++++ b/src/engine.h
+@@ -53,6 +53,7 @@
+ #define CMD_FORCE_LOGIN	(ENGINE_CMD_BASE+9)
+ #define CMD_RE_ENUMERATE	(ENGINE_CMD_BASE+10)
+ #define CMD_KEYGEN		(ENGINE_CMD_BASE+11)
++#define CMD_RW			(ENGINE_CMD_BASE+12)
```
garw, line 148: I'm not really happy with this approach. Why do we need to explictly call this?

garw (changes requested):
Let's have a quick chat

PR #138, opened 2023-02-22
hsm: cleanse HSM token PIN after usage

istepic:


src/mococrw/hsm.h
```diff
@@ -109,7 +109,7 @@ class HsmEngine : public HSM
     /** Token label used to uniquely identify a token on which objects reside */
     const std::string _tokenLabel;
     /** Pin to access PKCS11 Engine. */
-    const std::string _pin;
+    std::string _pin;
```
johnfxgalea-e: I am wondering why we keep pin in memory and not just cleanse & forget it after hsm construction?

johnfxgalea-e: Why not remove field in class? _ENGINE_ctrl_cmd_string(_engine.get(), "PIN", _pin); is called in constructor.

johnfxgalea-e (approved):

src/mococrw/hsm.h
```diff
@@ -86,8 +86,8 @@ class HsmEngine : public HSM
 public:
     /**
      * @brief Constructor for an object that can manage keys on HSM using OpenSSL Engine
-     * @note Each HsmEngine object is associated with a specific token and a pin to login to that
-     * token
+     * @note Each HsmEngine object is associated with a specific token
+     * @warning PIN value is not safely cleaned from memory. Make sure you clean it.
```
johnfxgalea-e: How do they clean it - please help the user? Why did we remove the cleanse wrapped function? Couldn't the users use that for convenience?

johnfxgalea-e (changes requested):

src/mococrw/hsm.h

johnfxgalea-e (approved):
Looks good. Just don't forget to update the changelog

vkrikun (commented):
lgtm, but example shall be updates (see comment below).

If we have to deal with PINs, passwords and other sensitive data we should consider creating a "Secret" wrapper class instead of using plain std:. Then we can erase data in destructor.

vkrikun (changes requested):
Pls. update HSM example to erase PIN from memory

istepic:
> If we have to deal with PINs, passwords and other sensitive data we should consider creating a "Secret" wrapper class instead of using plain std:. Then we can erase data in destructor.

Good point. This would break the API and we will have to move to provider interface soon anyways so I recommend we do it like this for now.

vkrikun (approved):

src/mococrw/hsm.h
```diff
@@ -88,6 +88,8 @@ class HsmEngine : public HSM
      * @brief Constructor for an object that can manage keys on HSM using OpenSSL Engine
      * @note Each HsmEngine object is associated with a specific token and a pin to login to that
      * token
+     * @warning PIN value is not safely cleaned from memory. Make sure you clean it. Check
```
wusto, line 91:
```suggestion
@warning PIN value is not automatically cleaned from memory in a safe manner after HSM is unlocked. Make sure you clean it after unlocking. Check
```

src/mococrw/util.h
```diff
@@ -78,6 +78,11 @@ void vectorCleanse(std::vector<T> &vec)
     }
```
wusto, line 78: Can you please change the line above to also use `openssl::_OPENSSL_cleanse`?

PR #139, opened 2023-02-28
utils: Add unit test for cleanse utilities

istepic:
std::vector::clear() was added to vector cleanse to clear the vector meta data with may be sensitive

### DIFF
--- a/tests/unit/test_util.cpp
+++ b/tests/unit/test_util.cpp
@@ -78,3 +78,41 @@ TEST_F(UtilTest, testFromHex)
 
     ASSERT_THROW(result = mococrw::utility::fromHex("  deadbeef"), mococrw::MoCOCrWException);
 }
+
+TEST_F(UtilTest, stringCleanse)
+{
+    std::string secret{"1234"};
+    mococrw::utility::stringCleanse(secret);
+    ASSERT_EQ(0, secret.size());
+}
+
+TEST_F(UtilTest, stringCleanseCheckMemory)
+{
+    std::string secret{"1234"};
+    auto secretPtr = &secret[0];
+    auto secretSize = secret.size();
+    std::string zeroedOutString(secretSize, '\0');
+    mococrw::utility::stringCleanse(secret);
+    // Capacity of std::string may change, depending on implementation.
+    // Only test this if the memory is still allocated.
+    if (secret.capacity() >= secretSize) {
+        ASSERT_EQ(0, memcmp(secretPtr, &zeroedOutString[0], secretSize));
+    }
+}
+
+TEST_F(UtilTest, vectorCleanse)
+{
+    std::vector<uint8_t> secret = {222, 173, 190, 239};
+    mococrw::utility::vectorCleanse(secret);
+    ASSERT_EQ(0, secret.size());
+}
+
+TEST_F(UtilTest, vectorCleanseCheckMemory)
+{
+    std::vector<uint8_t> secret = {222, 173, 190, 239};
+    auto secretPtr = &secret[0];
+    auto secretSize = secret.size();
+    std::vector<uint8_t> zeroedOutVector(secretSize, 0);
+    mococrw::utility::vectorCleanse(secret);
+    ASSERT_EQ(0, memcmp(secretPtr, &zeroedOutVector[0], secretSize));
+}
--- a/src/mococrw/util.h
+++ b/src/mococrw/util.h
@@ -75,6 +75,8 @@ void vectorCleanse(std::vector<T> &vec)
     // std::vector<T>::data() is allowed to return a nullptr if the size is 0.
     if (vec.size() > 0) {
         OPENSSL_cleanse(vec.data(), vec.size() * sizeof(T));
+        // Clear all vector metadata
+        vec.clear();
     }
 }
 


tests/unit/test_util.cpp
```diff
@@ -78,3 +78,17 @@ TEST_F(UtilTest, testFromHex)
...
+    std::string secret{"1234"};
+    mococrw::utility::stringCleanse(secret);
+    ASSERT_THAT(secret, Eq(std::string()));
+}
+
+TEST_F(UtilTest, vectorCleanse)
+{
+    std::vector<uint8_t> secret = {222, 173, 190, 239};
+    mococrw::utility::vectorCleanse(secret);
+    ASSERT_THAT(secret, Eq(std::vector<uint8_t>()));
```
vkrikun: Since we do not change capacity of the vector inside `vectorCleanse()` you can assert of data was actually overwritten with 0-s. To be on a safe side, you can check object `.capacity()` before  checking for zeros.

BTW: I have noticed that we `.clear()` strings but don't `.clear()` vectors.  This is rather inconsistent.

vkrikun (changes requested):

tests/unit/test_util.cpp
```diff
@@ -78,3 +78,35 @@ TEST_F(UtilTest, testFromHex)
...
+}
+
+TEST_F(UtilTest, vectorCleanseCheckMemory)
+{
+    std::vector<uint8_t> secret = {222, 173, 190, 239};
+    auto secretSize = secret.size();
+    std::vector<uint8_t> zeroedOutVector(secretSize, 0);
+    mococrw::utility::vectorCleanse(secret);
+    ASSERT_THAT(0, memcmp(&secret[0], &zeroedOutVector[0], secretSize));
```
vkrikun: Using `.capacity()` to check boundaries will be safer
```
    std::vector<uint8_t> secret = {222, 173, 190, 239};
    mococrw::utility::vectorCleanse(secret);
    std::vector<uint8_t> zeroedOutVector(secret.capacity(), 0);
    ASSERT_THAT(0, memcmp(&secret[0], &zeroedOutVector[0], secret.capacity()));
```

vkrikun (changes requested):

vkrikun (approved):

tests/unit/test_util.cpp
```diff
@@ -78,3 +78,37 @@ TEST_F(UtilTest, testFromHex)
...
+}
+
+TEST_F(UtilTest, stringCleanseCheckMemory)
+{
+    std::string secret{"1234"};
+    auto secretSize = secret.size();
+    std::string zeroedOutString(secretSize, '\0');
+    mococrw::utility::stringCleanse(secret);
+    ASSERT_GE(secret.capacity(), secretSize);
+    ASSERT_THAT(0, memcmp(&secret[0], &zeroedOutString[0], secretSize));
```
garw: This test relies on undefined behavior.
The reference actually says that 
> If pos > size(), the behavior is undefined.

It does not refer to capacity. So while I guess this mostly works, therei s no strong guarantee for this. If you want to test if the memory was nulled, it needs to happen before the size is reset. 
I also don't think that an implementation needs to keep the space around. reset() may free the buffer.
So, while I appreciate the effort here, I don't think that this is the best way to test this.
If we really want to test for this, we should get a pointer to the heap fragment that contains the string before cleaning and then checking lateron that nothing from the original data is left there.

tests/unit/test_util.cpp
```diff
@@ -78,3 +78,37 @@ TEST_F(UtilTest, testFromHex)
...
+    ASSERT_THAT(secret, Eq(std::string()));
+}
+
+TEST_F(UtilTest, stringCleanseCheckMemory)
+{
+    std::string secret{"1234"};
+    auto secretSize = secret.size();
+    std::string zeroedOutString(secretSize, '\0');
+    mococrw::utility::stringCleanse(secret);
+    ASSERT_GE(secret.capacity(), secretSize);
```
garw: This tests something that isn't actually guaranteed.
The reference says

> Unlike for [std::vector::clear](https://en.cppreference.com/w/cpp/container/vector/clear), the C++ standard does not explicitly require that [capacity](https://en.cppreference.com/w/cpp/string/basic_string/capacity) is unchanged by this function, but existing implementations do not change capacity. This means that they do not release the allocated memory (see also [shrink_to_fit](https://en.cppreference.com/w/cpp/string/basic_string/shrink_to_fit)).

tests/unit/test_util.cpp
```diff
@@ -78,3 +78,37 @@ TEST_F(UtilTest, testFromHex)
...
+    ASSERT_THAT(0, memcmp(&secret[0], &zeroedOutString[0], secretSize));
+}
+
+TEST_F(UtilTest, vectorCleanse)
+{
+    std::vector<uint8_t> secret = {222, 173, 190, 239};
+    mococrw::utility::vectorCleanse(secret);
+    ASSERT_THAT(secret, Eq(std::vector<uint8_t>()));
```
garw: This looks like a really complex way to test if secret.size() == 0 to me?

garw (changes requested):

tests/unit/test_util.cpp
istepic: Done

tests/unit/test_util.cpp
istepic: Done

tests/unit/test_util.cpp
istepic: Done

garw (approved):

PR #140, opened 2023-03-07
hsm: don't allow creating keys with the same ID

istepic:
PKCS#11 standard does not enforce unique ids but we want to because
having duplicate ids causes problems on all sides. Before key
generation, key is loaded to check if it already exists. Loading a key
creates a RO session on the token. This represented a problem for key
generation because RW session is needed and libp11 reused previously
created RO session. Key generation patch was slightly modified so that
we always open a RW session before generating a key.

### DIFF
--- a/dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
+++ b/dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
@@ -1,4 +1,4 @@
-From e39fd1f3831294c6b1d3d636e4563bff86b9cf8c Mon Sep 17 00:00:00 2001
+From 828282b17f40237bb875e93df0e216e12c4f2504 Mon Sep 17 00:00:00 2001
 From: istepic <ivan.stepich@gmail.com>
 Date: Mon, 5 Dec 2022 22:44:25 +0100
 Subject: [PATCH] Introduce generic keypair generation interface and engine
@@ -46,12 +46,13 @@ Signed-off-by: istepic <ivan.stepich@gmail.com>
  src/libp11-int.h     |  15 ++-
  src/libp11.h         |  48 +++++++---
  src/p11_front.c      |  32 +++++--
- src/p11_key.c        |  94 ++++++++++++++++++-
+ src/p11_key.c        | 119 +++++++++++++++++++++++-
  src/p11_misc.c       |  75 +++++++++++++++
+ src/p11_slot.c       |   1 +
  tests/Makefile.am    |   6 +-
  tests/keygen.c       | 215 +++++++++++++++++++++++++++++++++++++++++++
  tests/keygen.softhsm |  39 ++++++++
- 11 files changed, 562 insertions(+), 32 deletions(-)
+ 12 files changed, 587 insertions(+), 33 deletions(-)
  create mode 100644 tests/keygen.c
  create mode 100755 tests/keygen.softhsm
 
@@ -329,7 +330,7 @@ index f74f209..f82c9a3 100644
  {
  	PKCS11_OBJECT_private *key = PRIVKEY(pkey);
 diff --git a/src/p11_key.c b/src/p11_key.c
-index ec7f279..8466700 100644
+index ec7f279..57cba43 100644
 --- a/src/p11_key.c
 +++ b/src/p11_key.c
 @@ -252,8 +252,8 @@ int pkcs11_reload_object(PKCS11_OBJECT_private *obj)
@@ -343,16 +344,31 @@ index ec7f279..8466700 100644
  
  	PKCS11_CTX_private *ctx = slot->ctx;
  	CK_SESSION_HANDLE session;
-@@ -266,8 +266,6 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
+@@ -266,10 +266,20 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
  	CK_OBJECT_HANDLE pub_key_obj, priv_key_obj;
  	int rv;
  
 -	(void)algorithm; /* squash the unused parameter warning */
 -
- 	if (pkcs11_get_session(slot, 1, &session))
+-	if (pkcs11_get_session(slot, 1, &session))
++	// R/W session is mandatory for key generation.
++	if (pkcs11_open_session(slot, 1)) {
++		return -1;
++	}
++	// open_session might call C_CloseAllSessions if current session is not R/W.
++	// C_CloseAllSessions logs everyone out
++	if (slot->logged_in <= 0) {
++		if (pkcs11_login(slot, 0, slot->prev_pin)) {
++			return -1;
++		}
++	}
++	if (pkcs11_get_session(slot, 1, &session)) {
  		return -1;
++	}
  
-@@ -310,6 +308,94 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
+ 	/* pubkey attributes */
+ 	pkcs11_addattr(&pubtmpl, CKA_ID, id, id_len);
+@@ -310,6 +320,105 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
  	return 0;
  }
  
@@ -375,6 +391,17 @@ index ec7f279..8466700 100644
 +	ASN1_OBJECT *curve_obj = NULL;
 +	int curve_nid = NID_undef;
 +
++	// R/W session is mandatory for key generation.
++	if (pkcs11_open_session(slot, 1)) {
++		return -1;
++	}
++	// open_session might call C_CloseAllSessions if current session is not R/W.
++	// C_CloseAllSessions logs everyone out
++	if (slot->logged_in <= 0) {
++		if (pkcs11_login(slot, 0, slot->prev_pin)) {
++			return -1;
++		}
++	}
 +	if (pkcs11_get_session(slot, 1, &session)) {
 +		return -1;
 +	}
@@ -531,6 +558,18 @@ index 1b0e64d..1d9a845 100644
 +
 +
  /* vim: set noexpandtab: */
+diff --git a/src/p11_slot.c b/src/p11_slot.c
+index 3c00e22..c5c322b 100644
+--- a/src/p11_slot.c
++++ b/src/p11_slot.c
+@@ -111,6 +111,7 @@ int pkcs11_open_session(PKCS11_SLOT_private *slot, int rw)
+ 	if (rw != slot->rw_mode) {
+ 		CRYPTOKI_call(ctx, C_CloseAllSessions(slot->id));
+ 		slot->rw_mode = rw;
++		slot->logged_in = -1;
+ 	}
+ 	slot->num_sessions = 0;
+ 	slot->session_head = slot->session_tail = 0;
 diff --git a/tests/Makefile.am b/tests/Makefile.am
 index b1bc0fb..a71327d 100644
 --- a/tests/Makefile.am
@@ -822,5 +861,5 @@ index 0000000..83f8175
 +
 +exit 0
 -- 
-2.25.1
+2.39.2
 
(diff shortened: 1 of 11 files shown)


src/hsm.cpp
```diff
@@ -63,12 +64,26 @@ HsmEngine::HsmEngine(const std::string &id,
 
 HsmEngine::~HsmEngine() { _ENGINE_finish(_engine.get()); }
 
+std::string HsmEngine::_constructPkcs11URI(const std::string &keyLabel,
```
garw: I'd prefer an API that features std::optional or an overload with/without key label. Having the empty string special is somewhat ugly.

src/hsm.cpp
```diff
@@ -104,36 +117,60 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
...
-    pkcs11RSAKeygen.key_label = keyLabel.c_str();
+    try {
+        // We need to make sure that we don't have 2 keys with the same ID.
+        // For that we need to pass empty keyLabel. Otherwise libp11 tries to find
+        // a key with exact keyLabel/keyID combination. This means that libp11 might
+        // not recognize that the key with the same ID is already there.
+        loadPrivateKey("", keyID);
+    } catch (const MoCOCrWException &e) {
+        if (e.what() == std::string("Unable to load private key. Private key not found!")) {
+            std::string keyIDHexString = utility::toHex(keyID);
```
garw: It feels a bit weird that all this code is now in the catch blog. I'm also not a 100% sure if this is good. What happens if this code throws again (while its still handling an exception?
Can we move this out of the catch block?

src/hsm.cpp
```diff
@@ -104,36 +117,60 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
...
-    pkcs11ECCKeygen.key_label = keyLabel.c_str();
+    try {
+        // We need to make sure that we don't have 2 keys with the same ID.
+        // For that we need to pass empty keyLabel. Otherwise libp11 tries to find
+        // a key with exact keyLabel/keyID combination. This means that libp11 might
+        // not recognize that the key with the same ID is already there.
+        loadPrivateKey("", keyID);
+    } catch (const MoCOCrWException &e) {
+        if (e.what() == std::string("Unable to load private key. Private key not found!")) {
+            std::string curve = spec.curveName();
```
garw: s.a.

src/mococrw/hsm.h
```diff
@@ -45,8 +45,9 @@ class HSM
     /**
      *  Loads public key from HSM.
      *
-     *  @param keyLabel String based identifer of a key on the token
-     *  @param keyID Vector of raw bytes that identifies a key on the token
+     * @param keyLabel String based identifer of a key on the token
+     * @param keyID Vector of raw bytes that identifies a key on the token
+     * @note keyID can not be empty
```
garw: nit: We'd normally write "must not be empty"

tests/integration/hsm-integration-test.cpp
```diff
@@ -275,20 +275,60 @@ int main(void)
...
+    std::string emptyLabel{};
+    std::vector<uint8_t> keyId_1{0x11};
+    std::vector<uint8_t> keyId_2{0x12};
+    std::string keyLabel_1{"key-label-1"};
+    std::string keyLabel_2{"key-label-2"};
+
+    std::cerr << "Generating a key without specifying the key ID...";
+    try {
+        auto keypair =
+                AsymmetricPrivateKey::generateKeyOnHSM(hsmEngine, eccSpec, keyLabel_1, emptyKeyId);
```
garw: Shouldn't we somehow fail if this succeeds?

tests/integration/hsm-integration-test.cpp
```diff
@@ -275,20 +275,60 @@ int main(void)
...
+    } catch (const MoCOCrWException &e) {
+        std::cerr << std::string(e.what()) + "...";
+    }
+    std::cerr << "Success\n";
+    std::cerr << "Loading a key without specifying the key ID...";
+    try {
+        auto keypair =
+                AsymmetricPrivateKey::readPrivateKeyFromHSM(hsmEngine, keyLabel_1, emptyKeyId);
```
garw: s.a.?

tests/integration/hsm-integration-test.cpp
```diff
@@ -275,20 +275,60 @@ int main(void)
...
+    } catch (const MoCOCrWException &e) {
+        std::cerr << std::string(e.what()) + "...";
+    }
+    std::cerr << "Success\n";
+    std::cerr << "Generating a key without specifying the label...";
+    auto key1 = AsymmetricPrivateKey::generateKeyOnHSM(hsmEngine, eccSpec, emptyLabel, keyId_1);
```
garw: Error handling and proper reporting? Currently, it just crashes with "uncaught exception" if this fails?!

tests/unit/test_hsm.cpp
```diff
@@ -124,31 +125,49 @@ TEST_F(HSMTest, testHSMKeygen)
...
     EXPECT_NO_THROW(AsymmetricKeypair::generateKeyOnHSM(*hsm, eccSpec, keyLabel, keyId));
 }
 
+TEST_F(HSMTest, testHSMTryLoadKeyWithEmptyId)
+{
+    auto hsm = initialiseEngine();
+    const std::string keyLabel{"key-label"};
+    const std::vector<uint8_t> keyId{};
+
+    EXPECT_THROW(AsymmetricPublicKey::readPublicKeyFromHSM(*hsm, keyLabel, keyId),
```
garw, line 149: Doesn't this lack the mock setup?

tests/integration/CMakeLists.txt
```diff
@@ -15,9 +15,10 @@ if(MOCOCRW_HSM_ENABLED)
     add_executable(hsm-integration-test hsm-integration-test.cpp)
     target_link_libraries(hsm-integration-test PRIVATE MoCOCrW::mococrw)
     target_compile_definitions(hsm-integration-test PRIVATE HSM_ENABLED)
+    # libp11 writes to stderr on unsuccessful key loading. Pipe it to devnull
```
garw, line 18: Isn't this bad for debugging?

tests/integration/hsm-integration-test.cpp
```diff
@@ -280,91 +280,123 @@ int main(void)
...
-        std::cerr << "Generated keys with different IDs should not be the same\n";
+        std::cout << "Generated keys with different IDs and empty labels should not be the same"
+                  << std::endl;
+        exit(1);
+    }
+    std::cout << "Success" << std::endl;
+
+    std::cout << "Try to generate a key with the same ID as in previous steps...";
+    try {
+        AsymmetricPrivateKey::generateKeyOnHSM(hsmEngine, eccSpec, keyLabel_1, keyId_1);
```
garw: Fail if succeeds?

garw (changes requested):

tests/unit/test_hsm.cpp
istepic: It immediately throws without any openssl calls

tests/integration/hsm-integration-test.cpp
istepic: Done

tests/integration/hsm-integration-test.cpp
istepic: Done

tests/integration/hsm-integration-test.cpp
istepic: Done

tests/integration/CMakeLists.txt
istepic: Otherwise we are getting output on stderr when key fails to load during precheck in key generation.  This output mixes with integration test output

src/hsm.cpp
istepic: Done

src/mococrw/hsm.h

src/hsm.cpp
istepic: Done

src/hsm.cpp
istepic: Done

tests/integration/hsm-integration-test.cpp
istepic: Added everything to try-catch block

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -343,16 +344,31 @@ index ec7f279..8466700 100644
...
+@@ -266,10 +266,20 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
  	CK_OBJECT_HANDLE pub_key_obj, priv_key_obj;
  	int rv;
  
 -	(void)algorithm; /* squash the unused parameter warning */
 -
- 	if (pkcs11_get_session(slot, 1, &session))
+-	if (pkcs11_get_session(slot, 1, &session))
++	// R/W session is mandatory for key generation.
++	if (pkcs11_open_session(slot, 1)) {
```
wusto, line 355: 1 is a magic number. Do we want to do "#define RW 1" somewhere?
```suggestion
+	if (pkcs11_open_session(slot, RW)) {
```
?

src/hsm.cpp
```diff
@@ -104,6 +160,18 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
                                                  const std::string &keyLabel,
                                                  const std::vector<uint8_t> &keyID)
 {
+    try {
+        // We need to make sure that we don't have 2 keys with the same ID.
+        // For that we need to pass empty keyLabel. Otherwise libp11 tries to find
+        // a key with exact keyLabel/keyID combination. This means that libp11 might
+        // not recognize that the key with the same ID is already there.
+        loadPrivateKey(keyID);
```
wusto, line 166: So we don't expect public keys in the HSM without the corresponding private key?

src/hsm.cpp
```diff
@@ -122,6 +189,18 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const ECCSpec &spec,
...
```
wusto, line 195: s.a.

src/hsm.cpp
```diff
@@ -104,6 +160,18 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
...
+        throw MoCOCrWException("Key with that keyID already exists");
+    } catch (const MoCOCrWException &e) {
+        if (e.what() != std::string("Unable to load private key. Private key not found!")) {
```
wusto: I am wondering if we shouldn't make this a private static const string.
Somehow I don't have a good feeling checking for the same string, which is used somewhere else, if there is a better option.

tests/integration/CMakeLists.txt
wusto: Can you write it to a file?
Then we can at least check the content of the file on local execution. Maybe there is also a way to make this file available after test execution in github actions.

wusto (changes requested):

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
istepic: This is how it's done across all of libp11's codebase :(

src/hsm.cpp
istepic: Pretty much yeah. The most common scenario I would say is that you have a keypair. I think having a private key without a corresponding public key also might be a valid scenario. But having a public key on HSM without the corresponding private doesn't make sense to me.

src/hsm.cpp
istepic: Done

src/hsm.cpp
wusto: at least if you have other mechanisms to protect the public key like secure boot + dm-verity. :)

So the question might be, if there will be ECUs using the keystore without having secure boot + dm-verity.

@garw: Any strong opinion on this?

I would be also fine with just going for the privat key.

wusto (approved):

garw (approved):
I'm still no super happy with the hard-coded test log path, but otherwise, it looks ok to me.

PR #142, opened 2023-03-29
hsm: update libp11 patch

istepic:
Some PKCS#11 implementations don't support all the features libp11 is using. Some of them are removed.

### DIFF
--- a/dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
+++ b/dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
@@ -1,4 +1,4 @@
-From 828282b17f40237bb875e93df0e216e12c4f2504 Mon Sep 17 00:00:00 2001
+From 8468ecc844664c2f066492ed08569d5b7db37756 Mon Sep 17 00:00:00 2001
 From: istepic <ivan.stepich@gmail.com>
 Date: Mon, 5 Dec 2022 22:44:25 +0100
 Subject: [PATCH] Introduce generic keypair generation interface and engine
@@ -46,13 +46,14 @@ Signed-off-by: istepic <ivan.stepich@gmail.com>
  src/libp11-int.h     |  15 ++-
  src/libp11.h         |  48 +++++++---
  src/p11_front.c      |  32 +++++--
- src/p11_key.c        | 119 +++++++++++++++++++++++-
+ src/p11_key.c        | 133 +++++++++++++++++++++++---
+ src/p11_load.c       |   1 -
  src/p11_misc.c       |  75 +++++++++++++++
  src/p11_slot.c       |   1 +
  tests/Makefile.am    |   6 +-
  tests/keygen.c       | 215 +++++++++++++++++++++++++++++++++++++++++++
  tests/keygen.softhsm |  39 ++++++++
- 12 files changed, 587 insertions(+), 33 deletions(-)
+ 13 files changed, 595 insertions(+), 40 deletions(-)
  create mode 100644 tests/keygen.c
  create mode 100755 tests/keygen.softhsm
 
@@ -330,7 +331,7 @@ index f74f209..f82c9a3 100644
  {
  	PKCS11_OBJECT_private *key = PRIVKEY(pkey);
 diff --git a/src/p11_key.c b/src/p11_key.c
-index ec7f279..57cba43 100644
+index ec7f279..5a30e79 100644
 --- a/src/p11_key.c
 +++ b/src/p11_key.c
 @@ -252,8 +252,8 @@ int pkcs11_reload_object(PKCS11_OBJECT_private *obj)
@@ -344,7 +345,12 @@ index ec7f279..57cba43 100644
  
  	PKCS11_CTX_private *ctx = slot->ctx;
  	CK_SESSION_HANDLE session;
-@@ -266,10 +266,20 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
+@@ -262,36 +262,45 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
+ 		CKM_RSA_PKCS_KEY_PAIR_GEN, NULL_PTR, 0
+ 	};
+ 	CK_ULONG num_bits = bits;
+-	CK_BYTE public_exponent[] = { 1, 0, 1 };
++	CK_BYTE public_exponent[] = { 1, 0, 0, 0, 1 };
  	CK_OBJECT_HANDLE pub_key_obj, priv_key_obj;
  	int rv;
  
@@ -353,7 +359,7 @@ index ec7f279..57cba43 100644
 -	if (pkcs11_get_session(slot, 1, &session))
 +	// R/W session is mandatory for key generation.
 +	if (pkcs11_open_session(slot, 1)) {
-+		return -1;
+ 		return -1;
 +	}
 +	// open_session might call C_CloseAllSessions if current session is not R/W.
 +	// C_CloseAllSessions logs everyone out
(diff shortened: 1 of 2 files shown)


dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -368,7 +369,32 @@ index ec7f279..57cba43 100644
  
  	/* pubkey attributes */
  	pkcs11_addattr(&pubtmpl, CKA_ID, id, id_len);
-@@ -310,6 +320,105 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
+ 	if (label)
+ 		pkcs11_addattr_s(&pubtmpl, CKA_LABEL, label);
+ 	pkcs11_addattr_bool(&pubtmpl, CKA_TOKEN, TRUE);
+-	pkcs11_addattr_bool(&pubtmpl, CKA_ENCRYPT, TRUE);
```
garw, line 381: This is for RSA keys? 
I'm not sure if RSA keys can do derive. you can usually use ENCRYPT

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -368,7 +369,32 @@ index ec7f279..57cba43 100644
...
++	pkcs11_addattr_bool(&pubtmpl, CKA_DERIVE, TRUE);
+ 	pkcs11_addattr_bool(&pubtmpl, CKA_VERIFY, TRUE);
+-	pkcs11_addattr_bool(&pubtmpl, CKA_WRAP, TRUE);
+ 	pkcs11_addattr_var(&pubtmpl, CKA_MODULUS_BITS, num_bits);
+ 	pkcs11_addattr(&pubtmpl, CKA_PUBLIC_EXPONENT, public_exponent, 3);
```
garw: public exponent 3 is strongly discouraged. Why do we set this? Do we really need to? The common public exponent for RSA is nowadays 0x10001

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -368,7 +369,32 @@ index ec7f279..57cba43 100644
...
+ 
+@@ -286,12 +295,11 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
+ 	pkcs11_addattr(&privtmpl, CKA_ID, id, id_len);
+ 	if (label)
+ 		pkcs11_addattr_s(&privtmpl, CKA_LABEL, label);
+-	pkcs11_addattr_bool(&privtmpl, CKA_TOKEN, TRUE);
+ 	pkcs11_addattr_bool(&privtmpl, CKA_PRIVATE, TRUE);
++	pkcs11_addattr_bool(&privtmpl, CKA_TOKEN, TRUE);
+ 	pkcs11_addattr_bool(&privtmpl, CKA_SENSITIVE, TRUE);
+-	pkcs11_addattr_bool(&privtmpl, CKA_DECRYPT, TRUE);
```
garw: s.a. This probably needs to depend on the algorithm

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -417,36 +443,31 @@ index ec7f279..57cba43 100644
...
-+	ecdsa_params_len = i2d_ASN1_OBJECT(curve_obj, NULL);
-+	ecdsa_params = (unsigned char *)OPENSSL_malloc(ecdsa_params_len);
-+	if (!ecdsa_params)
++	ec_params_len = i2d_ASN1_OBJECT(curve_obj, NULL);
++	ec_params = (unsigned char *)OPENSSL_malloc(ec_params_len);
++	if (!ec_params)
 +		return -1;
-+	tmp = ecdsa_params;
++	tmp = ec_params;
 +	i2d_ASN1_OBJECT(curve_obj, &tmp);
```
garw: readability: Why do we pass tmp here and not ec_params? I just looked up what I believe to be the reason for this in OpenSSLs documentation. tmp is a bad variable name if the purpose isn't obvious and it deserves a comment here to explain whats going on with strange OpenSSL API.

return value check?

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -417,36 +443,31 @@ index ec7f279..57cba43 100644
...
 +
 +	/* pubkey attributes */
 +	pkcs11_addattr(&pubtmpl, CKA_ID, id, id_len);
 +	if (label)
 +		pkcs11_addattr_s(&pubtmpl, CKA_LABEL, label);
 +	pkcs11_addattr_bool(&pubtmpl, CKA_TOKEN, TRUE);
-+	pkcs11_addattr_bool(&pubtmpl, CKA_DERIVE, FALSE);
-+	pkcs11_addattr_bool(&pubtmpl, CKA_WRAP, FALSE);
++	pkcs11_addattr_bool(&pubtmpl, CKA_DERIVE, TRUE);
```
garw: I don't think we need DERIVE on the pubkey. Not sure if this even works.

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -474,6 +495,19 @@ index ec7f279..57cba43 100644
...
+diff --git a/src/p11_load.c b/src/p11_load.c
+index e89b0c7..3f69cec 100644
+--- a/src/p11_load.c
++++ b/src/p11_load.c
+@@ -81,7 +81,7 @@ int pkcs11_CTX_load(PKCS11_CTX *ctx, const char *name)
+ 	/* Tell the PKCS11 to initialize itself */
+ 	memset(&args, 0, sizeof(args));
+ 	/* Unconditionally say using OS locking primitives is OK */
+-	args.flags |= CKF_OS_LOCKING_OK;
++//	args.flags |= CKF_OS_LOCKING_OK;
```
garw: If we comment this out we can remove it entirely?

garw (changes requested):

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
istepic: Removed derive

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
istepic: 3 is the length. Changed it to 0x10001

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
istepic: Documented and return value checked

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
istepic: Removed

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
istepic: yes

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -417,36 +448,38 @@ index ec7f279..57cba43 100644
...
 +	/* privkey attributes */
 +	pkcs11_addattr(&privtmpl, CKA_ID, id, id_len);
 +	if (label)
 +		pkcs11_addattr_s(&privtmpl, CKA_LABEL, label);
-+	pkcs11_addattr_bool(&privtmpl, CKA_TOKEN, TRUE);
 +	pkcs11_addattr_bool(&privtmpl, CKA_PRIVATE, TRUE);
++	pkcs11_addattr_bool(&privtmpl, CKA_TOKEN, TRUE);
 +	pkcs11_addattr_bool(&privtmpl, CKA_SENSITIVE, TRUE);
 +	pkcs11_addattr_bool(&privtmpl, CKA_DERIVE, TRUE);
-+	pkcs11_addattr_bool(&privtmpl, CKA_UNWRAP, FALSE);
```
johnfxgalea-e, line 447: Overall, can there be a comment explaining rationale for disabling/enabling certain attributes? It is hard to follow what your intent is, backed by the standard.

Keep in mind that some PKCS#11 attributes have default values so it could be that removal does not imply falsifying the attribute.

For instance, we don't want wrapping functionality, right? so leaving it false explicitly is a good thing no?

istepic: Where do we stop in that case? There are many object attributes. Do we want to have an explicit opinion on each and every one or do we leave some to the default values of PKCS#11 implementation? Also, problems could arise from PKCS#11 implementation not supporting setting a particular attribute. I think an argument can be made to make these flags configurable on the API level but it's too late for that now and we can think about that in the future when we move to OpenSSL 3.

I will update the commit message trying to explain the change.

garw (approved):

moh299814 (approved):
Based on http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850449 and related templates the flags seems fine, but it should be good to have some comments for the flags or at least a link to the related section in the reference document.

istepic:
> Based on http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850449 and related templates the flags seems fine, but it should be good to have some comments for the flags or at least a link to the related section in the reference document.

I added a comment about what the flags are used for. People using libp11 should be able to find the rest in the standard.

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
johnfxgalea-e: Commit message makes things clearer, thanks.

johnfxgalea-e (approved):

PR #146, opened 2023-05-08
hsm: get RW session instead of creating it

istepic:
Due to engine_finish not being called on HsmEngine object destruction, sessions remain open. Approach to acquiring sessions before key generation is changed. If RW session already exists, a new one is not opened but reused. This approach is better irrespective of problems with engine_finish not being called.

### DIFF
--- a/dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
+++ b/dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
@@ -1,4 +1,4 @@
-From f5bb68262f9f32c3b121a25ac5ecc7b9094f7348 Mon Sep 17 00:00:00 2001
+From 236cf8101b3df3914e3986ce685b866f880f02df Mon Sep 17 00:00:00 2001
 From: istepic <ivan.stepich@gmail.com>
 Date: Mon, 5 Dec 2022 22:44:25 +0100
 Subject: [PATCH] Introduce generic keypair generation interface and engine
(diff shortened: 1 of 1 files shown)


dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
```diff
@@ -424,12 +423,11 @@ index ec7f279..5a30e79 100644
 +	int curve_nid = NID_undef;
 +
 +	// R/W session is mandatory for key generation.
-+	if (pkcs11_open_session(slot, 1)) {
-+		return -1;
-+	}
-+	// open_session might call C_CloseAllSessions if current session is not R/W.
-+	// C_CloseAllSessions logs everyone out
-+	if (slot->logged_in <= 0) {
++	if (slot->rw_mode == 0) {
```
johnfxgalea-e: I think there is a race condition here

johnfxgalea-e (changes requested):
The general idea of the guard looks okay but it appears there is a race condition.

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
johnfxgalea-e: Also, can't rw_mode also be -1 if not initialised? https://github.com/OpenSC/libp11/blob/281254bf3883ed17f4756243d78ac906342211d8/src/p11_slot.c#L427

istepic: > I think there is a race condition here

Can you clarify?

> Also, can't rw_mode also be -1 if not initialised? https://github.com/OpenSC/libp11/blob/281254bf3883ed17f4756243d78ac906342211d8/src/p11_slot.c#L427

True, done.

johnfxgalea-e (approved):

dockerfiles/feature-support/hsm-patches/0001-Introduce-generic-keypair-generation-interface-and-e.patch
johnfxgalea-e: Discussed offline with @istepic on thread safety guarantees (or lack there of by MoCOCrW).
See:
https://github.com/bmwcarit/MoCOCrW/blob/cc2acba5d6e55c94a144cb9306c9fb2345153911/README.md?plain=1#LL84C48-L84C90

garw (approved):

garw (approved):

PR #147, opened 2023-06-15
Initials Support

cps-b:
This PR adds support for the initials attribute in the subject of the certificates.

### DIFF
--- a/src/mococrw/distinguished_name.h
+++ b/src/mococrw/distinguished_name.h
@@ -91,10 +91,7 @@ public:
      * @return string with the initials if defined in the certificate's subject
      * @return empty string if undefined
      */
-    std::string initials() const
-    {
-        return _getAttributeByNIDAsString(openssl::ASN1_NID::Initials);
-    }
+    std::string initials() const { return _getAttributeByNIDAsString(openssl::ASN1_NID::Initials); }
 
     std::string userId() const { return _getAttributeByNIDAsString(openssl::ASN1_NID::UserId); }
     std::string title() const { return _getAttributeByNIDAsString(openssl::ASN1_NID::Title); }
--- a/src/mococrw/distinguished_name.h
+++ b/src/mococrw/distinguished_name.h
@@ -85,6 +85,12 @@ public:
     {
         return _getAttributeByNIDAsString(openssl::ASN1_NID::GivenName);
     }
+
+    /**
+     * Get the initials attribute of the selected distinguished name
+     * @return string with the initials if defined in the certificate's subject
+     * @return empty string if undefined
+     */
     std::string initials() const
     {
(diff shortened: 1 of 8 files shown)


garw (changes requested):
Looks mostly good. Can we please also amend the tests to verify that X509 name population works correctly in the preserveOrder and the not preserve Order scenario? Your unit test currently only checks that reading it works correctly.

johnfxgalea-e (approved):
LGTM. I agree with Martin regarding additional tests. You can also add a test for when initials is not defined as well,

johnfxgalea-e (approved):
LGTM

src/mococrw/distinguished_name.h
```diff
@@ -85,6 +85,14 @@ class DistinguishedName
     {
         return _getAttributeByNIDAsString(openssl::ASN1_NID::GivenName);
     }
+
+    /**
```
istepic, line 89: nitpick: reconsider the value of this comment. It's stating a special case of existing comment on line 45.

src/mococrw/distinguished_name.h
```diff
@@ -245,4 +255,16 @@ DistinguishedName::Builder &DistinguishedName::Builder::title(T &&name)
     return *this;
 }
 
+/**
+ * Sets the initial attribute in the distinguished name builder
```
istepic, line 259: nitpick: reconsider the value of this comment.

src/mococrw/openssl_wrap.h
```diff
@@ -441,6 +441,7 @@ enum class ASN1_NID : int {
     GivenName = NID_givenName,
     UserId = NID_userId,
     Title = NID_title,
+    Initials = NID_initials,  ///< initials NID enum entry
```
istepic, line 444: nitpick: reconsider the value of this comment.

istepic (approved):

src/mococrw/openssl_wrap.h
cps-b: I agree, especially since the other values are also undocumented. I raised this inconsistency, however I was explicitly asked to add the comments for the doxygen. It is also part of the acceptance criteria.

src/mococrw/distinguished_name.h
cps-b: I agree, especially since the other functions are also undocumented. I raised this inconsistency, however I was explicitly asked to add the comments for the doxygen. It is also part of the acceptance criteria.

src/mococrw/distinguished_name.h
cps-b: I agree, especially since the other functions are also undocumented. I raised this inconsistency, however I was explicitly asked to add the comments for the doxygen. It is also part of the acceptance criteria.

garw (approved):

PR #155, opened 2023-10-30
Use Builder Pattern for HsmKeyParams

cps-b:
Adapt the HsmKeyParams to use a Builder Pattern instead of a struct

### DIFF
--- a/src/hsm.cpp
+++ b/src/hsm.cpp
@@ -216,8 +216,8 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
     pkcs11RSASpec.bits = spec.numberOfBits();
 
     PKCS11_params _params;
-    _params.extractable = static_cast<unsigned char>(params.cka_extractable);
-    _params.sensitive = static_cast<unsigned char>(params.cka_sensitive);
+    _params.extractable = static_cast<unsigned char>(params.isExtractable());
+    _params.sensitive = static_cast<unsigned char>(params.isSensitive());
 
     PKCS11_KGEN_ATTRS pkcs11RSAKeygen;
     pkcs11RSAKeygen.type = EVP_PKEY_RSA;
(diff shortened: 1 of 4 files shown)


src/mococrw/hsm.h
```diff
@@ -26,17 +26,49 @@ class ECCSpec;
...
+ * This class currently contains PKCS#11 attributes which are changeable on key creation.
  * In the future also parameters for other keystorage interfaces can be added.
  */
-struct HsmKeyParams
+class HsmKeyParams
 {
+public:
+    class Builder;
+
+    bool isExtractable() const { return cka_extractable; }
```
garw: Do we really want to expose PKCS#11 slang here? Wouldn't it be better if this provides some Abstraction and we say here that the param is that the is is extractable or not? Sensitive is really not self-explanatory.

src/hsm.cpp
```diff
@@ -233,7 +234,8 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const ECCSpec &spec,
                                                  const std::string &keyLabel,
                                                  const std::vector<uint8_t> &keyID)
 {
-    HsmKeyParams hsmKeyParams;
+    HsmKeyParams hsmKeyParams =
+            HsmKeyParams::Builder().setCkaExtractable(false).setCkaSensitive(true).build();
```
garw: Style: Most of the library uses {} for object construction.

garw (changes requested):

src/mococrw/hsm.h
```diff
@@ -34,34 +34,25 @@ class HsmKeyParams
...
 
-    bool isExtractable() const { return cka_extractable; }
-
-    bool isSensitive() const { return cka_sensitive; }
+    bool isExtractable() const { return extractable; }
 
 private:
-    bool cka_extractable;
-    bool cka_sensitive;
+    bool extractable;
```
garw: consistency: this lib prefixes private members with _

garw (dismissed):
looks cleaner to me now. Feel free to merge after the consistency problem has been fixed.

cps-b:
Merging in PR #156

PR #156, opened 2023-11-06
Use Builder Pattern for HsmKeyParams

cps-b:
Adapt the HsmKeyParams to use a Builder Pattern instead of a struct

### DIFF
--- a/src/hsm.cpp
+++ b/src/hsm.cpp
@@ -216,8 +216,8 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const RSASpec &spec,
     pkcs11RSASpec.bits = spec.numberOfBits();
 
     PKCS11_params _params;
-    _params.extractable = static_cast<unsigned char>(params.cka_extractable);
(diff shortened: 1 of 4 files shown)


src/hsm.cpp
```diff
@@ -261,8 +261,9 @@ openssl::SSL_EVP_PKEY_Ptr HsmEngine::generateKey(const ECCSpec &spec,
     pkcs11ECCSpec.curve = curve.c_str();
 
     PKCS11_params _params;
-    _params.extractable = static_cast<unsigned char>(params.cka_extractable);
-    _params.sensitive = static_cast<unsigned char>(params.cka_sensitive);
+    // If the key is extractable it shouldn't be sensitive and vice versa
```
istepic, line 264: I guess this makes sense in principle but is there a strong reason why do we want to enforce it?

cps-b: @istepic This was requested in the review from @garw in #155. We should hide the PKCS#11 internals...

istepic (approved):
Don't forget to squash

PR #158, opened 2023-11-22
Reintroduce deprecated cka api for compatibility

cps-b:
This PR reintroduces the API using HsmKeyParams that was removed in #156 . We want to keep it around and then remove it in a subsequent release.

### DIFF
--- a/src/key.cpp
+++ b/src/key.cpp
@@ -246,6 +246,27 @@ AsymmetricKeypair AsymmetricKeypair::generateKeyOnHSM(HSM &hsm,
 }
 
 AsymmetricKeypair AsymmetricKeypair::generateKeyOnHSM(HSM &hsm,
+                                                      const RSASpec &spec,
+                                                      const std::string &keyLabel,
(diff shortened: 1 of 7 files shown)


src/key.cpp
```diff
@@ -245,6 +245,27 @@ AsymmetricKeypair AsymmetricKeypair::generateKeyOnHSM(HSM &hsm,
     }
 }
 
+AsymmetricKeypair AsymmetricKeypair::generateKeyOnHSM(HSM &hsm,
+                                                      const RSASpec &spec,
+                                                      const std::string &keyLabel,
+                                                      const std::vector<uint8_t> &keyID,
+                                                      const HsmKeyParams &params)
+{
+    // libp11 uses 128 byte buffer
```
yannik-dittmar: As someone with very little knowledge about this software I'm a bit confused what this comment should tell me... Is this the limiting factor for the key size and if yes why? Do you think it would make sense to describe it with one or two more words and also change the comment in line 234/294/314?

yannik-dittmar (approved):
LGTM

